Issue #11, opened 2022-08-10
[crypto] zeroize all secret cryptographic materials


punwai:
### Description
Ensure that we zeroize on drop all secrets (private key and key pairs) for every signature scheme implemented.

- [x] BLS12-381
- [x] edDSA
- [x] ecDSA
- [x] BLS12-377

kchalkias:
@joyqvq are we done with that issue?

joyqvq:
need to confirm the last two. will do this week

PR #23, opened 2022-08-25
crypto: zeroize ed25519 and secp256k1

joyqvq:
https://github.com/MystenLabs/fastcrypto/issues/11

### DIFF
--- a/src/ed25519.rs
+++ b/src/ed25519.rs
@@ -16,6 +16,7 @@ use std::{
     fmt::{self, Display},
     str::FromStr,
 };
+use zeroize::{Zeroize, ZeroizeOnDrop};
 
 use crate::{
     pubkey_bytes::PublicKeyBytes,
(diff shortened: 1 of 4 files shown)


src/ed25519.rs
```diff
@@ -490,7 +491,7 @@ impl KeyPair for Ed25519KeyPair {
     }
 
     fn private(self) -> Self::PrivKey {
-        self.secret
+        Ed25519PrivateKey::from_bytes(self.secret.as_ref()).unwrap()
```
punwai, line 494: Is the idea of this being we drop ‘self’ and zeroize it vs risking the compiler copying the secret in memory and leaving it there?

huitseeker: I think this is brought by implementing `Drop` which prevents you from moving out of self. I have my issues with this `private` method ...

huitseeker (approved):
LGTM!

Issue #15, opened 2022-07-16
[cryptography] Integrate Wycheproof tests in unit testing


huitseeker:
As we are integrating more crypto schemes in our library (see e.g. MystenLabs/narwhal#484), it would sure be nice to integrate wycheproof ( https://github.com/google/wycheproof ) to our tests to make sure our current and future implementations are not too clowny. In rust, that has recently been made much more easy using https://github.com/randombit/wycheproof-rs

PR #25, opened 2022-08-28
Add wycheproofs

joyqvq:
https://github.com/MystenLabs/fastcrypto/issues/15


src/tests/secp256k1_tests.rs
```diff
@@ -370,3 +373,47 @@ proptest::proptest! {
...
+            let mut res = TestResult::Invalid;
+
+            for i in 0..4 {
+                n_bytes[64] = i;
+                let sig = <Secp256k1Signature as ToFromBytes>::from_bytes(&n_bytes).unwrap();
+                if pk
+                    .verify_hashed(&k256::sha2::Sha256::digest(&test.msg), &sig)
+                    .is_ok()
+                {
+                    res = TestResult::Valid;
```
kchalkias, line 403: Ok, so if it least one pk succeeds we assume it's fine. I'm fine with that just because wycheproof tests do not support ecrecover.

kchalkias (approved):

src/tests/ed25519_tests.rs
```diff
@@ -549,3 +550,24 @@ fn test_sk_zeroization_on_drop() {
         unsafe { ::std::slice::from_raw_parts(bytes_ptr, ED25519_PRIVATE_KEY_LENGTH) };
     assert_ne!(sk_memory, &sk_bytes[..]);
 }
+
+#[test]
+fn wycheproof_test() {
```
kchalkias, line 555: Note that we use zip215 (the cofactored version of ed25519 verification). 

I'm happy that the wycheproof tests work independently on the variation, so I guess they were carefully selected to be agnostic to the ed25519 mode. That said approving this PR.

Note: But if in the future (ie they update their json edge cases) any of these tests fail, we should check that it's not because of our zip215 mode.

joyqvq: if we want to be on top of the wycheproof, we need to import the json file from google/wycheproof directly, the wycheproof crate may be lagging behind

PR #29, opened 2022-09-04
Code hygiene fixes

kchalkias:
- It's a good hygiene (+ for consistency and readability purposes) that inherited functions are implemented in the same order with the trait definitions.
- various other fixes

### DIFF
--- a/src/tests/bls12381_tests.rs
+++ b/src/tests/bls12381_tests.rs
@@ -205,7 +205,7 @@ fn verify_batch_missing_parameters_length_mismatch() {
     let (digest1, digest2, pubkeys1, pubkeys2, aggregated_signature1, aggregated_signature2) =
         verify_batch_aggregate_signature_inputs();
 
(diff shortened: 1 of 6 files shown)


src/tests/bls12381_tests.rs
```diff
@@ -427,7 +427,7 @@ fn test_sk_zeroization_on_drop() {
     // Check that self.privkey is zeroized
     unsafe {
         for i in 0..BLS12381PrivateKey::LENGTH {
-            assert!(*ptr.add(i) == 0);
+            assert_eq!(*ptr.add(i), 0);
```
joyqvq, line 430: worth updating https://github.com/MystenLabs/fastcrypto/blob/f7e53259893857bb792bb6933c465ded0026c93f/src/tests/secp256k1_tests.rs#L297 as well

joyqvq (dismissed):
lgtm!

Issue #1, opened 2022-08-17
Official address encoding base58check Vs bech32 vs base64


joyqvq:
discuss after public testnet

joyqvq:
after some attempts on using bech32 as Address and ObjectID encoding: https://github.com/MystenLabs/sui/pull/5643

there are few challenges that is hard to overcome due to `AccountAddress` defined in Move: 
1. All system object we use 0x5 instead of 0x000000000000000005 for representation, this makes it difficult for an objectID to be represented succinctly in bech32. 
2. In move contract, an address type uses AccountAddress in move as underlying, which is represented as hex, example https://github.com/MystenLabs/sui/blob/ea71d821684ccc1bae7034f2229c509c1b5318b8/crates/sui-framework/sources/governance/sui_system.move#L361
`assert!(tx_context::sender(ctx) == @0x0, 0);` this will cause a visual discrepancy on how sui uses an address vs what an address looks like onchain. 
3. the changes is quite invasive across the board, since we use AccountAddress in rust interchangeablely with SuiAddress and ObjectID (PR attempt: https://github.com/MystenLabs/sui/pull/5643)

Decision: to keep the address in hex instead of Bech32

joyqvq:
Another question is whether we want to use 32 or 20 bytes for an hex address? this should be less invasive since there is a flag in Move to do this already. See all changes needed in https://github.com/MystenLabs/sui/pull/6700

joyqvq:
cc @Jordan-Mysten to add considerations for address 32 bytes frontend 

also cc @kchalkias

benr-ml:
@kchalkias, @joyqvq can we close this one? do you think we will want to change it in the future?

joyqvq:
https://github.com/MystenLabs/sui/pull/8542 considers 32 bytes hex as final address format for mainnet.

PR #30, opened 2022-09-05
feat: implement display/debug traits for private keys

### LINKED ISSUE #14: [crypto] Implement display for our keys
At the moment our logs are riddled with the following output:
```
EdwardsPoint{
	X: FieldElement51([986312278720011, 308481451238697, 499818313507985, 2043434245230999, 529376140600966]),
	Y: FieldElement51([1122918900483925, 1840648545462405, 2049000554172011, 588549553128558, 2195713730548988]),
	Z: FieldElement51([1, 0, 0, 0, 0]),
	T: FieldElement51([167659880001565, 819468384591364, 1320141236765074, 698068994910468, 216442548444488])
}))
```

We should make sure we have a decent `Display`/`Debug` implemented for all public key types, especially as we're about to embed those data types that can contain keys of several types (for user keys) : we will want to e.g. derive / reuse `Display` on the Sui side.

Conversely, for the `PrivateKey` type, we should make sure our implementations of `Debug` / `Display` **never** display the byte content of the keys, but rather some "elided private ed25519 key" sort of string.

erwanor:
Fixes #14, at least partially! For the public keys, base64 will have to do until #1 is resolved. 

For consideration: do you think a `Secret` derive macro that:
1. derives `Zeroize`
2. implements safe `Display` and `Debug`
3. implements `unsafe_write_secret: &self -> String` (for debugging)

makes sense? I could slot it into this PR. This would make the code nicer:

```rust
#[derive(SecretMaterial)]
 pub struct BLS12381PrivateKey {
     pub privkey: blst::SecretKey,
     pub bytes: OnceCell<[u8; BLS_PRIVATE_KEY_LENGTH]>,
     ...
}
```


punwai:
This is perfect @erwanor. The PR looks good to me!

src/ed25519.rs
```diff
@@ -13,7 +13,7 @@ use serde_bytes::{ByteBuf, Bytes};
 use serde_with::serde_as;
 use signature::{rand_core::OsRng, Signature, Signer, Verifier};
 use std::{
-    fmt::{self, Display},
+    fmt::{self, Display, Debug},
```
kchalkias, line 16: reorder to `fmt::{self, Debug, Display}` to make rustfmt happy :)

src/bls12381.rs
```diff
@@ -372,6 +372,18 @@ impl Signer<BLS12381Signature> for BLS12381PrivateKey {
...
 
+impl Display for BLS12381PrivateKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "[elided BLS12381PrivateKey]")
+    }
+}
+
+impl Debug for BLS12381PrivateKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self)
```
kchalkias, line 383: How are these printed (please include an example debug log per scheme)? 
We'll probably need a pretty-print or hex here.

erwanor: Both `Display::fmt` and `Debug::fmt` will return the same `[elided XXXPrivateKey]`, in fact in that snippet `Debug::fmt` returns `Display::fmt`. 

My reasoning is that if displaying secret materials for debugging purposes is necessary, then it should be through a non-standard method labeled as such, for example: `unsafe_fmt_private_key(&PrivateKey) -> String`. 

This makes it less likely that a downstream debug statement `{:?}` slipping through review turns into a secret leak in the logs (imo)

src/ed25519.rs
erwanor: Ah yes, will lint

huitseeker (commented):
Thanks @erwanor ! This is good, but a tad tedious to implement for each scheme, would it be worth implementing those as a proc-macro [à la Diem](https://github.com/diem/diem/blob/185ae2e28bdd4c9b15fc5478eb17b95a8da1e092/crypto/crypto-derive/src/lib.rs#L113-L141)?

src/bls12381.rs
erwanor: Option B sounds nice and balanced! I can hack a PoC later, but my hunch is that it will need a `PrivateKey` trait if we want a derive macro `#[derive(SafeDisplay)]` that works across schemes.

kchalkias: Thanks @erwanor, no need to update or work on that now, it's still requiring some extra thoughts + there is a debate among cryptographers re the best strategy. We could probably allow it as a feature in a future version of fastcrypto, but not today.

huitseeker (approved):

Cargo.toml
```diff
@@ -36,6 +36,7 @@ blst = "0.3.10"
 digest = "0.10.3"
 once_cell = "1.13.1"
 readonly = "0.2.2"
+diem-crypto-derive = "0.0.3"
```
kchalkias, line 39: I'd personally avoid a diem dependency, especially because this code part is not complex to be ported (copy pasted with a comment reference) without maintainability issues. However, we can do in an upcoming PR. If we do in upcoming PR, please merge and add a github issue.

kchalkias (dismissed):

Cargo.toml
erwanor: I can do it now, but macros [need their own special crate type](https://doc.rust-lang.org/reference/procedural-macros.html), so we will have a nested crate `fastcrypto::crypto_derive.

I'll push the changes shortly.

huitseeker (approved):
😍

crypto-derive/src/lib.rs
```diff
@@ -0,0 +1,36 @@
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::DeriveInput;
+
+// Imported from diem-crypto-derive@0.0.3
+// https://github.com/diem/diem/blob/release-1.4.3/crypto/crypto-derive/src/lib.rs#L113
+
+#[proc_macro_derive(SilentDisplay)]
```
kchalkias, line 8: Awesome! thanks

kchalkias (approved):

PR #32, opened 2022-09-06
chore: Add ci to detect semver and api changes

joyqvq:
- added semver check in ci
- public-api.txt is generated via https://crates.io/crates/cargo-public-api (some failed attempt to add this to the ci, it seems to be i need to follow up in a new pr)


public-api.txt
```diff
@@ -0,0 +1,384 @@
+pub async fn fastcrypto::SignatureService::request_signature(&mut self, digest: Digest) -> Signature
```
huitseeker, line 1: Is this auto-generated? If so how/by what?

.github/workflows/rust.yml
```diff
@@ -91,3 +91,20 @@ jobs:
...
+          toolchain: stable
+          override: true
+          profile: minimal
+      - run: cargo doc --all-features
+  
+  cargo-semver-checks:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: obi1kenobi/cargo-semver-checks-action@v1
```
huitseeker, line 110: This should work with the [default semver-check assumptions](https://github.com/obi1kenobi/cargo-semver-checks-action#cargo-semver-checks-action), given this is a mono-crate repo that follows the tagging convention (which we should not forget to use), but would open up the crate to a wild west of changes on every PR as soon as one commit bumps the version number. That is, once the version number is bumped, there's no longer any CI signal on say, PRs that change an API method, because the "major version change" has already been enacted.

That's why I think it may be more useful to have `cargo-semver-checks-action` trigger on publish, such as in https://github.com/MystenLabs/mysten-infra/pull/145

But YMMV.

.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
+#
```
huitseeker, line 1: 👍

public-api.txt
joyqvq: using `cargo public-api`, worth adding to the ci but it needs an initial publick-api.txt and ci can detect changes

https://github.com/Enselic/cargo-public-api/tree/main

huitseeker (commented):
While you're here you may want to consider using https://github.com/MystenLabs/narwhal/blob/main/.github/workflows/docs.yml for pushing the rustdoc to Github Pages:

.github/workflows/publish.yml
```diff
@@ -0,0 +1,71 @@
...
+          set -euxo pipefail
+          export CURRENT_VERSION="$(./scripts/get_current_version.sh ${{ matrix.package}})"
+          git tag "${{ matrix.package}}-v$CURRENT_VERSION"
+          git push origin "${{ matrix.package}}-v$CURRENT_VERSION"
+      - uses: actions-rs/toolchain@v1
+        if: steps.check.outputs.is_new_version == 'yes'
+      - name: Publish
+        if: steps.check.outputs.is_new_version == 'yes'
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
```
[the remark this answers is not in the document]
joyqvq: done!

huitseeker (approved):
Nice, thanks!

PR #33, opened 2022-09-08
Cleaning up hkdf to adhere to rfc5869

benr-ml:


src/hkdf.rs
```diff
@@ -39,19 +39,19 @@ use hkdf::hmac::Hmac;
...
 ///     assert_ne!(native_sk.to_bytes(), my_keypair.private().as_bytes());
 /// # }
 /// ```
-pub fn hkdf_generate_from_ikm<'a, H: OutputSizeUser, K: KeyPair>(
-    ikm: &[u8],             // IKM (32 bytes)
-    salt: &[u8],            // Optional salt
-    info: Option<&'a [u8]>, // Optional domain
+pub fn hkdf_generate_from_ikm<H, K>(
+    ikm: &[u8],  // IKM (32 bytes).
+    salt: &[u8], // Salt (can be empty).
```
kchalkias, line 53: Ideally the salt should be an `Option`: checking the internals of the `hkdf` dep, and there is already a default salt if you add `None`. I guess this is preferable to comply with the underlying lib (which is already adheres to rfc5869), see https://github.com/RustCrypto/KDFs/blob/master/hkdf/src/lib.rs#L146

src/hkdf.rs
```diff
@@ -39,19 +39,19 @@ use hkdf::hmac::Hmac;
...
+    info: &[u8], // Info (can be empty).
```
kchalkias, line 54: Indeed for the `info` field both the underlying dependency lib and the rfc, do not optionally expect it and can be empty. This is different from the salt which defaults to a HashLen zero byte string.

Per RFC:

       salt     optional salt value (a non-secret random value);
               **if not provided, it is set to a string of HashLen zeros.**
               
       info     optional context and application specific information
               **(can be a zero-length string)**

kchalkias (changes requested):
Now that we're here, I will also suggest testing against [wycheproof](https://github.com/google/wycheproof) as we did with other crypto schemes, see https://github.com/MystenLabs/fastcrypto/blob/main/src/tests/ed25519_tests.rs#L555

src/hkdf.rs
benr-ml: Right, the library does that, which I find confusing since it also pads `salt` with zeros in case `salt` is shorter than the hash input length (as specified by HMAC). In other words, calling the external hkdf with `Some(&[])` or `None` results in the same output.

kchalkias: Good point, I'd prefer if we had 

- a test to ensure the equivalence of both None and &[] just to ensure this assumption will stay true as the hkdf dep version evolves (in the future)
- add a comment line to explain what you mentioned above for future code readers.

benr-ml: Done

kchalkias (approved):

kchalkias:
Awesome!!

src/tests/secp256k1_tests.rs
```diff
@@ -253,6 +253,37 @@ fn verify_invalid_signature() {
     assert!(kp.public().verify(&digest.0, &signature).is_err());
 }
 
+#[test]
+fn fail_to_verify_if_upper_s() {
```
kchalkias, line 257: Why do we need this here? I thought we've already added that functionality + this PR should be HKDF specific.

benr-ml: sorry, did a merge instead of rebase with main, reverted

kchalkias (approved):
Awesome and thanks!

PR #35, opened 2022-09-09
Add AES CTR, CBC, GCM

jonas-lj:
Added support for AES with 128 and 256 bit keys using either CTR, CBC, GCM with PKCS#7 padding. All modes are backed by crates from RustCrypto (https://github.com/RustCrypto/block-modes).


src/tests/aes_tests.rs
```diff
@@ -0,0 +1,161 @@
...
+use crate::{
+    aes::{
+        Aes128CbcPkcs7, Aes128Cfb, Aes128Cfb8, Aes128Ctr, Aes256CbcPkcs7, Aes256Cfb, Aes256Cfb8,
+        Aes256Ctr, Key,
+    },
+    traits::{Cipher, EncryptionKey},
+};
+use rand::{rngs::StdRng, SeedableRng};
+
+#[test]
```
kchalkias, line 12: can we also add `wycheproof` tests for completeness?

kchalkias (commented):
Another general comment is if we did some digging and comparison of AES implementations, as one of the things we want to advertise is faster crypto. The goal of this repo is to pick the fastest possible (but safe) Rust impl and wrap or augment them.

src/tests/aes_tests.rs
jonas-lj: There doesn't seem to be wycheproof tests for any of the modes I've added so far. There's only tests for authenticated encryption, but I think it would be a good idea to add those along with some wycheproof tests.

jonas-lj:
> 

I did some research on it, and there doesn't seem to be a lot of options. The RustCrypto modes seem to be by far the most used and best supported. However, I did look at https://github.com/keepsimple1/libaes which has a benchmark showing that it's about 3-4 times faster than RustCrypto for CBC. I tried it with the most recent version of RustCrypto, and the benchmark shows the same, so we could use this lib at least for the modes it supports.

src/traits.rs
```diff
@@ -193,3 +193,23 @@ pub trait AggregateAuthenticator:
...
+    const LENGTH: usize;
+
+    fn generate<R: CryptoRng + RngCore>(rng: &mut R) -> Self;
+}
+
+/// Trait impl'd by symmetric ciphers. This should also keep anything needed to encrypt and decrypt including keys and IVs
+///
+pub trait Cipher {
+    /// Encrypt `plaintext` and write result to `buffer`
+    fn encrypt_b2b(&self, plaintext: &[u8], buffer: &mut [u8]) -> Result<(), Error>;
```
kchalkias, line 211: Wondering if we should rename these to just `encrypt` and `decrypt`

Cargo.toml
```diff
@@ -26,6 +26,11 @@ zeroize = "1.5.7"
 bulletproofs = "4.0.0"
 curve25519-dalek-ng = "4.1.1"
 merlin = "3.0.0"
+aes = "0.8.1"
+ctr = "0.9.1"
+cbc = "0.1.2"
+cfb-mode = "0.8.1"
```
kchalkias, line 32: I think we could retire cfb for now, and replace it with GCM which is the most popular mode for authenticated encryption (and for which there is a wycheproof test).
All in all, we can start with CTR, CBC, GCM for which we know of Sui applications already

src/aes.rs
```diff
@@ -0,0 +1,227 @@
...
+            .unwrap();
+        Ok(())
+    }
+}
+
+///
+/// Aes128 in GCM mode (Authenticated)
+///
+pub struct Aes128Gcm<'a, NonceSize: ArrayLength<u8>> {
+    pub iv: &'a GenericArray<u8, NonceSize>,
```
kchalkias, line 155: qq: how about having a similar `iv` field type (bytearray), to have consistency with the other schemes?

src/traits.rs
```diff
@@ -191,3 +191,42 @@ pub trait AggregateAuthenticator:
...
+    ToFromBytes + 'static + Serialize + DeserializeOwned + Send + Sync
+{
+    fn generate<R: CryptoRng + RngCore>(rng: &mut R) -> Self;
+}
+
+/// Trait impl'd by symmetric ciphers.
+///
+pub trait Cipher {
+    /// Encrypt `plaintext` and write result to `buffer` using the given IV
+    fn encrypt(&self, iv: &[u8], plaintext: &[u8], buffer: &mut [u8]) -> Result<(), Error>;
```
kchalkias, line 207: Wondering if we should hide this from the end-user `buffer: &mut [u8])`, by experiencer it's very common to make mistakes here or be inconvenient, ie by providing wrong size input buffers, even the fact to be prepared to have an existing buffer already. Wdyt?

src/aes.rs
```diff
@@ -0,0 +1,267 @@
...
+    Aes: aes::cipher::KeySizeUser
+        + aes::cipher::KeyInit
+        + aes::cipher::BlockCipher
+        + aes::cipher::BlockSizeUser<BlockSize = U16>
+        + aes::cipher::BlockEncrypt
+        + aes::cipher::BlockDecrypt,
+{
+    fn encrypt(
+        &self,
+        iv: &[u8],
```
kchalkias, line 82: A few questions:
- do we already check iv size mismatch?
- we handle the AES `Key` as a special struct, what about the `IV`? maybe for consistency we should as well. Not strongly opinionated on that.
- For all of the encrypt/decrypt and their authenticated variants, let's have some tests on input [u8] sizes mismatch.

src/aes.rs
```diff
@@ -0,0 +1,267 @@
...
+use digest::{crypto_common::KeyIvInit, generic_array::ArrayLength, typenum::U16};
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+
+///
+/// An encryption key for a symmetric cipher consisting of `N` bytes.
+///
+#[serde_as]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Key<const N: usize> {
```
kchalkias, line 20: Similarly to what we do for other schemes (i.e., for [ed25519](https://github.com/MystenLabs/fastcrypto/blob/main/src/ed25519.rs#L42)), let's rename this to `AesKey` to avoid any future readability confusion against future apis ie HMAC keys etc.

src/aes.rs
```diff
@@ -0,0 +1,267 @@
...
+    // This is needed to derive Serialize and Deserialize since serde does not support derive for arrays
```
kchalkias, line 21: cc: @huitseeker to sign off the idiomatic way

src/traits.rs
jonas-lj: That makes sense. I just copied the interface from rustcrypto, assuming that was the common way to do it in rust. I think we should return a new array with the result then.

src/aes.rs
huitseeker: Yep, this is good.

huitseeker (commented):
Just food for thought: there may be a way to express a size floor on `Key<N>` by requiring a `MinSlice<_; 16>` or some such appropriate projection to a slice of a minimum length ([ref](https://lib.rs/crates/nslice))
In the old days, before const generics, we used to do it with something like https://crates.io/crates/generic-bytes

kchalkias (commented):
I guess the only remaining feature is IV to be in its own struct, right?

kchalkias (approved):
Awesome! LGTM

kchalkias (commented):
Thanks for merging! We generally prefer to squash and merge to avoid inflating our commit log and have all `main` commits in a stable state just in case we ever need to cherry pick, go back to specific commits etc.

jonas-lj:
> Thanks for merging! We generally prefer to squash and merge to avoid inflating our commit log and have all `main` commits in a stable state just in case we ever need to cherry pick, go back to specific commits etc.

That makes sense. I guess it doesn't make sense to revert the merge, because the commits will still be in the log, but I'll make sure to do that next time.

kchalkias (commented):
Yes, we're good with this PR merge, no need to revert.

PR #38, opened 2022-09-12
Add basic HMAC and HKDF functions

benr-ml:
Both use SHA256 as the underlying hash function.

#39


joyqvq (approved):

src/hmac.rs
```diff
@@ -79,3 +82,25 @@ where
...
+    hash.update(message);
+    let output: [u8; DIGEST_LEN] = hash.finalize().into_bytes().as_slice().try_into().unwrap();
+    Digest::new(output)
+}
+
+pub fn hkdf(
+    ikm: &[u8],
+    salt: &[u8],
+    info: &[u8],
+    output: &mut [u8],
```
mystenadmin, line 100: I think this is not aligned with `hmac`'s output. Generally I'd avoid providing output as input. It's very common that end-user won't know with what [u8] size to initiate it, and it's generally a a good hygiene to avoid mutable arrays for crypto outputs. TL;DR let's do the same with `hmac` where HKDF output array is initiated internally, then returned.

mystenadmin (changes requested):

src/hmac.rs
kchalkias, line 100: I think this is not aligned with `hmac`'s output. Generally I'd avoid providing output as input. It's very common that end-user won't know with what [u8] size to initiate it, and it's generally a a good hygiene to avoid mutable arrays for crypto outputs. TL;DR let's do the same with `hmac` where HKDF output array is initiated internally, then returned.

kchalkias (changes requested):

src/hmac.rs
benr-ml: sg, done

src/hmac.rs
```diff
@@ -79,3 +82,29 @@ where
...
 }
+
+////////////////////////////////////////////////////////////////////////
+/// HMAC-SHA256 based functions
+
+pub fn hmac(key: &[u8], message: &[u8]) -> Digest {
+    let mut hash = Hmac::<sha2::Sha256>::new_from_slice(key).unwrap();
+    hash.update(message);
+    let output: [u8; DIGEST_LEN] = hash.finalize().into_bytes().as_slice().try_into().unwrap();
+    Digest::new(output)
```
kchalkias, line 94: I kinda like that hmac returns a Digest object, it could of course be a byte array as in hkdf, but eventually I believe this is more consistent, assuming we do the same for simple hash function (that in an upcoming PR should also be wrapper functions in fastcrypto).

src/hmac.rs
```diff
@@ -79,3 +82,29 @@ where
     let keypair = K::from(secret_key);
     Ok(keypair)
 }
+
+////////////////////////////////////////////////////////////////////////
+/// HMAC-SHA256 based functions
+
+pub fn hmac(key: &[u8], message: &[u8]) -> Digest {
```
kchalkias, line 90: Although this is fine for now, HMAC key is not different than an AES key (re do we handle it as `[u8]` or we create a specific struct called `HmacKey` as we do in AES. I'd suggest you sync up with @jonas-lj to have a uniform solution for these.

src/hmac.rs
```diff
@@ -79,3 +82,29 @@ where
...
+    let mut hash = Hmac::<sha2::Sha256>::new_from_slice(key).unwrap();
+    hash.update(message);
+    let output: [u8; DIGEST_LEN] = hash.finalize().into_bytes().as_slice().try_into().unwrap();
+    Digest::new(output)
+}
+
+pub fn hkdf(
+    ikm: &[u8],
```
kchalkias, line 98: Ditto. The same applies to `ikm`. As this is a secret, do we handle it as a [u8] or create a `Seed` struct around it? We did something similar for signature seeds in Sui, see https://github.com/MystenLabs/sui/blob/main/crates/sui-types/src/signature_seed.rs

src/hmac.rs
benr-ml: Introduced HmacKey for this

src/hmac.rs
benr-ml: Introduced HkdfIkm for this

src/hmac.rs
```diff
@@ -72,10 +78,44 @@ where
...
 
     let keypair = K::from(secret_key);
     Ok(keypair)
 }
+
+////////////////////////////////////////////////////////////////////////
+/// HMAC-SHA256 based functions
+
+const HMAC_KEY_RECOMMENDED_LENGTH: usize = 32;
+const HKDF_KEY_RECOMMENDED_LENGTH: usize = 16;
```
kchalkias, line 93: why HKDF recommended with 16?

kchalkias (approved):

kchalkias (commented):
@benr-ml let's please rebase before merging; after recent merges there are now small conflicts.

src/hmac.rs
benr-ml: 128 bit security should be enough for hkdf

kchalkias: I would prefer setting it to 256bits. Note that this is a generic HKDF api and it depends on the app context where it will be used, but if used for key-pair derivation in a mass scale (ie deriving millions of keys), the effective security is reduced due to multi-target attacks. This is unfortunately one of the reasons that NO blockchain today (due to mnemonics being short) offer in practice 128 bits. See https://twitter.com/kostascrypto/status/1565141220115881984 -> we're preparing a related short abstract to publish in IACR.

benr-ml: done

PR #52, opened 2022-09-19
Fix publish.yml

joyqvq:


.github/workflows/publish.yml
```diff
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - ci1
```
huitseeker, line 6: Is this intentional?

huitseeker (approved):
👍 (see comment)

joyqvq:
after taking a closer look on https://crates.io/crates/cargo-semver-checks it achieves the same goal as crate-public-api by detecting api changes. it generates a temp file of the API in the current repo and against the published version. instead of committing to a json file, it will fail on the workflow if API change is detected, but the crate version is not bumped on major release. 

hence removing the public-api from CI

.github/workflows/publish.yml
joyqvq: just for testing, removed.

joyqvq:
foce landing bc the crate is published with the latest version and not editable

joyqvq:
@kchalkias good idea. i just yanked the new published crate from 1.0.0, we can revert to maybe 0.2.0?

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fastcrypto"
-version = "0.1.3"
+version = "1.0.0"
```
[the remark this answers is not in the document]
joyqvq: MAJOR version when you make incompatible API changes
MINOR version when you add functionality in a backwards compatible manner
PATCH version when you make backwards compatible bug fixes

^ following what semver CI is set to - since the API did change, we should probably increment MAJOR

Issue #48, opened 2022-09-16
We should add hash wrappers as well for all Sui/Narwhal used functions


kchalkias:
This will help to avoid surprises in Sui or Narwhal re using different sha2/sha3/keccak deps in different places. Ideally we should control every crypto algorithm in `fastcrypto`. After having these wrappers, we should update both Narwhal and Sui repos by invoking fastcrypto's api.

I was thinking that the same applies for base64, but let's wait for this, as Rust's popular base64 crate is working on the malleability fix.

jonas-lj:
@kchalkias Do we also need BLAKE2? It's used a few places in fastcrypto, but only in tests where it could be replaced with any other hash function.

jonas-lj:
The hash functions are added and there's an issue on the SUI repo to add them there (https://github.com/MystenLabs/sui/issues/4760), so could we close this issue now?

PR #63, opened 2022-09-21
Include hash functions (SHA-2, SHA-3, Keccak and SHAKE2)

jonas-lj:
Include wrappers for commonly used hash functions. Closing #48


kchalkias (approved):
Awesome! PRs should follow to update Sui and Narwhal repos by eliminating any external deps and always point to fastcrypto's hash impl. I'll create the issues.

Issue #69, opened 2022-09-26
Add unsecure/empty crypto schemes for benchmarking purposes 


kchalkias:
Inspired by this: https://github.com/MystenLabs/narwhal/pull/627#pullrequestreview-1115106962

PR #71, opened 2022-09-27
Add unsecure schemes for seemless benchmarking

jonas-lj:
Closing #69 
Depends on PR #83


kchalkias (commented):
Ideally and for easier PR {management + review}, let's split this PR into UnsecureSig and UnsecureHash.

jonas-lj:
@kchalkias Would it make sense to make the unsecure signature scheme slightly more complex by creating a signature as H(public_key || message)? I think that would make benchmarking even more seemless, since verification now fails with the wrong message and/or public key.

kchalkias:
@jonas-lj wow that's a great idea, especially when combined with the unsecure **fast** hash work. Yes I like that. I actually think it's the best first step we should take.

jonas-lj:
> @jonas-lj wow that's a great idea, especially when combined with the unsecure **fast** hash work. Yes I like that. I actually think it's the best first step we should take.

Great, I'll do that then. I'll create a PR with the dummy hash-function and then rewrite this PR afterwards.

kchalkias (approved):

PR #75, opened 2022-09-28
[small] Readable impl Debug for public keys

andll:
This diff makes readable Debug implementations for public keys. We often use `#[instrument]` that by default uses Debug implementation. Default debug implementation is extremely verbose and impossible to reason about in logs, this diff just prints base64 encoded string instead.


joyqvq (approved):
a nit comment otherwise lgtm!

joyqvq (approved):

PR #78, opened 2022-09-29
Cleanup hash

jonas-lj:
* Add docs for supported hash functions remove
* Remove Digest::from_bytes function since it was not used in the crate, and Digest should only be created by hash functions in the crate.

Should be merged after #77


joyqvq (approved):
nice!

PR #79, opened 2022-09-29
Use FastCryptoError on all schemes

jonas-lj:


src/aes.rs
```diff
@@ -108,21 +114,21 @@ where
...
     type IVType = InitializationVector<U16>;
 
-    fn encrypt(&self, iv: &Self::IVType, plaintext: &[u8]) -> Result<Vec<u8>, signature::Error> {
+    fn encrypt(&self, iv: &Self::IVType, plaintext: &[u8]) -> Result<Vec<u8>, FastCryptoError> {
         let mut buffer: Vec<u8> = vec![0; plaintext.len()];
         let mut cipher = ctr::Ctr128BE::<Aes>::new(&self.key.bytes, &iv.bytes);
         cipher
             .apply_keystream_b2b(plaintext, &mut buffer)
-            .map_err(|_| signature::Error::new())?;
+            .map_err(|_| FastCryptoError::GeneralError)?;
```
kchalkias, line 122: Wondering if we should augment the error list with more specific errors vs general when we have a known reason for throwing. It's ok to handle in an upcoming PR.

kchalkias (approved):

src/aes.rs
jonas-lj: I tried to do that where it was clear to me, what the reason was. But yes, let's go through that in another PR.

Issue #74, opened 2022-09-28
Unsafe Fast Hash support


kchalkias:
This will be required for benchmarking (similarly to empty/dummy signatures). It cannot empty (as collisions can break operations), maybe we can use the fastest non-cryptographically secure hash function for this purpose.

jonas-lj:
Based on a bit of desktop research, the fastest such non cryptographic hash function seems to be [xxHash3](https://cyan4973.github.io/xxHash/). This is based on the benchmarks they provide on their website and this [discussion](https://news.ycombinator.com/item?id=10667328). If no-one objects, I'll give xxHash a go and do some basic benchmarks to verify.

PR #83, opened 2022-10-03
Add unsecure hash function

jonas-lj:
Add a non-cryptographic hash function (Xxh3) for benchmarking. Closing #74


src/unsecure/mod.rs
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+/// This module contains an implementation of a negligible-cost (apart from some ocassional pk copying) trivial
+/// cryptographic signature scheme.
+///
+/// Warning: All schemes in this file are completely unsafe to use in production.
```
huitseeker, line 7: Besides the safety notice, I would add a directional comment indicating what we intend to use this for (benches). This would give a future auditor an easier time making heads or tails of this.

huitseeker (commented):
Same as #71: this needs at least couple of the simple remediation against accidental activation listed in https://gist.github.com/huitseeker/6f117d0157783b571c280c8462f4ab72 before merging.

src/unsecure/hash.rs
```diff
@@ -0,0 +1,57 @@
...
+    }
+}
+
+/// XXH128 hash function.
+///
+/// Warning: This is NOT a cryptographic hash function and should NOT be used in production.
+#[cfg(feature = "unsecure_schemes")]
+#[derive(Default)]
+pub struct XXH128Unsecure {
+    instance: twox_hash::xxh3::Hash128,
```
kchalkias, line 41: Wondering if we should on purpose use a scheme that outputs 32 bytes (so size  is compatible what most chains support): ie SuiAddress and Object derivation in Sui is truncating to 160 bits and having a shorter hash might cause issues.

jonas-lj: I've considered the same thing. Xxhash currently [does not support 256 bit hashes](https://cyan4973.github.io/xxHash/), and to my knowledge, none of the fast non-cryptographic hash functions does. Bit in this case, we could simply concatenate two 128 bit or pad with zeroes, because we mostly care about avoiding collisions.

jonas-lj:
> Same as #71: this needs at least couple of the simple remediation against accidental activation listed in https://gist.github.com/huitseeker/6f117d0157783b571c280c8462f4ab72 before merging.

I've added two: There's a check that debug_assertions should be set, and I've also added a "secure" feature by default, and if both "secure" and "unsecure_schemes" are enabled, there'll be a compile error.

huitseeker (approved):
> I tried to add a compile error if both features were enables, but the clippy check failed because of that, because it runs with all features enabled.

I think it's 100% fair to make clippy run on default features and possibly make it have a distinct run on `--no-default-features --features "unsecure_schemes"`. clippy running with all features is just a CI setup shortcut with no deep meaning.

Cargo.toml
```diff
@@ -54,8 +55,10 @@ name = "crypto"
 harness = false
 
 [features]
-default = []
+default = ["secure"]
+secure = []
```
huitseeker, line 59: Nit: we might want a comment here on how this is meant to control for accidental activations of `unsecure_schemes`.

src/unsecure/hash.rs
```diff
@@ -0,0 +1,87 @@
...
+    type OutputSize = typenum::U32;
+}
+
+impl HashFunction<typenum::U32> for Fast256HashUnsecure {
+    fn update(&mut self, data: &[u8]) {
+        self.instance.write(data);
+    }
+
+    fn finalize(self) -> Digest<typenum::U32> {
+        // Create a 32 byte digest consisting of two copies of a Xxh128 digest.
```
kchalkias, line 80: Thanks, that works! you could also pad it with zeros, but it's probably better this way to look like random 256bit string when printed.

kchalkias (approved):

Issue #82, opened 2022-10-01
Add benchmark for BLS operations (aggregate + verify aggregated sig)


kchalkias:
Atm we only compare batch verification between different schemes, but for BLS this is not ideal, because in practice we should compare an aggregated BLS vs batched ed25519/secp256k1

PR #84, opened 2022-10-03
Add aggregate BLS verfication to benchmarks

jonas-lj:
Closing #82 

Attached are the output of a benchmark and a plot of aggregate verification of BLS signatures vs batched verification of Ed25519 signatures.
 [benchmark.txt](https://github.com/MystenLabs/fastcrypto/files/9703761/benchmark.txt)
ED25519 batch vs BLS aggregate

### DIFF
--- a/benches/crypto.rs
+++ b/benches/crypto.rs
@@ -87,18 +87,8 @@ mod ed25519_benches {
             let blst_public_keys: Vec<_> = blst_keypairs
                 .iter()
                 .map(|key| key.public().clone())
-                .collect();
-
-            let aggregate_signature =
-                BLS12381AggregateSignature::aggregate(blst_signatures.clone()).unwrap();
-
-            c.bench_with_input(
-                BenchmarkId::new("BLS12381 aggregate verification", *size),
-                &(msg.clone(), blst_public_keys.clone(), aggregate_signature),
-                |b, i| {
-                    b.iter(|| i.2.verify(&i.1, &i.0));
-                },
-            );
+                .collect();            
+            let blst_aggregate_signature = BLS12381AggregateSignature::aggregate(blst_signatures.clone()).unwrap();
 
             c.bench_with_input(
                 BenchmarkId::new("Ed25519 batch verification", *size),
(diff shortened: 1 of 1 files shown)


benches/crypto.rs
```diff
@@ -89,6 +89,17 @@ mod ed25519_benches {
...
                 .collect();
 
+            let aggregate_signature =
+                BLS12381AggregateSignature::aggregate(blst_signatures.clone()).unwrap();
+
+            c.bench_with_input(
+                BenchmarkId::new("BLS12381 aggregate verification", *size),
+                &(msg.clone(), blst_public_keys.clone(), aggregate_signature),
+                |b, i| {
+                    b.iter(|| i.2.verify(&i.1, &i.0));
```
huitseeker, line 99: It would look a lot nicer if you destructured the `i` triplet in the bound variable pattern:
```
|b, (msg, keys, sig)| {
```

benches/crypto.rs
```diff
@@ -89,6 +89,17 @@ mod ed25519_benches {
                 .map(|key| key.public().clone())
                 .collect();
 
+            let aggregate_signature =
+                BLS12381AggregateSignature::aggregate(blst_signatures.clone()).unwrap();
```
huitseeker, line 93: Note to self: does `aggregate` *really* needs to consume the individual signatures, I wonder?

huitseeker (commented):
While I have you here, the bench is still named `ed25519_benches` overall, which is nowhere near its functionality any more. Would it be possible to rename?

benches/crypto.rs
```diff
@@ -89,6 +89,17 @@ mod ed25519_benches {
                 .map(|key| key.public().clone())
                 .collect();
 
+            let aggregate_signature =
```
kchalkias, line 92: Also, please add this at the end of this benchmark (not as a the first one) so we have Ed, BLS_batching_no_aggregation, BLS_aggregated

kchalkias (commented):
We could also add a benchmark group (independent from batch verification) to measure the cost of aggregating (not verifying). We might realize that we need to optimize that part too (especially on very large sets of signers (hundreds, thousands etc). This curve's operations are slower than ed25519, and there might be optimization opportunities for the aggregator too. It can be a different PR.

benches/crypto.rs
```diff
@@ -98,11 +100,26 @@ mod ed25519_benches {
...
+                    blst_signatures.clone(),
+                ),
                 |b, i| {
                     b.iter(|| VerifyingKey::verify_batch_empty_fail(&i.0, &i.1[..], &i.2[..]));
                 },
             );
+            c.bench_with_input(
+                BenchmarkId::new("BLS12381 aggregate verification", *size),
+                &(
+                    msg.clone(),
```
kchalkias, line 115: do we need to clone on the last use in the function?

jonas-lj: No, its just left over from when it was done before the other benchmarks. I'll remove it.

kchalkias (approved):
LGTM

kchalkias:
Can you also attach in the description a printed output of the benchmarks results to have it as a reference in this PR? We sometimes need to use a link whenever required to explain the differences. You can even do after merging.

benches/crypto.rs
jonas-lj: It doesn't seem like it. I've created a new PR, #90, to fix this

PR #90, opened 2022-10-04
Borrow list of signatures in signature aggregation

jonas-lj:


kchalkias (approved):

kchalkias:
Plz ensure after any fastcrypto version update to make the required changes to Sui / Narwhal

huitseeker (commented):
I like the direction where this is going but ... have you had a stab at adjusting the Sui code base for this API? Not that this is wrong in any way, but it's a lot of work. Moreover, since a slice is a very particular form of iterator (each element must be adjacent to the others), you may notice a couple of places where the sig vector is custom made for the purpose of calling aggregate. So with your PR, we custom-make a vector and ... immediately take a ref of it, and then drop the vector.

But the reason we spend so much time thinking about a copy of the `Vec` is because the `Vec` actually contains the actual `Sig`s If the Vec contained references, the copy of the top-level `Vec` / `Iterator` isn't that important any more.

One thing we might want to do here is extend the API to work with (nearly) any collection of things that resolve to a pointer to a signature, whatever this pointer happens to be. 

As far as these "generic pointers" are concerned, there is a trait that embodies just such a requirement: `Borrow<T>`, which is implemented by things that `.borrow()` to a `&T`. The [documentation](https://doc.rust-lang.org/std/borrow/trait.Borrow.html#examples) captures this in more detail.

There is also a requirement to ask for a collection of those things. Rather than haggle over slice or vector, considering the very first thing that we do in `fn aggregate` is extract an iterator of those things, why not ... require the ability to use an iterator in the first place? In Rust, that's the `IntoIterator` trait.

To combine the two and demonstrate the versatility of the approach, I've written an example in the playground, emulating signatures with `String`: 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=65df6b90d09795dd7a4a394a50f0b848
Note the trait `MyTrait` is superfluous in this approach, I just defined it to show what I wanted was the ability to access a method on `String`, i.e. that my goal was some `Borrow<String>`, which should be made clear by me forcing the first type in every invocation of `from_borrow`.

TL;DR: 
You're switching the signature of aggregate from
```rust
    fn aggregate(signatures: Vec<Self::Sig>) -> Result<Self, FastCryptoError>;
```
To:
```rust
    fn aggregate(signatures: &[Self::Sig]) -> Result<Self, FastCryptoError>;
```

I think you should use something close to this (modulo taking the `signature` argument by reference if you'd like):
```rust
    fn aggregate<K: Borrow<Self::Sig>, I: IntoIterator<Item = K>>(signatures: I) -> Result<Self, FastCryptoError>
```

jonas-lj:
> I like the direction where this is going but ... have you had a stab at adjusting the Sui code base for this API? Not that this is wrong in any way, but it's a lot of work. Moreover, since a slice is a very particular form of iterator (each element must be adjacent to the others in memory), you may notice a couple of places where the sig vector is custom made for the purpose of calling aggregate. So with your PR, we custom-make a vector and ... immediately take a ref of it, and then drop the vector.
> 
> But at a high level, the reason we spend so much time thinking about a copy of the `Vec` is because the `Vec` actually contains the actual `Sig`s If the Vec contained references, the copy of the top-level `Vec` / `Iterator` isn't that important any more.
> 
> One thing we might want to do here is extend the API to work with (nearly) any collection of things that resolve to a pointer to a signature, whatever this pointer happens to be.
> 
> As far as these "generic pointers" are concerned, there is a trait that embodies just such a requirement: `Borrow<T>`, which is implemented by things that `.borrow()` to a `&T`. The [documentation](https://doc.rust-lang.org/std/borrow/trait.Borrow.html#examples) captures this in more detail.
> 
> There is also a requirement to ask for a collection of those things. Rather than haggle over slice or vector, considering the very first thing that we do in `fn aggregate` is extract an iterator of those things, why not ... require the ability to use an iterator in the first place? In Rust, that's the `IntoIterator` trait (which is implemented by `Vec`).
> 
> To combine the two and demonstrate the versatility of the approach, I've written an example in the playground, emulating signatures with `String`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=65df6b90d09795dd7a4a394a50f0b848 Note the trait `MyTrait` is superfluous in this approach, I just defined it to show what I wanted was the ability to access a method on `String`, i.e. that my goal was some `Borrow<String>`, which should be made clear by me forcing the first type in every invocation of `from_borrow`.
> 
> TL;DR: You're switching the signature of aggregate from
> 
> ```rust
>     fn aggregate(signatures: Vec<Self::Sig>) -> Result<Self, FastCryptoError>;
> ```
> 
> To:
> 
> ```rust
>     fn aggregate(signatures: &[Self::Sig]) -> Result<Self, FastCryptoError>;
> ```
> 
> I think you should use something close to this:
> 
> ```rust
>     fn aggregate<K: Borrow<Self::Sig>, I: IntoIterator<Item = K>>(signatures: &I) -> Result<Self, FastCryptoError>
> ```
> 
> PS: I'm not married to the more complex function signature, something like
> 
> ```rust
>     fn aggregate<I: IntoIterator<Item = Self::Sig>>(signatures: &I) -> Result<Self, FastCryptoError>
> ```
> 
> is already an improvement on flexibility compared to a vector.



> I like the direction where this is going but ... have you had a stab at adjusting the Sui code base for this API? Not that this is wrong in any way, but it's a lot of work. Moreover, since a slice is a very particular form of iterator (each element must be adjacent to the others in memory), you may notice a couple of places where the sig vector is custom made for the purpose of calling aggregate. So with your PR, we custom-make a vector and ... immediately take a ref of it, and then drop the vector.
> 
> But at a high level, the reason we spend so much time thinking about a copy of the `Vec` is because the `Vec` actually contains the actual `Sig`s If the Vec contained references, the copy of the top-level `Vec` / `Iterator` isn't that important any more.
> 
> One thing we might want to do here is extend the API to work with (nearly) any collection of things that resolve to a pointer to a signature, whatever this pointer happens to be.
> 
> As far as these "generic pointers" are concerned, there is a trait that embodies just such a requirement: `Borrow<T>`, which is implemented by things that `.borrow()` to a `&T`. The [documentation](https://doc.rust-lang.org/std/borrow/trait.Borrow.html#examples) captures this in more detail.
> 
> There is also a requirement to ask for a collection of those things. Rather than haggle over slice or vector, considering the very first thing that we do in `fn aggregate` is extract an iterator of those things, why not ... require the ability to use an iterator in the first place? In Rust, that's the `IntoIterator` trait (which is implemented by `Vec`).
> 
> To combine the two and demonstrate the versatility of the approach, I've written an example in the playground, emulating signatures with `String`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=65df6b90d09795dd7a4a394a50f0b848 Note the trait `MyTrait` is superfluous in this approach, I just defined it to show what I wanted was the ability to access a method on `String`, i.e. that my goal was some `Borrow<String>`, which should be made clear by me forcing the first type in every invocation of `from_borrow`.
> 
> TL;DR: You're switching the signature of aggregate from
> 
> ```rust
>     fn aggregate(signatures: Vec<Self::Sig>) -> Result<Self, FastCryptoError>;
> ```
> 
> To:
> 
> ```rust
>     fn aggregate(signatures: &[Self::Sig]) -> Result<Self, FastCryptoError>;
> ```
> 
> I think you should use something close to this:
> 
> ```rust
>     fn aggregate<K: Borrow<Self::Sig>, I: IntoIterator<Item = K>>(signatures: &I) -> Result<Self, FastCryptoError>
> ```
> 
> PS: I'm not married to the more complex function signature, something like
> 
> ```rust
>     fn aggregate<I: IntoIterator<Item = Self::Sig>>(signatures: &I) -> Result<Self, FastCryptoError>
> ```
> 
> is already an improvement on flexibility compared to a vector.

I ran into a bit of trouble implementing and ended up having to include a lifetime in the signature and ensure that the iterator is over references: 
```rust
fn aggregate<'a, K: Borrow<Self::Sig> + 'a, I: IntoIterator<Item = &'a K>>(signatures: I)
```
because we need a slice of references in the underlying BLST crate. But this achieves the same goals as you point out, namely that it doesn't care what kind of collection it is given, nor whether the elements in it are references.

huitseeker (approved):
Yup, blst is a bit of a pain this way. Thanks @jonas-lj !

kchalkias (commented):
Awesome, thank you both (@huitseeker and @jonas-lj) for the pointers + solution. Wondering if the same pattern is applicable elsewhere across our lib as well (ie batching).

PR #93, opened 2022-10-05
Add test for BLS signature aggregation

jonas-lj:
Closing #91


huitseeker (approved):
Thanks @jonas-lj !
Nit: This could easily be converted to a [proptest](https://docs.rs/proptest/latest/proptest/).

PR #95, opened 2022-10-07
feat: re-organize the crates into a workspace

huitseeker:
Also switch to rust 1.64.0.

As discussed offline.


joyqvq:
did we ensure the publish github action still works? I used to exclude fastcrypto-derive in the publish, maybe good time to revive. 

while you are at it, maybe add a readme to fastcrypto-derive as well?

huitseeker:
Adding a Readme to the `fastcrypto-derive` crate would require splitting the top-level readme into one readme for each of fastcrypto and fastcrypto-derive, which may be a bit premature if we don't know how we're going to split our crates yet (see offline discussion). But I've made both crates link to the toplevel README, so that this information is pushed to crate.io

I've added fatcrypto-derive to the publih job. The main semantic difference over the job in mysten-infra (which successfully published mysten-network yesterday) is the following:
```
@@ -79,8 +73,8 @@
           git push origin "${{ matrix.package}}-v$CURRENT_VERSION"
       - uses: actions-rs/toolchain@v1
         if: steps.check.outputs.is_new_version == 'yes'
-      - name: Publish packages
+      - name: Login
+        run: cargo login ${{ secrets.CARGO_REGISTRY_TOKEN }}
+      - name: Publish fastcrypto
         if: steps.check.outputs.is_new_version == 'yes'
-        env:
-          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
-        run: cargo publish -p ${{ matrix.package }}
+        run: cargo publish --token ${{ secrets.CARGO_REGISTRY_TOKEN }}
```
I think those are equivalent.

Cargo.toml
```diff
@@ -1,72 +1,7 @@
...
-license = "Apache-2.0"
-authors = ["Mysten Labs <build@mystenlabs.com>"]
-edition = "2021"
-publish = ["crates-io"]
-description = "Common cryptographic library used at Mysten Labs"
-repository = "https://github.com/MystenLabs/fastcrypto"
+[workspace]
+resolver = "2"
+members = [
+    "fastcrypto",
```
kchalkias, line 4: so in the future all individual crates (fastcrypto-signature, fastcrypto-hash, fastcrypto-bulletproofs etc) will be extra members in the package?

joyqvq (approved):
lgtm!

Cargo.toml
huitseeker: That is how rust [workspaces](https://doc.rust-lang.org/cargo/reference/workspaces.html) work.

kchalkias: yeah my question is if our intention in the future is to retire `fastcrypto` and split it to individual crates that will live in a flat list along with `fastcrypto-derive`

PR #96, opened 2022-10-07
feat: Add a proving crate + benchmarks

huitseeker:
Adds basic split-up benchmarks with Arkworks-Groth16 on BLS12_381.

Opening early to set the crate layout.


fastcrypto-proving/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+repository = "https://github.com/MystenLabs/fastcrypto"
+
+[[bench]]
+name = "proving"
+harness = false
+
+[dependencies]
+
+[dev-dependencies]
+ark-bls12-381 = "0.3.0"
```
kchalkias, line 18: I guess this is a temp (initial) backend for benchmarking purposes, until we end up with blst backend?

fastcrypto-proving/benches/proving.rs
```diff
@@ -0,0 +1,149 @@
...
+        let a = cs.new_witness_variable(|| self.a.ok_or(SynthesisError::AssignmentMissing))?;
+        let b = cs.new_witness_variable(|| self.b.ok_or(SynthesisError::AssignmentMissing))?;
+        let c = cs.new_input_variable(|| {
+            let a = self.a.ok_or(SynthesisError::AssignmentMissing)?;
+            let b = self.b.ok_or(SynthesisError::AssignmentMissing)?;
+
+            Ok(a * b)
+        })?;
+
+        for _ in 0..(self.num_variables - 3) {
```
kchalkias, line 36: explain the requirement for `-3`

fastcrypto-proving/benches/proving.rs
```diff
@@ -0,0 +1,149 @@
...
+#[derive(Copy, Clone)]
+struct DummyCircuit<F: PrimeField> {
+    pub a: Option<F>,
+    pub b: Option<F>,
+    pub num_variables: usize,
+    pub num_constraints: usize,
+}
+
+impl<F: PrimeField> ConstraintSynthesizer<F> for DummyCircuit<F> {
+    fn generate_constraints(self, cs: ConstraintSystemRef<F>) -> Result<(), SynthesisError> {
```
kchalkias, line 26: I'd suggest at least a simple doc comment re the dummy circuit. TL;DR does it prove that "I know two numbers with whose product is X"? (in which case a, b are witnesses and X (their product) a public input)

fastcrypto-proving/benches/proving.rs
```diff
@@ -0,0 +1,149 @@
...
+) {
+    static VARIABLES: [usize; 5] = [16, 32, 64, 128, 256];
+
+    for size in VARIABLES.iter() {
+        let rng = &mut ark_std::test_rng();
+        let c = DummyCircuit::<F> {
+            a: Some(<F>::rand(rng)),
+            b: Some(<F>::rand(rng)),
+            num_variables: *size,
+            num_constraints: 65536,
```
kchalkias, line 61: how did you come up to this number?

fastcrypto-proving/Cargo.toml
huitseeker: It is, though I expect we will keep arkworks around for a while, for unit-testing purposes.

fastcrypto-proving/benches/proving.rs
huitseeker: a, b, c are defined right above

fastcrypto-proving/benches/proving.rs
huitseeker: I've actually refactored the test to focus on the # of constraints, which the proof timing scales with. The number of variables makes little sense considering we only provide one input.

Cargo.toml
```diff
@@ -2,6 +2,7 @@
 resolver = "2"
 members = [
     "fastcrypto",
-    "fastcrypto-derive"
+    "fastcrypto-derive",
+    "fastcrypto-proving"
```
kchalkias, line 6: I'd rename this to fastcrypto-zkproving or even better fastcrypto-zkp

kchalkias (dismissed):
Approving, but still prefer to name the crate fastcrypto-zkp, as we plan to have fastcrypto-signature, fastcrypto-hash etc

joyqvq (approved):

Issue #97, opened 2022-10-08
[signatures] implement `verify_batch_empty_fail_different_msg`


kchalkias:
Atm we only support batch verification of signatures over the same message, we should add functionality to batch verify signatures over different messages too. In practice we only expect impact on BLS timings, as ed25519 batch verification does not care if the message is the same or not.

PR #104, opened 2022-10-10
Add verify_batch_empty_fail_different_msg

jonas-lj:
Closing #97


jonas-lj:
In this implementation, there's no error if the number of messages and pks/signatures does not match because it's using zip and the messages are represented by an iterator. Using zipeq will cause a panic if they do not match, but I'm not sure we want this either. Alternatively, the messages can be represented by a slice of slices, but I like the solution with an iterator better as it's more versatile.

fastcrypto/src/traits.rs
```diff
@@ -110,6 +110,24 @@ pub trait VerifyingKey:
...
+    {
+        if sigs.is_empty() {
+            return Err(eyre!("Critical Error! This behaviour can signal something dangerous, and that someone may be trying to bypass signature verification through providing empty batches."));
+        }
+        if pks.len() != sigs.len() {
+            return Err(eyre!("Mismatch between number of signatures and public keys provided"));
+        }
+        pks.iter()
+            .zip(sigs)
+            .zip(msgs.into_iter())
```
jonas-lj, line 127: Note that there's no error if the number of messages and signatures/keys do not match. Is this an issue?

fastcrypto/src/tests/bls12381_tests.rs
```diff
@@ -102,6 +102,26 @@ fn signature_test_inputs() -> (Vec<u8>, Vec<BLS12381PublicKey>, Vec<BLS12381Sign
...
+        .zip(&digests)
+        .map(|(kp, digest)| {
+            let sig = kp.sign(digest.as_ref());
+            (kp.public().clone(), sig)
+        })
+        .unzip();
+
+    (digests, pubkeys, signatures)
+}
+
```
huitseeker, line 124: Wouldn't this work?

```rust
fn signature_test_inputs_different_msg<S: Authenticator<PublicKey = PK>, PK: VerifyingKey<Sig = S>>
) -> (Vec<Vec<u8>>, Vec<PK>, Vec<S>) {
    // Make signatures.
    let digests: Vec<Vec<u8>> = [b"Hello", b"world", b"!!!!!"]
        .iter()
        .map(|msg| hash::Sha256::digest(*msg).to_vec())
        .collect();
    let (pubkeys, signatures): (Vec<_>, Vec<_>) = keys()
        .into_iter()
        .take(3)
        .zip(&digests)
        .map(|(kp, digest)| {
            let sig = kp.sign(digest.as_ref());
            (kp.public().clone(), sig)
        })
        .unzip();

    (digests, pubkeys, signatures)
}
```

huitseeker (commented):
The tests look good, but perhaps the fixture can be shared between schemes by writing it generically?

fastcrypto/src/tests/bls12381_tests.rs
jonas-lj: That's a good idea. I think it could also work for many of the tests in BLS, ED25519 and secs256k1.

kchalkias: Are we going to apply @huitseeker's suggestion in this PR?

jonas-lj: Yea, I'll do that. Should we create a common_tests file for something like that?

kchalkias: signature_tests for now

jonas-lj: All the rebases made it hard to see what was actually added, so I've created a new PR, #119, and will close this one.

PR #109, opened 2022-10-11
Prepare for update of fastcrypto version in sui

jonas-lj:
This PR includes various updates to fastcrypto::hash to make it possible to update SUI's version of fastcrypto to the most recent. Specifically, in sui there are multiple places where digests are defined for specific contexts. fastcrypto::hash currently makes that cumbersome to do, since a hash function always had Digest<DigestLength> as output. In this PR we fix this to make adoption in sui/narwhal easier by assuming that a Digest is always 32 bytes.

When this is merged and the fastcrypto version is updated in mysten-util-mem to ensure that MallocSizeOf is implied for various structs in fastcrypto, we can merge https://github.com/MystenLabs/sui/pull/5147 which bumps fastcrypto to the most recent version.

There also seems to be a conflict between the version of tokio in fastcrypto (1.21.2) and sui (1.20.1) that also needs to be resolved.


fastcrypto/src/hash.rs
```diff
@@ -2,94 +2,109 @@
...
-use digest::OutputSizeUser;
-use generic_array::{ArrayLength, GenericArray};
 use serde::{Deserialize, Serialize};
 use std::fmt;
 
-/// Represents a hash digest of `DigestLength` bytes.
-#[derive(Hash, PartialEq, Eq, Clone, Deserialize, Serialize, Ord, PartialOrd)]
-pub struct Digest<DigestLength: ArrayLength<u8> + 'static>(pub GenericArray<u8, DigestLength>);
+/// The length of a digest.
+pub const DIGEST_LEN: usize = 32;
```
kchalkias, line 9: Is there a reason to PIN the length? What if in the future we support 384 or 512?

jonas-lj: It was possible in the previous change of the hash functions, but sui depended quite a bit on the length being constant. I'll gladly give it a go making the length variable and then put the constant in sui, but I think we should consider keeping it constant to make the merge into sui easier, and then add an issue about making the length variable.

jonas-lj: Actually it wasn't too bad, so I've reintroduced the variable digest length but this time as a constant instead of a typenum. It adds a bit of redundancy a few places, because we need the size as both types when calling the underlying crates, but I managed to keep it inside fastcrypto such that only the constant is exposed.

fastcrypto/Cargo.toml
```diff
@@ -22,7 +22,7 @@ serde = { version = "1.0.145", features = ["derive"] }
 serde_bytes = "0.11.7"
 serde_with = "2.0.1"
 signature = { version = "1.6.4", features = ["rand-preview"] }
-tokio = { version = "1.21.2", features = ["sync", "rt", "macros"] }
+tokio = { version = "1.21.1", features = ["sync", "rt", "macros"] }
```
jonas-lj, line 25: sui uses version 1.20.1 which has some features that conflict with 1.21.2, so we need degrade here to avoid a conflict until the version is upgraded in sui.

fastcrypto/src/hash.rs
```diff
@@ -2,126 +2,151 @@
...
-    type OutputSize = Variant::OutputSize;
+    fn digest(&self) -> Self::TypedDigest;
 }
 
-impl<Variant: digest::Digest + 'static + Default> HashFunction<Variant::OutputSize>
-    for HashFunctionWrapper<Variant>
+/// This wraps a `digest::Digest` as a `fastcrypto::hash::HashFunction`.
+#[derive(Default)]
+pub struct HashFunctionWrapper<Variant: digest::Digest + 'static, const DIGEST_LEN: usize>(Variant);
+
```
jonas-lj, line 106: We need both the `Variant`type and the `DIGEST_LEN`constant because the digest length in the underlying crates is defined as a type but we use a constant. This struct should only be used internally though.

joyqvq (approved):
lgtm!

PR #110, opened 2022-10-11
Type conversions between arkworks and blst

huitseeker:
This PR follows #96, which should be merged before it. 

The conversions will allow systematic conversions arkworks <-> blst to check operations match.

### DIFF
--- a/fastcrypto-zkp/src/conversions.rs
+++ b/fastcrypto-zkp/src/conversions.rs
@@ -1,0 +1,461 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+use ark_ff::{Fp384, FromBytes, PrimeField};
+use ark_serialize::{CanonicalSerialize, CanonicalSerializeWithFlags, EmptyFlags};
+use blst::{blst_fp, blst_fp12, blst_fp6, blst_fp_from_lendian, blst_p1_affine};
+use blst::{blst_fp2, blst_p1_deserialize};
+use blst::{blst_p1_affine_serialize, blst_uint64_from_fp};
+use blst::{blst_p2_affine, blst_p2_affine_serialize, blst_p2_deserialize, BLST_ERROR};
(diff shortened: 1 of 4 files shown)


kchalkias (commented):
Love the approach + round trip tests. Is there any special representation for edge cases? Ie zero elements that needs different handling?

huitseeker:
> Is there any special representation for edge cases? Ie zero elements that needs different handling?

Yes, essentially the infinity element for G1 & G2. 

Where I was going with this PR, considering the arkworks types are more used for testing than anything else (and that the arkworks serialization types are "wrong" in the sense that they do not follow the zcash spec), was to land the non-degenerate cases, to have an e2e test for Groth16, and then extend those conversion functions with the infinity representations.

fastcrypto-zkp/src/conversions.rs
```diff
@@ -0,0 +1,461 @@
...
+
+/// Affine point translations: those mostly allow us to receive the
+/// proof points, provided in affine form.
+pub fn bls_g1_affine_to_blst_g1_affine(pt: &BlsG1Affine) -> blst_p1_affine {
+    debug_assert_eq!(pt.uncompressed_size(), G1_UNCOMPRESSED_SIZE);
+    let tmp_p1 = blst_p1_affine {
+        x: bls_fq_to_blst_fp(pt.x),
+        y: bls_fq_to_blst_fp(pt.y),
+    };
+    // See https://github.com/arkworks-rs/curves/issues/14 for why the double serialize
```
kchalkias, line 136: +1

kchalkias: It would be great to have a a bulletpoint list of what we offer to the community after this epic. BLST backend, better timings, fixes / extra safety over existing applications (correct masks etc), better alignment with standards etc. This can be part of a readme subsection re importance and promising results of our implementation. It can be a follow up PR.

kchalkias (approved):
Fantastic!

PR #115, opened 2022-10-14
Set tokio to same version as sui

jonas-lj:


huitseeker:
Curious: why is that necessary? (as opposed to updating tokio in Sui)

PR #119, opened 2022-10-15
Add verify_batch_empty_fail_different_msg

jonas-lj:
Closing #97 
Closing #98 

Note that this PR contains the same changes as #104 and #107.

### DIFF
--- a/fastcrypto/src/bls12381.rs
+++ b/fastcrypto/src/bls12381.rs
@@ -206,7 +206,19 @@ impl VerifyingKey for BLS12381PublicKey {
         pks: &[Self],
         sigs: &[Self::Sig],
     ) -> Result<(), eyre::Report> {
-        let num_sigs = sigs.len();
+        // TODO: fix this, the identical message opens up a rogue key attack
+        let msgs_refs = (0..sigs.len()).map(|_| msg).collect::<Vec<_>>();
+        Self::verify_batch_empty_fail_different_msg(&msgs_refs, pks, sigs)
+    }
+
+    fn verify_batch_empty_fail_different_msg<'a, M>(
+        msgs: &[M],
+        pks: &[Self],
+        sigs: &[Self::Sig],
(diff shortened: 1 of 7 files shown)


fastcrypto/src/bls12381.rs
```diff
@@ -206,23 +206,35 @@ impl VerifyingKey for BLS12381PublicKey {
...
-                "Mismatch between number of signatures and public keys provided"
+                "Mismatch between number of messages, signatures and public keys provided"
             ));
         }
-        let mut rands: Vec<blst_scalar> = Vec::with_capacity(num_sigs);
+        let mut rands: Vec<blst_scalar> = Vec::with_capacity(sigs.len());
         let mut rng = OsRng;
 
-        for _ in 0..num_sigs {
+        for _ in 0..sigs.len() {
```
kchalkias, line 237: Unrelated to this PR, but we need a TODO fix here as the rand scalars should be at least 128bits (thus we should set vals[1] as well). Ideally I'd create a private function get_128bit_scalar and invoke that.

fastcrypto/src/tests/signature_tests.rs
```diff
@@ -0,0 +1,44 @@
...
+
+use rand::{rngs::StdRng, SeedableRng};
+
+use crate::{
+    hash::{HashFunction, Sha256},
+    traits::KeyPair,
+};
+
+/// Generate `n` random key pairs.
+pub fn keys<KP: KeyPair>(n: usize) -> Vec<KP> {
```
kchalkias, line 12: ah, these are util functions, not tests, so we could rename the file to signature_test_utils.rs

kchalkias (approved):
Some minor comments but logic wise we are golden! Thanks

jonas-lj:
> Some minor comments but logic wise we are golden! Thanks

Great! Thanks for the review. I'll merge and add issues based on your comments.

PR #126, opened 2022-10-17
Add verify_different_msg to aggregate signatures

jonas-lj:
Closing #98

### DIFF
--- a/fastcrypto/src/ed25519.rs
+++ b/fastcrypto/src/ed25519.rs
@@ -444,6 +444,26 @@ impl AggregateAuthenticator for Ed25519AggregateSignature {
             .map_err(|_| FastCryptoError::GeneralError)
     }
 
+    fn verify_different_msg(
+        &self,
+        pks: &[<Self::Sig as Authenticator>::PubKey],
+        messages: &[&[u8]],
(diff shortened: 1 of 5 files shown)


fastcrypto/src/ed25519.rs
```diff
@@ -444,6 +444,26 @@ impl AggregateAuthenticator for Ed25519AggregateSignature {
             .map_err(|_| FastCryptoError::GeneralError)
     }
 
+    fn verify_different_msg(
```
kchalkias, line 447: what's the difference between this and `batch_verify`?

jonas-lj: As I understand it, `batch_verify`verifies a batch of aggregate signatures, where each aggregate signature is on the same message, but `verify_different_msg`verifies a single aggregate signature different messages.

kchalkias: Ok will approve now and we can work on the ideal api later.

kchalkias (approved):

fastcrypto/src/ed25519.rs
jonas-lj: Great. To keep consistency, we might also want to add `batch_verify_different_msg`?

PR #136, opened 2022-10-20
Move BLS12377 from Sui

jonas-lj:
This PR copies the BLS12377 signature scheme (almost) verbatim from Sui.


kchalkias (approved):
Awesome! 2 follow up actions:
a) update narwhal if required
b) created an issue to add the curve in our benchmarks: https://github.com/MystenLabs/fastcrypto/issues/137

PR #138, opened 2022-10-20
[test-only] Bls12381 proptests

huitseeker:
This is intended as a simple set of proptests to get a bit of progress on issues like the remainders of #12 (if any) and #102. LMK what you think.


fastcrypto/src/traits.rs
```diff
@@ -190,7 +190,7 @@ pub trait AggregateAuthenticator:
     type PubKey: VerifyingKey<Sig = Self::Sig>;
     type PrivKey: SigningKey<Sig = Self::Sig>;
 
-    /// Parse a key from its byte representation
+    /// Combine signatures into a single aggregated signature.
```
kchalkias, line 193: thanks

fastcrypto/src/tests/bls12381_tests.rs
```diff
@@ -541,3 +543,134 @@ fn test_signature_aggregation() {
...
+            let pk: BLS12381PublicKey = kp.public().clone();
+            (Just(pk), maybe_valid_sig(kp.private()))
+        })
+        .prop_flat_map(|(pk, (msg, sig))| (Just(pk), Just(msg), Just(sig)))
+        .no_shrink()
+}
+
+const BLS_MAX_SIGNATURES: usize = 100;
+
+fn aggregate_treewise(sigs: &[BLS12381Signature]) -> BLS12381AggregateSignature {
```
kchalkias, line 592: Interesting, when is this useful, to ensure that different aggregation strategies are aligned?

kchalkias (approved):

fastcrypto/src/tests/bls12381_tests.rs
huitseeker: Oh, I just wanted to not limit my testing to `add_signature`. Eventually this will be useful in tests exercising https://eprint.iacr.org/2021/377.pdf section 3.2.1.

PR #141, opened 2022-10-21
Update unsecure schemes

jonas-lj:


fastcrypto/src/unsecure/signature.rs
```diff
@@ -426,6 +430,14 @@ impl AggregateAuthenticator for UnsecureAggregateSignature {
         }
         Ok(())
     }
+
+    fn verify_different_msg(
+        &self,
+        _pks: &[<Self::Sig as Authenticator>::PubKey],
+        _messages: &[&[u8]],
+    ) -> Result<(), FastCryptoError> where {
+        todo!()
```
kchalkias, line 439: Are we ok with this? does Sui invokes this function? We'll soon need to run a cluster test using the unsafe schemes to measure the impact of cryptography in execution % cost

jonas-lj: No, this should be implemented. I'll fix that right away.

kchalkias (approved):
LGTM

PR #143, opened 2022-10-21
Impl verify_different_msg for bls12377

jonas-lj:
The ark-bls12-377 create doesn't seem to have a good function to verify aggregate signatures over different messages, so this PR implements such a function using functions from ark-bls12-377 to compute pairings and products. Closing #142

### DIFF
--- a/fastcrypto/src/bls12377.rs
+++ b/fastcrypto/src/bls12377.rs
@@ -662,7 +662,7 @@ impl AggregateAuthenticator for BLS12377AggregateSignature {
         }
 
         // TODO: The ark-bls12-377 create doesn't have a good function to verify aggregate signatures over different messages,
-        // but it it eventually does expose such a function, we should use that instead of the implementation below.
+        // but if it eventually does expose such a function, we should use that instead of the implementation below.
         let mut pairs: Vec<_> = pks
             .iter()
             .zip(messages)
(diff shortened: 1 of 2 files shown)


kchalkias (commented):
I see, it's fine as a first step, lets please add a TODO comment not to forget to reconsider the currently naive solution.

jonas-lj:
> I see, it's fine as a first step, lets please add a TODO comment not to forget to reconsider the currently naive solution.

I don't think it's a naive solution, since it actually implements the aggregate verification (not just batched) and uses the optimised functions for products of pairings from the crate. So an implementation inside the crate would probably be very similar. But I'll add a todo that we should only use this until the crate eventually exposes a similar function.

fastcrypto/src/bls12377.rs
```diff
@@ -651,10 +652,42 @@ impl AggregateAuthenticator for BLS12377AggregateSignature {
...
+        messages: &[&[u8]],
+    ) -> Result<(), crate::error::FastCryptoError> {
+        if pks.len() != messages.len() {
+            return Err(crate::error::FastCryptoError::InputLengthWrong(
+                messages.len(),
+            ));
+        }
+
+        // TODO: The ark-bls12-377 create doesn't have a good function to verify aggregate signatures over different messages,
+        // but it it eventually does expose such a function, we should use that instead of the implementation below.
```
kchalkias, line 665: *if it

kchalkias (approved):
just a small typo. Thanks!!

PR #146, opened 2022-10-21
Move HMAC to SHA3-256

benr-ml:
Changing the hmac & hkdf we will expose in Move to use SHA3-256

### DIFF
--- a/fastcrypto/src/tests/hmac_tests.rs
+++ b/fastcrypto/src/tests/hmac_tests.rs
@@ -3,7 +3,7 @@
 
 use crate::error::FastCryptoError;
 use crate::hash::{Keccak256, ReverseWrapper, Sha3_256};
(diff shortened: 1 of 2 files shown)


fastcrypto/src/tests/hmac_tests.rs
```diff
@@ -81,26 +81,25 @@ fn test_regression_of_salt_padding() {
 
 #[test]
 fn test_regression_of_hmac() {
-    // Using test vectors from https://datatracker.ietf.org/doc/html/rfc4231#page-3
```
joyqvq, line 84: are you still using any test cases from standard? or these are just generated?

benr-ml: no, unfortunately there aren't ones in the rfc

PR #147, opened 2022-10-21
feat: allow converting Arkworks affine BLS12-381 points to / from Zcash binary format

huitseeker:
Arkworks 0.3.0 doesn't follow the RFC format for BLS compressed points, this does.

I would love a review from @joyqvq on this.

### DIFF
--- a/fastcrypto-zkp/src/conversions.rs
+++ b/fastcrypto-zkp/src/conversions.rs
@@ -1,6 +1,6 @@
 // Copyright (c) 2022, Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
-use ark_ff::{Fp384, FromBytes, PrimeField};
+use ark_ff::{BigInteger384, Fp384, FromBytes, PrimeField};
 use ark_serialize::{CanonicalSerialize, CanonicalSerializeWithFlags, EmptyFlags};
 use blst::{blst_fp, blst_fp12, blst_fp6, blst_fp_from_lendian, blst_p1_affine};
 use blst::{blst_fp2, blst_p1_deserialize};
@@ -16,7 +16,9 @@ use blst::{blst_fr, blst_fr_from_uint64, blst_uint64_from_fr};
 
 const SCALAR_SIZE: usize = 32;
 const G1_UNCOMPRESSED_SIZE: usize = 96;
+const G1_COMPRESSED_SIZE: usize = 48;
 const G2_UNCOMPRESSED_SIZE: usize = 192;
+const G2_COMPRESSED_SIZE: usize = 96;
 
 #[inline]
 fn u64s_from_bytes(bytes: &[u8; 32]) -> [u64; 4] {
@@ -135,7 +137,8 @@ pub fn bls_g1_affine_to_blst_g1_affine(pt: &BlsG1Affine) -> blst_p1_affine {
     };
     // See https://github.com/arkworks-rs/curves/issues/14 for why the double serialize
     // we're in fact applying correct masks that arkworks does not use. This may be solved alternatively using
-    // https://github.com/arkworks-rs/algebra/issues/308 in a later release of arkworks
+    // https://github.com/arkworks-rs/algebra/issues/308 in a later release of arkworks. Note:
+    // this is an uncompressed serialization - deserialization.
     let mut tmp2 = [0u8; 96];
     unsafe {
         blst_p1_affine_serialize(tmp2.as_mut_ptr(), &tmp_p1);
@@ -167,7 +170,8 @@ pub fn bls_g2_affine_to_blst_g2_affine(pt: &BlsG2Affine) -> blst_p2_affine {
     };
     // See https://github.com/arkworks-rs/curves/issues/14 for why the double serialize
     // we're in fact applying correct masks that arkworks does not use. This may be solved alternatively using
-    // https://github.com/arkworks-rs/algebra/issues/308 in a later release of arkworks
+    // https://github.com/arkworks-rs/algebra/issues/308 in a later release of arkworks. Note:
+    // this is an uncompressed serialization - deserialization.
     let mut tmp2 = [0u8; G2_UNCOMPRESSED_SIZE];
     unsafe {
         blst_p2_affine_serialize(tmp2.as_mut_ptr(), &tmp_p2);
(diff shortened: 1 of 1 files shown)


fastcrypto-zkp/src/conversions.rs
```diff
@@ -135,7 +137,8 @@ pub fn bls_g1_affine_to_blst_g1_affine(pt: &BlsG1Affine) -> blst_p1_affine {
     };
     // See https://github.com/arkworks-rs/curves/issues/14 for why the double serialize
     // we're in fact applying correct masks that arkworks does not use. This may be solved alternatively using
-    // https://github.com/arkworks-rs/algebra/issues/308 in a later release of arkworks
+    // https://github.com/arkworks-rs/algebra/issues/308 in a later release of arkworks. Note:
```
kchalkias, line 140: +1

fastcrypto-zkp/src/conversions.rs
```diff
@@ -191,6 +195,171 @@ pub fn blst_g2_affine_to_bls_g2_affine(pt: &blst_p2_affine) -> BlsG2Affine {
...
+/////////////////////////////////////////////////////////////
+
+// The standard for serialization of compressed G1, G2 points of BLS12-381
+// is not followed by Arkworks 0.3.0. This is a translation layer to allow
+// us to use (and receive) the standard serialization format.
+// See section 5.4.9.2 of https://zips.z.cash/protocol/protocol.pdf
+// and https://datatracker.ietf.org/doc/draft-irtf-cfrg-bls-signature/ Appendix A
+// for its choice as a standard.
+
+// A note on BLST: blst uses zcash point encoding formats across the board.
```
kchalkias, line 209: Didn't know that, thanks!

fastcrypto-zkp/src/conversions.rs
```diff
@@ -191,6 +195,171 @@ pub fn blst_g2_affine_to_bls_g2_affine(pt: &blst_p2_affine) -> BlsG2Affine {
...
+    result[16..24].copy_from_slice(&rep.0[3].to_be_bytes());
+    result[24..32].copy_from_slice(&rep.0[2].to_be_bytes());
+    result[32..40].copy_from_slice(&rep.0[1].to_be_bytes());
+    result[40..48].copy_from_slice(&rep.0[0].to_be_bytes());
+
+    result
+}
+
+fn bls_fq_from_zcash_bytes(bytes: &[u8; G1_COMPRESSED_SIZE]) -> Option<Fq> {
+    let mut tmp = BigInteger384([0, 0, 0, 0, 0, 0]);
```
kchalkias, line 227: that value could be a const and reusable

kchalkias (commented):
I guess this doesn't significantly affect our benchmarks, right?

fastcrypto-zkp/src/conversions.rs
```diff
@@ -191,6 +195,171 @@ pub fn blst_g2_affine_to_bls_g2_affine(pt: &blst_p2_affine) -> BlsG2Affine {
...
+    result[8..16].copy_from_slice(&rep.0[4].to_be_bytes());
+    result[16..24].copy_from_slice(&rep.0[3].to_be_bytes());
+    result[24..32].copy_from_slice(&rep.0[2].to_be_bytes());
+    result[32..40].copy_from_slice(&rep.0[1].to_be_bytes());
+    result[40..48].copy_from_slice(&rep.0[0].to_be_bytes());
+
+    result
+}
+
+fn bls_fq_from_zcash_bytes(bytes: &[u8; G1_COMPRESSED_SIZE]) -> Option<Fq> {
```
joyqvq, line 226: in what case this would be None? Do we just want this to be `Fq`?

fastcrypto-zkp/src/conversions.rs
```diff
@@ -191,6 +195,171 @@ pub fn blst_g2_affine_to_bls_g2_affine(pt: &blst_p2_affine) -> BlsG2Affine {
...
+        bls_fq_from_zcash_bytes(&tmp)?
+    };
+    let xc0 = {
+        let mut tmp = [0; G1_COMPRESSED_SIZE];
+        tmp.copy_from_slice(&bytes[48..96]);
+
+        bls_fq_from_zcash_bytes(&tmp)?
+    };
+
+    let x = Fq2::new(xc0, xc1);
```
joyqvq, line 339: what is `Fq2`?

fastcrypto-zkp/src/conversions.rs
huitseeker: The call `Fq::from_repr` at the end of this function returns `Option<Fq>`. The typical case for returning `None` is if the input is provided in unreduced form, i.e. if the input `BigInt` is larger than the modulus of the field.

fastcrypto-zkp/src/conversions.rs
huitseeker: `Fq2` is the quadratic extension field which forms the base field over which G2 is defined.

fastcrypto-zkp/src/conversions.rs
huitseeker: Yes, except this is actually not used as a value at all but as an initializer, which is exhaustively rewritten. 

The dominating cost is the allocation, and setting this as a const would actually not change. Further, the entire type `BigInteger384` is not used anywhere else, as this line shows:
https://github.com/MystenLabs/fastcrypto/pull/147/files#diff-b99b0ce42bc8ac2a0026f647de15eb9f02f0c7bcbf3b7ad6b7ff23d628041177R3

huitseeker:
> I guess this doesn't significantly affect our benchmarks, right?

This is writing or reading three bits in the first byte of the representation, so no.

kchalkias (approved):
Ok, got familiar with the logic, seems legit and straightforward. Thanks

joyqvq (approved):
lgtm!

fastcrypto-zkp/src/conversions.rs
kchalkias: maybe there is .zero()
https://github.com/arkworks-rs/algebra/blob/666e80dcfc0754f89580240c854fed2e86080449/ff/src/biginteger/mod.rs#L128

huitseeker: I'll replace it with `default`, which makes more sense here (we don't care about the value).

PR #149, opened 2022-10-21
[zkp] Test infinity throughout our test suite as well

huitseeker:
As it turns out, Arkworks also has a non-standard representation of the infinity point. This integrates the infinity point in all randomly-generated tests, and corrects the discrepancies.

### DIFF
--- a/fastcrypto-zkp/src/conversions.rs
+++ b/fastcrypto-zkp/src/conversions.rs
@@ -489,11 +489,15 @@ pub(crate) mod tests {
 
     // QFE roundtrips
     fn arb_bls_fq2() -> impl Strategy<Value = Fq2> {
-        (arb_bls_fq(), arb_bls_fq()).prop_map(|(fp1, fp2)| Fq2::new(fp1, fp2))
+        (arb_bls_fq(), arb_bls_fq())
+            .prop_map(|(fp1, fp2)| Fq2::new(fp1, fp2))
+            .no_shrink()
     }
 
     fn arb_blst_fp2() -> impl Strategy<Value = blst_fp2> {
-        (arb_blst_fp(), arb_blst_fp()).prop_map(|(fp1, fp2)| blst_fp2 { fp: [fp1, fp2] })
+        (arb_blst_fp(), arb_blst_fp())
+            .prop_map(|(fp1, fp2)| blst_fp2 { fp: [fp1, fp2] })
+            .no_shrink()
     }
 
     proptest! {
@@ -517,12 +521,15 @@ pub(crate) mod tests {
     fn arb_bls_fq6() -> impl Strategy<Value = Fq6> {
         (arb_bls_fq2(), arb_bls_fq2(), arb_bls_fq2())
             .prop_map(|(f_c0, f_c1, f_c2)| Fq6::new(f_c0, f_c1, f_c2))
+            .no_shrink()
     }
 
     fn arb_blst_fp6() -> impl Strategy<Value = blst_fp6> {
-        (arb_blst_fp2(), arb_blst_fp2(), arb_blst_fp2()).prop_map(|(f_c0, f_c1, f_c2)| blst_fp6 {
-            fp2: [f_c0, f_c1, f_c2],
-        })
+        (arb_blst_fp2(), arb_blst_fp2(), arb_blst_fp2())
+            .prop_map(|(f_c0, f_c1, f_c2)| blst_fp6 {
+                fp2: [f_c0, f_c1, f_c2],
+            })
+            .no_shrink()
     }
 
     proptest! {
@@ -542,11 +549,15 @@ pub(crate) mod tests {
     }
 
     fn arb_bls_fq12() -> impl Strategy<Value = Fq12> {
-        (arb_bls_fq6(), arb_bls_fq6()).prop_map(|(f_c0, f_c1)| Fq12::new(f_c0, f_c1))
+        (arb_bls_fq6(), arb_bls_fq6())
+            .prop_map(|(f_c0, f_c1)| Fq12::new(f_c0, f_c1))
+            .no_shrink()
     }
 
     fn arb_blst_fp12() -> impl Strategy<Value = blst_fp12> {
-        (arb_blst_fp6(), arb_blst_fp6()).prop_map(|(f_c0, f_c1)| blst_fp12 { fp6: [f_c0, f_c1] })
+        (arb_blst_fp6(), arb_blst_fp6())
+            .prop_map(|(f_c0, f_c1)| blst_fp12 { fp6: [f_c0, f_c1] })
+            .no_shrink()
     }
 
     proptest! {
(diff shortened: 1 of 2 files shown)


kchalkias (commented):
Seems good code-wise, but need to study the details of the fixed values before approving: 0x40, mask the flag bits, and 10% maybe infinity?

fastcrypto-zkp/src/conversions.rs
```diff
@@ -143,8 +143,18 @@ pub fn bls_g1_affine_to_blst_g1_affine(pt: &BlsG1Affine) -> blst_p1_affine {
...
-    let mut g1 = blst_p1_affine::default();
+    // BLST disagrees with Arkworks on the uncompressed representation of the infinity point on G1
+    // https://github.com/supranational/blst/blob/6382d67c72119d563975892ed49ba32e92d3d0da/src/e1.c#L153-L162
+    // the infinity bit notwithstanding,
+    // Arkworks: x = Fq::zero(), y = Fq::one()
+    // BLST: x = y = Fq::zero()
+    // BLST follows the standard here (see https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-pairing-friendly-curves-10#appendix-C)
+    if pt.infinity {
+        tmp2[0] = 0x40; // set the infinity bit
+        tmp2[48..].copy_from_slice(&[0u8; 48]); // zeroize the rest of the point
```
kchalkias, line 154: Ok now I get it, but I guess for converting y=Fq::one() ->  Fq::zero() you only needed to set one bit Vs copy the whole 48 bytes slice?

fastcrypto-zkp/src/conversions.rs
```diff
@@ -154,12 +164,25 @@ pub fn blst_g1_affine_to_bls_g1_affine(pt: &blst_p1_affine) -> BlsG1Affine {
...
+    // BLST: x = y = Fq::zero()
+    // BLST follows the standard here (see https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-pairing-friendly-curves-10#appendix-C)
+    let y = if infinity {
+        Fq::one()
+    } else {
+        Fp384::from_be_bytes_mod_order(&out[48..])
+    };
+    let x = {
+        // Mask away the flag bits (which Arkworks doesn't use)
+        out[0] &= 0b0001_1111;
```
kchalkias, line 182: I read the RFC, thanks, it makes sense. Because the curve uses 381 bits, the first 3 bits in a 384bit stream (48bytes) are used as flags, but Arkworks didn't take that into account and instead they used the 0,1 point as infinity :(

fastcrypto-zkp/src/conversions.rs
```diff
@@ -176,6 +199,16 @@ pub fn bls_g2_affine_to_blst_g2_affine(pt: &BlsG2Affine) -> blst_p2_affine {
...
     };
+    // BLST disagrees with Arkworks on the uncompressed representation of the infinity point on G2
+    // https://github.com/supranational/blst/blob/6382d67c72119d563975892ed49ba32e92d3d0da/src/e2.c#L194-L203
+    // the infinity bit notwithstanding,
+    // Arkworks: x = Fq2::zero(), y = Fq2::one()
+    // BLST: x = y = Fq2::zero()
+    // BLST follows the standard here (see https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-pairing-friendly-curves-10#appendix-C)
+    if pt.infinity {
+        tmp2[0] = 0x40; // set the infinity bit
+        tmp2[96..].copy_from_slice(&[0u8; 96]); // zeroize the rest of the point
```
kchalkias, line 210: ditto, I think you don't need to copy the full slice but just set one bit to convert `Fq2::one()` -> `Fq2::zero()` (all zero bytes)

fastcrypto-zkp/src/conversions.rs
```diff
@@ -532,38 +579,65 @@ pub(crate) mod tests {
...
+
     pub(crate) fn arb_bls_g1_affine() -> impl Strategy<Value = BlsG1Affine> {
         // slow, but good enough for tests
-        arb_bls_fr().prop_map(|s| {
-            BlsG1Affine::prime_subgroup_generator()
-                .mul(s.into_repr())
-                .into_affine()
+        (arb_bls_fr(), any::<f32>()).prop_map(|(s, maybe_infinity)| {
+            if maybe_infinity < 0.1 {
+                // 10% chance of being the point at infinity
```
kchalkias, line 592: Can you please explain or provide a reference for this 10% probability?

kchalkias:
> As it turns out, Arkworks also has a non-standard representation of the infinity point. This integrates the infinity point in all randomly-generated tests, and corrects the discrepancies.

I remember mentioning this edge case in a previous PR, thanks!

fastcrypto-zkp/src/conversions.rs
huitseeker: Good catch, this is a clunky and erroneous attempt at making this a minority case. The reason it kinda-sorta worked experimentally is that proptest divides floats in 7 strategies: https://docs.rs/proptest/1.0.0/proptest/num/f32/constant.ANY.html
I've refactored this (and other instances) to actually reflect what I intended to do.

fastcrypto-zkp/src/conversions.rs
huitseeker: That works, if I'm being given a legitimate Arkworks point produced by the Arkworks library. 

Except Arkworks conveys the infinity using a separate flag here, so by using custom code, or simply abusing the somewhat permissive deserialization, the caller could send whatever they want in the second coordinate.

fastcrypto-zkp/src/conversions.rs
huitseeker: Same as [the above](https://github.com/MystenLabs/fastcrypto/pull/149#discussion_r1003412930): I think it's safer to not assume the caller will use Arkworks conventions, and return the correct result over the whole input range.

fastcrypto-zkp/src/conversions.rs
kchalkias: interesting, so it's malleable, pretty bad IMO

kchalkias: I'm happy to err on the safe side, so your approach meets the bar.

fastcrypto-zkp/src/conversions.rs
kchalkias: another option is to set this a const and just use that. IMO that would safer vs having code to reconstruct the infinity point every single time.

huitseeker: Right, so given the `blst` definitions all call into FFI functions (all non-const), I don't have the ability to make a const definition, here. But I can certainly make auxiliary functions for clarity.

fastcrypto-zkp/src/conversions.rs
```diff
@@ -154,16 +176,48 @@ pub fn blst_g1_affine_to_bls_g1_affine(pt: &blst_p1_affine) -> BlsG1Affine {
...
         blst_p1_affine_serialize(out.as_mut_ptr(), pt);
     }
-    let infinity = out[0] & (1 << 6) != 0;
-    BlsG1Affine::new(
-        Fp384::from_be_bytes_mod_order(&out[..48]),
-        Fp384::from_be_bytes_mod_order(&out[48..]),
-        infinity,
-    )
+    let infinity = out[0] == 0x40;
+    println!("out: {:?}, infinity: {:?}", out, infinity);
```
kchalkias, line 180: forgotten `println`

mystenadmin (approved):
LGTM

kchalkias (approved):

PR #156, opened 2022-10-24
Let digest type impl debug

jonas-lj:
Compiling Sui with cargo 1.66.0-nightly gives an error when deriving an implementation for Debug for a struct who uses the `Hash` trait to restrict some generic traits. In order to fix this, we restrict the Debug trait for the `TypedDigest`type.

Note that it works with the stable cargo 1.64.0, so it's possible to build using that.


benr-ml (approved):

kchalkias (approved):
Many thanks for debugging and providing a fix!

PR #160, opened 2022-10-24
crypto: Add Base64 and Hex

joyqvq:


benr-ml (changes requested):
Looks great, perhaps let's add a simple tests file checking that encoding-decoding works for all variants (Base64, hex)?

fastcrypto/src/encoding.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+}
+
+impl Encoding for Hex {
+    fn decode(s: &str) -> Result<Vec<u8>, eyre::Report> {
+        decode_bytes_hex(s)
+    }
+
+    fn encode<T: AsRef<[u8]>>(data: T) -> String {
+        format!("0x{}", encode_bytes_hex(&data))
```
joyqvq, line 78: this is a direct copy from Sui, is this a good behavior or too idiomatic?

benr-ml (approved):

fastcrypto/src/encoding.rs
benr-ml: looks good to me

PR #164, opened 2022-10-25
Short aggregate signatures for the unsecure scheme

jonas-lj:
The aggregate signatures for the unsecure scheme was a vector of signatures, but that made it hard to work with, so this PR changes that such that aggregated signatures are the single signatures xor'ed together.

### DIFF
--- a/fastcrypto/src/unsecure/signature.rs
+++ b/fastcrypto/src/unsecure/signature.rs
@@ -28,15 +28,15 @@ use crate::traits::{
     VerifyingKey,
 };
 
-use super::hash::XXH128Unsecure;
+use super::hash::Fast256HashUnsecure;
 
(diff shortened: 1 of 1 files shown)


benr-ml (approved):

fastcrypto/src/unsecure/signature.rs
```diff
@@ -353,13 +356,12 @@ impl FromStr for UnsecureKeyPair {
 }
 
 ///
-/// Implement AggregateAuthenticator
+/// Implement AggregateAuthenticator. Aggregate signatures are implemented as xor's of the individual signatures.
```
kchalkias, line 359: good idea!

kchalkias (approved):

PR #173, opened 2022-10-26
Enable signatures and keys to have different sizes for unsecure scheme

jonas-lj:


kchalkias (approved):
I agree, but for mock dummy sig benchmarking we don't necessarily need larger sizes, as we should optimize on packet size as well.

PR #175, opened 2022-10-26
Extend the BLS library to support also min_pk using macros

benr-ml:
Extending the BLS code to work with both modes.
All objects exist for both modes, under different internal modules: min_sig and min_pk.


kchalkias (commented):
So just to clarify, how external apps will use one or the other when using fastcrypto as a dep?

kchalkias:
Ok I see, so the fastcrypto dependency will be something like `fastcrypto::bls12381::min_sig::BLS12381KeyPair`, and the end user doesn't know anything about macros etc?

benr-ml:
@kchalkias yes, similar abstraction to what we had so far, just with additional modules to differentiate between the modes

kchalkias (approved):
Fantastic, let's gradually have some documentation in a related readme file (future PR), this one is good to be landed!

PR #183, opened 2022-10-27
feat: make parallelism in BLS12-377 configurable

huitseeker:
Also activates asm in ark-ff.

### DIFF
--- a/fastcrypto/Cargo.toml
+++ b/fastcrypto/Cargo.toml
@@ -35,10 +35,10 @@ generic-array = { version = "0.14.6", features = ["serde"] }
 typenum = "1.15.0"
 # TODO: switch to https://github.com/celo-org/celo-bls-snark-rs
 # when https://github.com/celo-org/celo-bls-snark-rs/issues/228 is solved
-celo-bls = { git = "https://github.com/huitseeker/celo-bls-snark-rs", branch = "updates-2", package = "bls-crypto" }
+celo-bls = { git = "https://github.com/huitseeker/celo-bls-snark-rs", branch = "updates-2-with-parallelism-toggle", package = "bls-crypto", default-features = false }
 
 ark-ec = { version = "0.3.0", features = ["std"] }
-ark-ff = { version = "0.3.0", features = ["std"] }
+ark-ff = { version = "0.3.0", features = ["std", "asm"] }
 ark-serialize = { version = "0.3.0", features = ["std"] }
 ark-std = { version = "0.3.0", features = ["std"] }
 blake2 = "0.10.4"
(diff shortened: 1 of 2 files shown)


jonas-lj (approved):

fastcrypto/Cargo.toml
```diff
@@ -64,11 +64,12 @@ name = "encoding"
 harness = false
 
 [features]
-default = ["secure"]
+default = ["secure", "threads-bls_12_377"]
 secure = []
 copy_key = []
 unsecure_schemes = ["dep:twox-hash"]
 no-threads-blst = ["blst/no-threads"]
+threads-bls_12_377 = ["celo-bls/parallel", "ark-ec/parallel", "ark-ff/parallel", "ark-std/parallel"]
```
kchalkias, line 72: It's funny that the 2 BLS projects take a completely different approach re default mode for parallelism :) :)
- BLST... hey we automatically do, please expose `no-threads-blst` if you don't want it.
- celo-bls...  our default behavior is to be sequential, but you can enable the `parallel` feature (you need to enable a few features for that in our different crates)

fastcrypto/Cargo.toml
```diff
@@ -35,10 +35,10 @@ generic-array = { version = "0.14.6", features = ["serde"] }
 typenum = "1.15.0"
 # TODO: switch to https://github.com/celo-org/celo-bls-snark-rs
 # when https://github.com/celo-org/celo-bls-snark-rs/issues/228 is solved
-celo-bls = { git = "https://github.com/huitseeker/celo-bls-snark-rs", branch = "updates-2", package = "bls-crypto" }
+celo-bls = { git = "https://github.com/huitseeker/celo-bls-snark-rs", branch = "updates-2-with-parallelism-toggle", package = "bls-crypto", default-features = false }
```
kchalkias, line 38: Is there a way to point to a non personal branch?

kchalkias (approved):

fastcrypto/Cargo.toml
huitseeker: No. https://github.com/celo-org/celo-bls-snark-rs/issues/228

fastcrypto/Cargo.toml
huitseeker: > celo-bls... our default behavior is to be sequential, but you can enable the parallel feature (you need to enable a few features for that in our different crates)

That's actually our crate, and the default behavior is parallelism https://github.com/celo-org/celo-bls-snark-rs/commit/78f379fe4b16348999a81f57c231102b0913224e

PR #192, opened 2022-10-31
Require AllowedRng everywhere

benr-ml:


fastcrypto/src/lib.rs
```diff
@@ -98,12 +98,12 @@ pub mod unsecure;
 ////////////////////////////////////////////////////////////////
 
 pub fn generate_production_keypair<K: traits::KeyPair>() -> K {
-    generate_keypair::<K, _>(&mut OsRng)
+    generate_keypair::<K, _>(&mut thread_rng())
```
benr-ml, line 101: Please note that I changed OsRng to thread_rng()

joyqvq (approved):
lgtm. this may involve updating rng usage in sui as well when fastcrypto pointer upgrades.

Issue #122, opened 2022-10-16
Ensure that random scalars are at least 128 bit


jonas-lj:
Unrelated to this PR, but we need a TODO fix here as the rand scalars should be at least 128bits (thus we should set vals[1] as well). Ideally I'd create a private function get_128bit_scalar and invoke that.

__Originally posted by @kchalkias in https://github.com/MystenLabs/fastcrypto/pull/119#discussion_r996384609__

Issue #120, opened 2022-10-16
[BLS] Optimize BLS aggregation on different msgs by omitting the z_0 scalar.


kchalkias:
Similar to https://github.com/MystenLabs/fastcrypto/issues/117, but for BLS.

PR #194, opened 2022-10-31
Add get_128bit_scalar function

jonas-lj:
Closing #122

### DIFF
--- a/fastcrypto/src/bls12381.rs
+++ b/fastcrypto/src/bls12381.rs
@@ -254,8 +254,14 @@ impl VerifyingKey for BLS12381PublicKey {
         }
         let mut rands: Vec<blst_scalar> = Vec::with_capacity(sigs.len());
 
+        let mut one = blst_scalar::default();
+        unsafe {
+            blst_scalar_from_uint64(&mut one, 1 as *const u64);
+        }
+        rands.push(one);
+
         let mut rng = rand::thread_rng();
-        for _ in 0..sigs.len() {
+        for _ in 1..sigs.len() {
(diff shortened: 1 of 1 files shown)


fastcrypto/src/bls12381.rs
```diff
@@ -254,21 +253,16 @@ impl VerifyingKey for BLS12381PublicKey {
...
-                // Reject zero as it is used for multiplication.
-                vals[0] = rng.next_u64();
-            }
-            let mut rand_i = MaybeUninit::<blst_scalar>::uninit();
-            unsafe {
-                blst_scalar_from_uint64(rand_i.as_mut_ptr(), vals.as_ptr());
-                rands.push(rand_i.assume_init());
-            }
+
+        let mut one = blst_scalar::default();
```
kchalkias: Let's add a comment that this is an optimization, in accordance to https://github.com/MystenLabs/fastcrypto/issues/120

kchalkias (approved):
Just a small comment

fastcrypto/src/bls12381.rs
jonas-lj: It was put in this PR by accident, so I've removed it again. Will include it in another PR.

jonas-lj: @benr-ml just asked me why it is necessary with 128 bits here. The numbers are chosen by the verifier, so statistical security should be enough, right?

jonas-lj:
> Closing #122

@benr-ml just asked me why 128 bits is necessary here. The numbers are chosen by the verifier, so statistical security should be enough?

PR #195, opened 2022-10-31
Set first coefficient to one in verify_batch_empty_fail_different_msg

jonas-lj:
Closing #120

### DIFF
--- a/fastcrypto/src/bls12381.rs
+++ b/fastcrypto/src/bls12381.rs
@@ -254,8 +254,11 @@ impl VerifyingKey for BLS12381PublicKey {
         }
         let mut rands: Vec<blst_scalar> = Vec::with_capacity(sigs.len());
 
+        // The first coefficient can safely be set to 1 (see https://github.com/MystenLabs/fastcrypto/issues/120)
+        rands.push(get_one());
+
         let mut rng = rand::thread_rng();
-        for _ in 0..sigs.len() {
+        for _ in 1..sigs.len() {
             rands.push(get_128bit_scalar(&mut rng));
         }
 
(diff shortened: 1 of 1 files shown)


fastcrypto/src/bls12381.rs
```diff
@@ -295,6 +298,16 @@ fn get_128bit_scalar<Rng: AllowedRng>(rng: &mut Rng) -> blst_scalar {
     }
 }
 
+fn get_one() -> blst_scalar {
+    let mut one = blst_scalar::default();
+    let mut vals = [0u64; 4];
```
kchalkias, line 303: As it's not clear if blst uses little-endian, can we add a test that this is really 1?

jonas-lj: The API don't really allow an easy test of that, but I've replaces blst_scalar_from_uint64 with blst_scalar_from_le_bytes to make it more clear, that the encoding is little endian.

fastcrypto/src/bls12381.rs
```diff
@@ -254,8 +254,11 @@ impl VerifyingKey for BLS12381PublicKey {
         }
         let mut rands: Vec<blst_scalar> = Vec::with_capacity(sigs.len());
 
+        // The first coefficient can safely be set to 1 (see https://github.com/MystenLabs/fastcrypto/issues/120)
+        rands.push(get_one());
```
benr-ml, line 258: blst_scalar_from_uint32(1)

benr-ml: oh, nvm, just saw that you are changing it to blst_scalar_from_uint64 which is the same

jonas-lj: That won't work. Like blst_scalar_from_uint64, which was used in [30132c0](https://github.com/MystenLabs/fastcrypto/pull/195/commits/30132c035466f13aebbc0020b56db3348412d9e7), the second argument is expected to be an array, so we still need to assume what the encoding is.

benr-ml (approved):

fastcrypto/src/bls12381.rs
kchalkias: @jonas-lj another way is to multiply a BLS elliptic curve point with that scalar, and if it was `one`, the resulted point would remain the same.

jonas-lj: Yea, I started writing a test that the corresponding public key is the generator until I found this solution.

PR #209, opened 2022-11-02
Group benchmarks

jonas-lj:
Put benchmarks in groups for better overview in reports generated by cargo criterion.

This branch was also used to experiment with a github action for continuous benchmarking, hence all the "Update benchmarking.yml" commits, but they have all been reverted and put in another PR #211 .

Issue #181, opened 2022-10-27
[TB-crypto] Define a crypto-standard for Timestamped Benchmarks


kchalkias:
As our backends evolve, we might realize that down the road the efficiency of some schemes might be updated, perhaps also have additional parallelization improvements etc. Let's maintain a timestamped bench-results page with diagrams to track improvement. **Then protocol designers (and crypto paper authors) can really have a fair (and evolving) primitive comparison matrixes**. 

TL;DR We receive incredibly positive feedback that fastcrypto gradually starts being a de facto reference implementation for future research papers and protocol designers due to providing a well-defined comparison matrix between some of the fastest implementations in the space + for batch/aggregated (in the future w precomputed table) + parallelization-friendly versions etc. Our results already:

- changed the perception on how modern BLS impl. perform compared to the fastest ECDSA and EdDSA crates.
- identified unfair or outdated comparisons reported by some lib authors.
- some implementations perform extra serialization operations and/or others provide different security guarantees. for instance:
    -- BLS fast_aggregation_verify can be applied when rogue key attack protection already took place (ie via PoP sigs)
    -- another case in ed25519 batch-verify where two different libs use as inputs Objects Vs raw bytes (and thus extra serialization steps are required in the 2nd, skewing the comparison results a bit).
    -- another case is base64 malleable vs non-malleable and/or constant vs variable time implementations.
- how can someone combine different backends for different setups and input sizes (ie see the results for base64 encoders where for different input sizes we have another winner impl).
- what are the sweet points when we compare the two fastest Rust implementations (or ports). Example: Many ask about "When aggregated BLS starts winning against batched ed25519?", "what if we used a parallelized version?" etc. 
- we already have evidence that some cryptographic protocols, which were supposed to be hanging behind re performance, are in practice a lot faster than expected and vice versa.
- how parallel and/or vectorized implementations perform against each other (ie AVX2, no-threads BLST etc)?
- easily compare against optimized and experimental versions (i.e., with precomputed tables, removing one half-scalar in random combinations, half-aggregation etc).

jonas-lj:
I suggest we do this by adding continuous benchmarking in the CI. The benchmarks could eg. be done on 'publish'. There are tools to do this and generate diagrams: https://github.com/marketplace/actions/continuous-benchmark. I'll test one out on a fork to see if it works.

jonas-lj:
[...]

PR #211, opened 2022-11-03
Add benchmarking workflow to CI

jonas-lj:
Closing #181. PR #209 should be merged first.

This PR adds benchmarking to the CI.
- Running all benchmarks takes ~1 hour, so we only do it on publish, but it may also be triggered manually.
- We use cargo criterion to run benchmarks and generate reports:
    - In order to keep history of benchmarks, we first checkout the old reports from gh-pages
    - Then we use cargo criterion to run all benchmarks with the most recent version of fastcrypto and generate reports,
    - The new reports (including history) are pushed to gh-pages. See https://mystenlabs.github.io/fastcrypto/benchmarks/criterion/reports/ for an incomplete report (not all benchmarks have been run yet), and https://mystenlabs.github.io/fastcrypto/benchmarks/criterion/reports/Verify/Ed25519/history.html for an example of how the history of a benchmark is shown.
    - The results are stored as a JSON-file and stored under benchmarks/history with the latest commit hash as the filename. This allows analysis of historic data and comparison of versions using various analysis tools.

This restructures the gh-pages such that documentation is put under "docs" and benchmarks under "benchmarks".

### DIFF
--- a/.github/workflows/benchmarking.yml
+++ b/.github/workflows/benchmarking.yml
@@ -19,7 +19,7 @@ name: Benchmarking
        - name: Install criterion
          run: cargo install cargo-criterion
        - name: Run benchmarks
-         run: cargo criterion Verify/ --message-format=json > ${{ github.sha }}.json
+         run: cargo criterion --message-format=json > ${{ github.sha }}.json
        - name: Deploy latest benchmark report
          uses: peaceiris/actions-gh-pages@v3
          with:
--- a/.github/workflows/benchmarking.yml
+++ b/.github/workflows/benchmarking.yml
@@ -4,7 +4,7 @@ on:
   release:
     types: [published]
 
- jobs:
+jobs:
    benchmark:
      name: Continuous benchmarking
      runs-on: ubuntu-ghcloud
--- a/.github/workflows/benchmarking.yml
+++ b/.github/workflows/benchmarking.yml
@@ -1,0 +1,36 @@
+name: Benchmarking
+  on:
+    workflow_dispatch:
+    release:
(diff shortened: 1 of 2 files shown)


joyqvq (approved):

kchalkias (commented):
I'd prefer that for CI requirements and to save some CI cycles (and eventually cost) to 
- reduce sampling size to a minimum accepted (even 10-20 might suffice Vs default)
- set some frequency or enabled by triggering it only by a PR author.
 Until we have a better budget and latency estimation.

jonas-lj:
> I'd prefer that for CI requirements and to save some CI cycles (and eventually cost) to
> 
> * reduce sampling size to a minimum accepted (even 10-20 might suffice Vs default)
> * set some frequency or enabled by triggering it only by a PR author.
> 
> ... until we have a better budget and latency estimation plan.

I'll fix the sample size and adjust the significance level also.

Regarding the frequency I think running it only on publish makes sense, both because it doesn't happen that often and because it only makes sense to re-run the benchmarks when we have made significant changes. What do you think about that, @kchalkias?

jonas-lj:
> will there be any alert or warning if performance degraded significantly?
> 
> overall looks great! only compute benchmark on publishing makes sense.

Not as it is now. But there's a Github action (https://github.com/marketplace/actions/continuous-benchmark) which allows this. I think it would give some false positives since the benchmarks fluctuate quite a lot, even when they are running locally. But we could definitely look into it, if you think it would make sens.

kchalkias (approved):
Ok, agree with the plan

PR #216, opened 2022-11-04
chore: expose PrimeField trait

joyqvq:
I missed a trait pub::use for sui to use. this should be the last one 🤞


kchalkias (approved):
approving, but do we want to expose fields from other libs directly or wrap them then expose?

joyqvq:
> approving, but do we want to expose fields from other libs directly or wrap them then expose?

yeah thats the other route - decide to expose them for now faster development. we can wrap them in the future

PR #223, opened 2022-11-08
crypto: Add API for verify groth16 proof in bytes representation

joyqvq:
this reverts the `pub use` added in some previous PRs, and direct implements a method that takes in the serialized form (instead of doing this in Sui). 

the move native function can now ignores all the arkworks specific structs and just invoke this method.

discussed with francois offline that we will just assume the arkworks serialization of all the inputs, instead of using the blst (which uses the zcash standard). if there is a need in the future we can add ser/de conversion support in the api


fastcrypto-zkp/src/api.rs
```diff
@@ -0,0 +1,30 @@
...
+use crate::conversions::BlsFr;
+use crate::verifier::{process_vk_special, verify_with_processed_vk};
+
+/// Verify Groth16 proof using the serialized representation of verifying key, input and proof.
+/// This assumes canoncical ser/de defined in Arkworks, instead of the Zcash bytes representation.
+pub fn verify_groth16_in_bytes(
+    vk_bytes: Vec<u8>,
+    x_bytes: Vec<u8>,
+    proof_bytes: Vec<u8>,
+) -> Result<bool, SynthesisError> {
```
joyqvq, line 17: should we use FastCryptoError here? (currently the enum is not accessible in fastcrypto-zkp)

fastcrypto-zkp/src/unit_tests/verifier_tests.rs
```diff
@@ -212,3 +236,150 @@ fn test_multipairing_with_processed_vk() {
...
+            191, 216, 218, 162, 149, 54, 207, 45, 220, 160, 152, 111, 136, 20, 123, 116, 157, 27,
+            229, 148, 55, 97, 10, 175, 125, 12, 52, 178, 0, 245, 142, 45, 42, 147, 244, 236, 209,
+            66, 8, 163, 20, 88, 56, 4, 221, 42, 59, 194, 131, 236, 0, 222, 1, 236, 247, 137, 56,
+            69, 7
+        ]
+    )
+    .is_err());
+
+    // length of proof is incorrect
+    assert!(verify_groth16_in_bytes(
```
joyqvq, line 293: bytes used in this test is derived from the previous test https://github.com/MystenLabs/fastcrypto/pull/223/files#diff-d842dec425f6894bd38212716a7896634076a8693bd4621554d48476775b9e91L173

fastcrypto-zkp/src/api.rs
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
```
joyqvq, line 1: i created a new mod here to isolate the dependencies. happy to put this inside verifier.rs as well

fastcrypto-zkp/src/api.rs
```diff
@@ -0,0 +1,32 @@
...
+use ark_bls12_381::Bls12_381;
+use ark_groth16::{Proof, VerifyingKey};
+use ark_relations::r1cs::SynthesisError;
+use ark_serialize::CanonicalDeserialize;
+
+use crate::conversions::BlsFr;
+use crate::verifier::{process_vk_special, verify_with_processed_vk};
+
+/// Verify Groth16 proof using the serialized representation of verifying key, input and proof.
+/// This assumes canoncical ser/de defined in Arkworks, instead of the Zcash bytes representation.
```
kchalkias, line 14: @huitseeker which is the official serialization format, based on the current version of RFC?

fastcrypto-zkp/src/api.rs
kchalkias: It should be preferable to have our own controllable fastcrypto error types (this will also make our system backend agnostic, cause now SynthesisError requires an akrworks dependency when outputting errors)

fastcrypto-zkp/src/api.rs
```diff
@@ -0,0 +1,32 @@
...
+pub fn verify_groth16_in_bytes(
```
huitseeker, line 15: I don't quite know how you intend to use this, but just to make clarify: the verification of a Groth16 proof is inherently broken up in two phases:
1. setting up the assets of the verifying key: this involves computation (one pairing e(P, Q)), and "binds" the verification to one particular proof statement (e.g. "I know a sha-256 pre-image that ends in `0xdeadbeef`"),
2. verifying one proof instance: which also involves computation, obviously, e.g. "those points prove I have a valid pre-image, which first 3 bits are public and passed as input".

The expectation is that a smart contract setting up the proof would call 1. once, and then call 2. repeatedly for many proof instances.

Here, the risk with presenting a single API is that the pre-processing of the verifying key contains computation that should be mutualized across the proof instances. If it is not, the price of verification will be one more pairing than necessary per instance.

fastcrypto-zkp/src/unit_tests/verifier_tests.rs
huitseeker: It would be great to be able to take the example in the following test, defined as a constant, clone it, edit the bytes, and then call an assert - it would make your comment clear from the code.

fastcrypto-zkp/src/api.rs
```diff
@@ -0,0 +1,32 @@
...
+    vk_bytes: Vec<u8>,
+    x_bytes: Vec<u8>,
+    proof_bytes: Vec<u8>,
+) -> Result<bool, SynthesisError> {
+    let reader_vk = Cursor::new(vk_bytes);
+    let reader_x = Cursor::new(x_bytes);
+    let reader_proof = Cursor::new(proof_bytes);
```
huitseeker, line 22: It would be really cool if we could use `Untrusted` instead, which should allow panic-free APIs:
https://github.com/briansmith/untrusted
https://briansmith.org/rustdoc/untrusted/index.html

fastcrypto-zkp/src/api.rs
huitseeker: I'm up for exposing our own error type as well, but a minimalistic approach should be fine IMHO (i.e. no need to re-implement all of `SynthesisError`).

fastcrypto-zkp/src/api.rs
```diff
@@ -0,0 +1,32 @@
...
+
+use crate::conversions::BlsFr;
+use crate::verifier::{process_vk_special, verify_with_processed_vk};
+
+/// Verify Groth16 proof using the serialized representation of verifying key, input and proof.
+/// This assumes canoncical ser/de defined in Arkworks, instead of the Zcash bytes representation.
+pub fn verify_groth16_in_bytes(
+    vk_bytes: Vec<u8>,
+    x_bytes: Vec<u8>,
+    proof_bytes: Vec<u8>,
```
huitseeker, line 18: I am not super confident of the separation of the inputs from the three proof points. At least I think they should be renamed as "public inputs" (or similar evocative naming). It also makes sense to me to bundle the three proof points and the scalars together.

huitseeker (commented):
I left a few comments. I think the ones we want to think of first are:
- split API between prepare and verify,
- ergonomics of providing inputs & proof points together / apart
the rest are thing that are important, but relatively quick tweaks (and I'm happy to help).

fastcrypto-zkp/src/api.rs
kchalkias: I support 1 with good documentation

kchalkias (commented):
@joyqvq and @huitseeker lets have a call on Monday for a consensus re api (and multiple options if required). Generally it's not bad to provide two apis if ie one of them uses less inputs and the other more. We do the same for any api that would originally take `Option` in a field. This is because we (almost always) optimize in gas for blockchains.

fastcrypto-zkp/src/api.rs
huitseeker: We can start with 1. and possibly ask for feedback from the community to then support 2, or, push comes to shove, expose the conversion functions themselves.

joyqvq:
addressed most comments, remaining todo is to use Untrusted instead of cursor everywhere.

fastcrypto-zkp/src/api.rs
joyqvq: leaving them separate is much easier for deserialization, what do we gain by bundling them together?

fastcrypto-zkp/src/api.rs
```diff
@@ -0,0 +1,38 @@
...
+
+use ark_bls12_381::Bls12_381;
+use ark_groth16::{Proof, VerifyingKey};
+use ark_serialize::CanonicalDeserialize;
+use fastcrypto::error::FastCryptoError;
+
+use crate::conversions::BlsFr;
+use crate::verifier::{process_vk_special, verify_with_processed_vk, PreparedVerifyingKey};
+
+/// Deserialize bytes as a Arkwork representation of a verifying key, and return a serialized prepared verified key.
```
kchalkias, line 13: a**n** Arkwork

fastcrypto-zkp/src/api.rs
```diff
@@ -0,0 +1,38 @@
...
+    let vk = VerifyingKey::<Bls12_381>::deserialize(reader_vk)
+        .map_err(|_| FastCryptoError::InvalidInput)?;
+
+    process_vk_special(&vk).as_serialized()
+}
+
+/// Verify Groth16 proof using the serialized representation of verifying key, input and proof.
+pub fn verify_groth16_in_bytes(
+    pvk_bytes: Vec<Vec<u8>>,
+    x_bytes: Vec<u8>,
```
kchalkias, line 25: I'm personally ok with the 3 different inputs mainly for readability and deserialization complexity, but let's rename `x` to something meaningful (public_input?).

fastcrypto-zkp/src/api.rs
huitseeker: I think @kchalkias expressed my thinking better here: https://github.com/MystenLabs/fastcrypto/pull/223/files#r1023359491 TL;DR, we should make clear, through renaming variables, that both the proof points and the public inputs are part of a proof instance.
One example : `proof_points_as_bytes`, `proof_public_inputs_as_bytes`

kchalkias (approved):
Fantastic!

PR #226, opened 2022-11-11
Add accumulator multiset hash function

jonas-lj:
Sui implements and uses an accumulator hash function, but it should reside in fastcrypto instead.

### DIFF
--- a/fastcrypto/src/hash.rs
+++ b/fastcrypto/src/hash.rs
@@ -16,6 +16,8 @@
 //! assert_eq!(digest1, digest2);
 //! ```
 
+use core::fmt::Debug;
+use curve25519_dalek_ng::ristretto::RistrettoPoint;
 use digest::OutputSizeUser;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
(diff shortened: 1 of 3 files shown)


fastcrypto/src/hash.rs
```diff
@@ -192,3 +194,76 @@ impl HashFunction<32> for Blake3 {
...
+/// The Accumulator is a homomorphic multiset hash function, which hashes arbitrary multisets
+/// of objects such that the hash of the union of two collections is easy to compute from the
+/// hashes of the two collections.
+///
+/// Concretely, each element is mapped to a point on an elliptic curve on which the DL problem
+/// is hard. The accumulator is the sum of all points.
+///
+/// See for more information about the construction and its security: https://arxiv.org/abs/1601.06502.
+#[derive(Default, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Accumulator {
```
kchalkias, line 207: We should prefer a descriptive name

fastcrypto/src/hash.rs
```diff
@@ -192,3 +194,76 @@ impl HashFunction<32> for Blake3 {
...
+    accumulator: RistrettoPoint,
+}
+
+impl Accumulator {
```
kchalkias, line 211: Wondering if all of these functions are common between accumulators and we should put in the trait then implement them here. That will help adding more accumulator schemes in the future.

jonas-lj: I've now added a trait, _MultisetHash_, and I think that made the code more clear. Let me know if you think it's too abstract for our purpose, because I don't know if we plan to actually have multiple implementations.

fastcrypto/src/hash.rs
jonas-lj: It's not changed to _EllipticCurveMultisetHash_ which is the name used in the paper.

fastcrypto/src/hash.rs
```diff
@@ -192,3 +194,115 @@ impl HashFunction<32> for Blake3 {
...
 }
+
+/// A Multiset Hash is a homomorphic hash function, which hashes arbitrary multisets of objects such
+/// that the hash of the union of two multisets is easy to compute from the hashes of the two multisets.
+///
+/// The hash may be computed incrementally, adding items one at a time, and the order does not affect the
+/// result. The hash of two multisets can be compared by using the Eq trait iml'd for the given hash function,
+/// and the hash function should be collision resistant.
+///
+/// See https://link.springer.com/chapter/10.1007/978-3-540-40061-5_12 for a discussion of this type of hash
```
kchalkias, line 205: Please add the paper's title as well: "Incremental Multiset Hash Functions and Their Application to Memory Integrity Checking" by D. Clarke et al.

kchalkias (approved):

fastcrypto/src/hash.rs
```diff
@@ -192,3 +194,115 @@ impl HashFunction<32> for Blake3 {
         }
     }
 }
+
+/// A Multiset Hash is a homomorphic hash function, which hashes arbitrary multisets of objects such
+/// that the hash of the union of two multisets is easy to compute from the hashes of the two multisets.
+///
+/// The hash may be computed incrementally, adding items one at a time, and the order does not affect the
+/// result. The hash of two multisets can be compared by using the Eq trait iml'd for the given hash function,
```
kchalkias, line 202: impl'd

PR #230, opened 2022-11-14
crypto: Hex::encode do not prefix 0x

joyqvq:
addressing: https://github.com/MystenLabs/sui/pull/5958#discussion_r1020566947


kchalkias (approved):
Thanks @joyqvq, this makes our lib compatible with other hex libs and the 0x formatting is moving to a helper function!

PR #232, opened 2022-11-15
Add ristretto255 module

jonas-lj:


fastcrypto/src/ristretto255.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+/// Represents a point in the Ristretto group for Curve25519.
+#[derive(Debug)]
+pub struct RistrettoPoint {
+    point: curve25519_dalek_ng::ristretto::RistrettoPoint,
+    bytes: OnceCell<[u8; 32]>,
+}
+
+impl RistrettoPoint {
```
kchalkias, line 44: Do we need scalar to point multiplication + is_identity as well for a complete api?

fastcrypto/src/ristretto255.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    /// Create a scalar from the given `u64`.
+    pub fn from(value: u64) -> RistrettoScalar {
+        RistrettoScalar(curve25519_dalek_ng::scalar::Scalar::from(value))
+    }
+
+    /// Create a scalar from the low 255 bits of the given 256-bit integer.
+    pub fn from_bits(value: [u8; 32]) -> RistrettoScalar {
+        RistrettoScalar(curve25519_dalek_ng::scalar::Scalar::from_bits(value))
+    }
```
joyqvq, line 34: would be nice to add some tests for these methods, qq, do we want to return Option instead of Result?

kchalkias: Yes agree with always `returning` a Result for operations that can fail.

fastcrypto/src/ristretto255.rs
jonas-lj: Scalar to point multiplication and map2point is already there as `impl ops::Mul<RistrettoPoint> for RistrettoScalar` and `pub fn from_uniform_bytes(bytes: &[u8; 64]) -> Self` resp. Or is that not what you are referring to? And is_identity should definitely be added.

kchalkias: Indeed re scalar to point.
Map to point is a special function, referred to hash to point sometimes and we usually use the Elligator2 hash function.
Need to check if the Ristretto lib implements Elligator or Elligator2 in https://docs.rs/curve25519-dalek/latest/curve25519_dalek/ristretto/index.html#random-points-and-hashing-to-ristretto

jonas-lj: The `from_uniform_bytes` method in the ristretto lib uses the Elligator map so I think it's already there (see the `hash_from_bytes` function), but maybe it should be renamed. I don't know if it's Elligator or Elligator2 though. It's referred to in the documentation as the "Elligator map" so I'm guessing it's the former, but let me just check.

kchalkias: image

fastcrypto/src/groups/mod.rs
```diff
@@ -0,0 +1,54 @@
...
+
+use serde::de::DeserializeOwned;
+use serde::Serialize;
+
+/// Impl operator overload for the elements of a given group.
+/// * `scalar` -  Type of scalars. Should be equal to group::Scalar.
+/// * `element` - Type of elements. Should be equal to group::Element.
+/// * `group` - Type of group.
+#[macro_export]
+macro_rules! impl_group {
```
jonas-lj, line 12: This macro ensures that operator overloads are implemented for all combinations of borrowed and owned instances.

fastcrypto/src/groups/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+use serde::Serialize;
+
+/// Impl operator overload for the elements of a given group for all combinations of owned and borrowed
+/// values.
+/// * `scalar` -  Type of scalars. Should be equal to group::Scalar.
+/// * `element` - Type of elements. Should be equal to group::Element.
+/// * `group` - Type of group.
+#[macro_export]
+macro_rules! impl_group {
+    ($scalar:ident, $element:ident, $group:ident) => {
```
jonas-lj, line 14: I haven't been able to write a macro that gets the element and scalar types from the group, eg. $group::Element and $group::Scalar, so they have to be given as arguments for now.

huitseeker: We should chat, this is a general place where proc_macros shine.

fastcrypto/src/tests/ristretto255_tests.rs
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
```
huitseeker, line 2: I'd be interested in extending this to the full set of RFC vectors, taking inspiration from https://github.com/pornin/crrl/blob/main/src/ristretto255.rs#L743

huitseeker (commented):
We should chat about macros!
Here's one [old way of doing this](https://github.com/arkworks-rs/snark/pull/62/files#diff-965315447fac5427fb5f047081a2d596ef3416eca2e2dc60698a2235555ab704) (in a different project)

fastcrypto/src/tests/ristretto255_tests.rs
jonas-lj: Done!

fastcrypto/src/groups/ristretto255.rs
```diff
@@ -0,0 +1,178 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Implementations of the [ristretto255 group](https://ristretto.group/) which is a group of
+//! prime order 2^{252} + 27742317777372353535851937790883648493 built over Curve25519.
```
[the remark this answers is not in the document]
jonas-lj: I'm not sure I understand your first point. It this different from the curve25519_dalek_ng imports currently used?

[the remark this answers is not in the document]
jonas-lj: I've added a link to the RFC -- good idea! 

Regarding the imports, it is on purpose, that I've used qualified paths for curve25519_dalek_ng's RistrettoPoint, since I wanted to make it clear when we are using fast cryptos RistrettoPoint and when we are referring to curve25519_dalek_ng's. Wouldn't there be a conflict if we import it and also use the same name for a new struct?

fastcrypto/src/groups/mod.rs
jonas-lj: Sure!

fastcrypto/src/groups/ristretto255.rs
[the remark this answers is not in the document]
huitseeker: Fair concern! I think the best approach here is to use the as keyword to alias the import:
https://doc.rust-lang.org/book/ch07-04-bringing-paths-into-scope-with-the-use-keyword.html#providing-new-names-with-the-as-keyword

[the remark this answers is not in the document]
jonas-lj: That could work. So something like `use curve25519_dalek_ng::ristretto::RistrettoPoint as ExternalRistrettoPoint`?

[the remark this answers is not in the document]
huitseeker: Exactly!

fastcrypto/src/groups/mod.rs
jonas-lj: I've committed a proc_macro to overload the ops now. Let me know what you think!

fastcrypto/src/groups/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+use serde::Serialize;
+
+pub mod ristretto255;
+
+pub trait AdditiveGroupElement: PartialEq + Eq + Serialize + DeserializeOwned {
+    type Group: AdditiveGroup<Element = Self>;
+}
+
+/// Trait impl'd by additive groups.
+pub trait AdditiveGroup {
```
huitseeker, line 14: This may just be a perception, but the usual way in which this is addressed in a whole bunch of rust libraries is to implement `core::ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub, SubAssign};` which `auto_ops` doesn't seem to (or does it?). 

See this example in the original Ristretto impl:
https://github.com/dalek-cryptography/curve25519-dalek/blob/90292dccf84cdc614c1a5640f2d6cb13bfff9761/src/ristretto.rs#L798-L804

We're newtyping this for encapsulation inside the fastcrypto library. I wonder if we could dramatically reduce the amount of new code we have to maintain by using those traits, specifically, the ability of the `derive_more` crate to derive them on a newtype (when the inner type satisfies them):
https://jeltef.github.io/derive_more/derive_more/#operators

The change that we'd see in this trait then is to remove the arithmetic operations, and put the definition of the `core::ops` as required trait bounds. The present trait would then only contain the stuff that isn't already covered, such as the identity.

Does that make sense?

huitseeker (commented):
While I have left an alternative line of though in #247 I think this is OK for now, with the exception of implementing the group trait on a stub stuct. I think it makes more sense to peg this trait on the EC curve point instead - since this is where most operations will originate from. 

In other terms:
```
pub trait AdditiveGroup {
  type Scalar: ...;
  fn add(self, b: &Self) -> Self:
  fn mul(self, b: &Self::Scalar) -> Self;
   ...
}
```
rather than having two associated types (`Element` and `Scalar`).

jonas-lj:
> While I have left an alternative line of though in #247 I think this is OK for now, with the exception of implementing the group trait on a stub stuct. I think it is simpler to peg this trait on the EC curve point instead - since this is where most operations will originate from.
> 
> In other terms:
> 
> ```
> pub trait AdditiveGroup {
>   type Scalar: ...;
>   fn add(self, b: &Self) -> Self:
>   fn mul(self, b: &Self::Scalar) -> Self;
>    ...
> }
> ```
> 
> rather than having two associated types (`Element` and `Scalar`). I'd happily stamp a refactoring that would transform that.

Thanks for taking the time to help out with this!

While it might be simpler to omit the element trait and put it as part of the group, I think it makes more sense to keep them apart. Both because they represent different properties (representation vs algebraic), and because the above interface now suggests that the `add` method adds two groups which is not the case. And also because I could imagine other implementations of groups where the group struct would not be a stub, holding for example cached values or group parameters.

huitseeker:
> While it might be simpler to omit the Element trait and put it as part of the Group trait, I think it makes more sense to keep them apart. Both because they represent different properties (representation vs algebraic), and because the above interface now suggests that the add method adds two groups which is not the case. 

While I think I understand where you're coming from, I am not sure I have seen instances where that distinction is practiced in Rust. e.g.[`std::ops::Add`](https://doc.rust-lang.org/std/ops/trait.Add.html), [`std::cmp::PartialEq`](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html), or the classic [`zkcrypto::group::Group`](https://github.com/zkcrypto/group/blob/3ec0ce7dec5865b2100e266cad82fc7af1b96f71/src/lib.rs#L55-L93) harbor binary `self`/`Self`-based methods. Yet I haven't seen those traits mistaken for (resp) adding additions, comparing equalities, or adding groups.

I'd be fine with a `AdditiveGroupElement` name for the main trait FWIW.

> And also because I could imagine other implementations of groups where the implementation of the Group trait would not be a stub, holding for example cached values or group parameters.

I'm not sure what makes those utilities hard to define without that struct. Further, each usage of the group's operations would then need to allocate a copy of the struct, which depending on its contents, may be costly.

Where such a stub would be really irreplaceable is if we would like to have several, overlapping instances of `Group` instantiated from the same `GroupElement` type. While that's technically possible (e.g. for a quotient group manipulated implicitly), I'm not sure I see where we'd be likely to use that capability.

jonas-lj:
> > While it might be simpler to omit the Element trait and put it as part of the Group trait, I think it makes more sense to keep them apart. Both because they represent different properties (representation vs algebraic), and because the above interface now suggests that the add method adds two groups which is not the case.
> 
> While I think I understand where you're coming from, I am not sure I have seen instances where that distinction is practiced in Rust. e.g.[`std::ops::Add`](https://doc.rust-lang.org/std/ops/trait.Add.html), [`std::cmp::PartialEq`](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html), or the classic [`zkcrypto::group::Group`](https://github.com/zkcrypto/group/blob/3ec0ce7dec5865b2100e266cad82fc7af1b96f71/src/lib.rs#L55-L93) harbor binary `self`/`Self`-based methods. Yet I haven't seen those traits mistaken for (resp) adding additions, comparing equalities, or adding groups.
> 
> I'd be fine with a `AdditiveGroupElement` name for the main trait FWIW.
> 
> > And also because I could imagine other implementations of groups where the implementation of the Group trait would not be a stub, holding for example cached values or group parameters.
> 
> I'm not sure what makes those utilities (values, parameters) hard to define _without_ that struct. Further, each usage of the group's operations would then need to allocate a copy of the struct, which depending on its contents, may be costly.
> 
> Where such a stub would be really irreplaceable is if we would like to have several, overlapping instances of `Group` instantiated from the same `GroupElement` instance. While that's technically possible (e.g. for a quotient group manipulated implicitly), I'm not sure I see where we'd be likely to use that capability.

I'm not sure there would be any confusion either (unless our interface has to handle something like direct sums of groups), but I still think the interface is unclear it ops are put on the group.

The performance argument is a good point. I think using a `AdditiveGroupElement` name for the main trait will work since it's both well-aligned with `std::ops` and makes sense algebraically. I'll get to it right away!

fastcrypto-derive/src/lib.rs
```diff
@@ -38,3 +38,40 @@ pub fn silent_debug(source: TokenStream) -> TokenStream {
     };
     gen.into()
 }
+
+/// Extend implementations of Add, Sub, Mul and Neg into implementations of Add, Sub, Neg, AddAssign,
+/// SubAssign and MulAssign for all combinations of borrowed and owner inputs.
```
jonas-lj, line 43: This will allow the macro to be used to implement combinations of borrowed and owned even when derive_more cannot be used.

huitseeker (approved):
Thanks for the patience & the discussion, @jonas-lj, awesome work!

jonas-lj:
> Thanks for the patience & the discussion, @jonas-lj, awesome work!

You're welcome, and thanks for all the help and feedback!

PR #236, opened 2022-11-17
First draft of mskr

jonas-lj:


fastcrypto/src/bls12377.rs
```diff
@@ -150,8 +150,8 @@ pub struct BLS12377PrivateKey {
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(tag = "type")] // necessary so as not to deser under a != type
 pub struct BLS12377KeyPair {
-    name: BLS12377PublicKey,
-    secret: BLS12377PrivateKey,
+    pub name: BLS12377PublicKey,
+    pub secret: BLS12377PrivateKey,
```
asonnino, line 154: Are we sure about seeing the secret `pub`?

fastcrypto/src/mskr.rs
```diff
@@ -0,0 +1,116 @@
...
+//
+// Implement MSKR for BLS12377
+//
+
+struct BLS12377Hash {}
+
+impl HashToScalar<Fr> for BLS12377Hash {
+    fn hash_to_scalar(bytes: &[u8]) -> Fr {
+        let digest = Sha256::digest(bytes);
+        let b: [u8; 16] = digest.digest[0..16].try_into().unwrap();
```
asonnino, line 37: are 16 enough bytes for collision resistance? What security properties break if there is a collision?

asonnino (approved):

fastcrypto/src/mskr.rs
jonas-lj: I'm not sure. But scalars in  BVL12377 are 253 bits, so I've changed it such that we sample 248 bit now.

fastcrypto/src/bls12377.rs
jonas-lj: I've refactored such that the mksr implementation for bls12377 lives inside the module, so this is not needed anymore.

fastcrypto/src/bls12377.rs
```diff
@@ -760,3 +760,92 @@ impl Drop for BLS12377KeyPair {
...
+    use ark_ec::group::Group;
+    use ark_ff::BigInteger256;
+    use celo_bls::{PrivateKey, PublicKey};
+    use once_cell::sync::OnceCell;
+    use std::ops::Mul;
+
+    pub struct BLS12377Hash {}
+
+    impl HashToScalar<Fr> for BLS12377Hash {
+        fn hash_to_scalar(bytes: &[u8]) -> Fr {
```
kchalkias, line 786: we could use the same method that we generate private keys in BLS where some 32 bytes entropy bytes are an input, in our case the whole digest. However @asonnino it's very common for rogue key protection to use half scalars, because we care about pre-image attacks; you try to cancel-out another pub key, so combinatorics are already very expensive. 

We do use half scalars in many apps (ie in the EdDSA half aggregation) for performance reasons to make that exponentiation half-expensive. In fact though, here we don't care, as this applies once at private key randomization (and with our mode where new-priv_key = old_priv_key \times this_scalar, we don't even have an ECC operation). 

@asonnino: You remember the discussion we had at overleaf, shall we update the private key or for each signature we multiply by this scalar? if we do the former half or full scalar doesn't really matter. In the second you have to pay that cost every time you sign.

at this stage and while experimental, I'll approve the mode as it is. But @jonas-lj let's have a TODO.

kchalkias (approved):

PR #238, opened 2022-11-18
chore: bump version to publish

joyqvq:


fastcrypto-zkp/Cargo.toml
```diff
@@ -22,7 +22,7 @@ ark-serialize = "0.3.0"
 blst = "0.3.10"
 byte-slice-cast = "1.2.2"
 untrusted = "0.9.0"
-fastcrypto = { path = "../fastcrypto" }
+fastcrypto = "0.1.3"
```
joyqvq, line 25: currently this crate just uses the error enum from fastcrypto, ok to lag behind by one version that is definitely published.

kchalkias: but if we publish 0.1.4 why not wait and then do at 0.1.4?

joyqvq:
done via #249

PR #240, opened 2022-11-21
Add Secp256r1

### LINKED ISSUE #217: Support ECDSA secp256r1  (NIST-P1) curve as well

jonas-lj:
Fixes #217 and paves the way for https://github.com/MystenLabs/sui/issues/6235

### DIFF
--- a/fastcrypto/Cargo.toml
+++ b/fastcrypto/Cargo.toml
@@ -37,7 +37,8 @@ typenum = "1.15.0"
 auto_ops = "0.3.0"
 derive_more = "0.99.16"
 p256 = { version = "0.11.1", features = ["ecdsa"] }
-ecdsa = "0.14.8"
+ecdsa = { version = "0.14.8", features = ["rfc6979"] }
+rfc6979 = "0.3.1"
 blake2 = "0.10.5"
 blake3 = "1.3.3"
 blst = "0.3.10"
--- a/fastcrypto/Cargo.toml
+++ b/fastcrypto/Cargo.toml
@@ -37,6 +37,7 @@ typenum = "1.15.0"
 auto_ops = "0.3.0"
 derive_more = "0.99.16"
 p256 = { version = "0.11.1", features = ["ecdsa"] }
+ecdsa = "0.14.8"
 blake2 = "0.10.5"
 blake3 = "1.3.3"
 blst = "0.3.10"
--- a/fastcrypto/Cargo.toml
+++ b/fastcrypto/Cargo.toml
@@ -36,6 +36,7 @@ generic-array = { version = "0.14.6", features = ["serde"] }
 typenum = "1.15.0"
 auto_ops = "0.3.0"
 derive_more = "0.99.16"
+p256 = { version = "0.11.1", features = ["ecdsa"] }
 blake2 = "0.10.5"
 blake3 = "1.3.3"
 blst = "0.3.10"
--- a/fastcrypto/src/secp256r1.rs
+++ b/fastcrypto/src/secp256r1.rs
@@ -32,20 +32,17 @@ use p256::ecdsa::Signature as ExternalSignature;
 use p256::ecdsa::SigningKey as ExternalSecretKey;
 use p256::ecdsa::VerifyingKey as ExternalPublicKey;
 use p256::elliptic_curve::group::GroupEncoding;
-
-use crate::error::FastCryptoError::GeneralError;
 use crate::hash::HashFunction;
 use crate::hash::Sha256;
 use ecdsa::elliptic_curve::bigint::Encoding as OtherEncoding;
 use ecdsa::elliptic_curve::subtle::Choice;
 use ecdsa::elliptic_curve::ScalarCore;
 use ecdsa::RecoveryId;
-use generic_array::GenericArray;
 use p256::elliptic_curve::bigint::ArrayEncoding;
 use p256::elliptic_curve::ops::Reduce;
 use p256::elliptic_curve::sec1::ToEncodedPoint;
 use p256::elliptic_curve::IsHigh;
-use p256::elliptic_curve::{AffineXCoordinate, Curve, DecompactPoint, DecompressPoint, Field};
+use p256::elliptic_curve::{AffineXCoordinate, Curve, DecompressPoint, Field};
 use p256::{AffinePoint, FieldBytes, NistP256, ProjectivePoint, Scalar, U256};
 use serde::{de, Deserialize, Serialize};
 use signature::{Signature, Signer, Verifier};
@@ -122,7 +119,7 @@ impl VerifyingKey for Secp256r1PublicKey {
 
 impl Verifier<Secp256r1Signature> for Secp256r1PublicKey {
     fn verify(&self, msg: &[u8], signature: &Secp256r1Signature) -> Result<(), signature::Error> {
-        // TODO: Until we have a recovery id, we recover both candidates for public keys
+
         let pk = signature
             .recover(msg)
             .map_err(|_| signature::Error::new())?;
@@ -131,9 +128,7 @@ impl Verifier<Secp256r1Signature> for Secp256r1PublicKey {
             return Err(signature::Error::new());
         }
 
-        self.pubkey
-            .verify(msg, &signature.sig)
-            .map_err(|_| signature::Error::new())
+        Ok(())
     }
 }
 
@@ -496,18 +491,18 @@ impl Secp256r1Signature {
     ///
     /// An [FastCryptoError::GeneralError] is returned if no public keys can be recovered.
     pub fn recover(&self, msg: &[u8]) -> Result<Secp256r1PublicKey, FastCryptoError> {
-        // Code copied from recover_verify_key_from_digest_bytes in the k256 crate
+        // Code copied from recover_verify_key_from_digest_bytes in k256@0.11.6
         let (r, s) = self.sig.split_scalars();
 
         let z = Scalar::from_be_bytes_reduced(FieldBytes::from(Sha256::digest(msg).digest));
-        let R = AffinePoint::decompress(&r.to_bytes(), Choice::from(self.recovery_id & 1));
+        let big_r = AffinePoint::decompress(&r.to_bytes(), Choice::from(self.recovery_id & 1));
 
-        if R.is_some().into() {
-            let R = ProjectivePoint::from(R.unwrap());
+        if big_r.is_some().into() {
+            let big_r = ProjectivePoint::from(big_r.unwrap());
             let r_inv = r.invert().unwrap();
             let u1 = -(r_inv * z);
             let u2 = r_inv * *s;
-            let pk = ((ProjectivePoint::GENERATOR * u1) + (R * u2)).to_affine();
+            let pk = ((ProjectivePoint::GENERATOR * u1) + (big_r * u2)).to_affine();
 
             Ok(Secp256r1PublicKey {
                 pubkey: ExternalPublicKey::from_affine(pk)
--- a/fastcrypto/src/secp256r1.rs
+++ b/fastcrypto/src/secp256r1.rs
@@ -413,18 +413,26 @@ impl From<Secp256r1PrivateKey> for Secp256r1KeyPair {
 }
 
 impl Secp256r1Signature {
-    /// Recover public key(s) from signature. Between 0 and 4 potential public keys,
+    /// Recover public key(s) from signature. Either 2 or 4 potential public keys,
     /// which could be the corresponding public key for the signature are returned.
     /// This is based on section 4.1.6 in https://www.secg.org/sec1-v2.pdf.
+    ///
+    /// An [FastCryptoError::GeneralError] is returned if no public keys can be recovered.
     pub fn recover(&self, msg: &[u8]) -> Result<Vec<Secp256r1PublicKey>, FastCryptoError> {
         let (r, s) = self.sig.split_scalars();
         let r_plus_n = U256::from(r.as_ref()).wrapping_add(&NistP256::ORDER);
 
         // Find points with r or r+n as x-coordinate.
-        let pts = vec![
+        let mut pts = vec![
             AffinePoint::decompact(&r.to_bytes()),
             AffinePoint::decompact(&r_plus_n.to_be_byte_array()),
         ];
+
+        // Only keep valid points and return err if there a no such points
+        pts.retain(|pt| pt.is_some().into());
+        if pts.is_empty() {
+            return Err(FastCryptoError::GeneralError);
+        }
 
         // Hash of the message.
         let e = <Scalar as Reduce<U256>>::from_be_bytes_reduced(GenericArray::from(
@@ -436,10 +444,6 @@ impl Secp256r1Signature {
         let g_term = ProjectivePoint::GENERATOR * -r_inv * e;
         let mut candidates: Vec<Secp256r1PublicKey> = Vec::new();
         for affine_pt in pts {
-            if affine_pt.is_none().into() {
-                continue;
-            }
-
             let r_term = ProjectivePoint::from(affine_pt.unwrap()) * r_inv * *s;
             candidates.push(Secp256r1PublicKey {
                 pubkey: ExternalPublicKey::from_affine((g_term + r_term).to_affine()).unwrap(),
--- a/fastcrypto/src/secp256r1.rs
+++ b/fastcrypto/src/secp256r1.rs
@@ -37,6 +37,7 @@ use crate::hash::HashFunction;
 use crate::hash::Sha256;
 use ecdsa::elliptic_curve::bigint::Encoding as OtherEncoding;
 use ecdsa::elliptic_curve::ScalarCore;
+use ecdsa::RecoveryId;
 use generic_array::GenericArray;
 use p256::elliptic_curve::bigint::ArrayEncoding;
 use p256::elliptic_curve::ops::Reduce;
@@ -80,6 +81,7 @@ pub struct Secp256r1PrivateKey {
 pub struct Secp256r1Signature {
     pub sig: ExternalSignature,
     pub bytes: OnceCell<[u8; SIGNATURE_SIZE]>,
+    pub recoveryId: u8,
 }
 
 impl std::hash::Hash for Secp256r1PublicKey {
@@ -281,6 +283,7 @@ impl Signature for Secp256r1Signature {
                     Ok(result) => OnceCell::with_value(result),
                     Err(_) => OnceCell::new(),
                 },
+                recoveryId: 0u8,
             }),
             Err(_) => Err(signature::Error::new()),
         }
@@ -329,6 +332,7 @@ impl Default for Secp256r1Signature {
             sig: ExternalSignature::from_scalars(Scalar::ONE.to_bytes(), Scalar::ONE.to_bytes())
                 .unwrap(),
             bytes: OnceCell::new(),
+            recoveryId: 0u8,
         }
     }
 }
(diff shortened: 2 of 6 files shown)


fastcrypto/src/secp256r1.rs
```diff
@@ -0,0 +1,495 @@
...
+
+impl From<Secp256r1PrivateKey> for Secp256r1KeyPair {
+    fn from(secret: Secp256r1PrivateKey) -> Self {
+        let name = Secp256r1PublicKey::from(&secret);
+        Secp256r1KeyPair { name, secret }
+    }
+}
+
+impl Secp256r1Signature {
+    /// Recover public key from signature
```
kchalkias, line 417: Can you refer to the original impl this was based, just to ensure no malleability and consistency with the secpK1 lib. Ideally we should copy paste that logic (if not).
@joyqvq please have a look as well

fastcrypto/Cargo.toml
```diff
@@ -37,6 +37,7 @@ typenum = "1.15.0"
 # TODO: switch to https://github.com/celo-org/celo-bls-snark-rs
 # when https://github.com/celo-org/celo-bls-snark-rs/issues/228 is solved
 celo-bls = { git = "https://github.com/huitseeker/celo-bls-snark-rs", branch = "updates-2-with-parallelism-toggle", package = "bls-crypto", default-features = false }
+p256 = { version = "0.11.1", features = ["ecdsa"] }
```
kchalkias, line 40: Indeed, this is unfortunately the lib to used, unless @huitseeker has seen any fork of the one we use secpk1.

fastcrypto/src/secp256r1.rs
jonas-lj: I will. It is mostly based on the specs (section 4.1.6 in https://www.secg.org/sec1-v2.pdf) and also on an implementation in the k256 crate. I have two somewhat related questions regarding this @kchalkias :
* I'm assuming we wan't to recover and check public keys every time because of the attack you found recently, but both verifying AND recovering public keys seem to be at least double work. Should we make public key recovery optional?
* Should we add a recovery id byte to signatures as it is done in secp256k1? Since the endgoal is interoperability with hardware wallets, I guess the answer to this depends on how the hardware wallets are handling this, but I haven't been able to find some good documentation.

fastcrypto/Cargo.toml
kchalkias: We will also need the regular verification in both ECDSAs as i/ requires for our fair benchmarks + to switch if in the future we decide that cost is more important than size.

kchalkias (commented):
Clarifications re ecrecover and v flag to be 100% compatible with the other ECDSA.

Note that wallets that sign with libraries that do not support outputs with ecrecover, we're going to provide converters.

We should also normalise signatures as we do for k1 curve, by reducing s value.

fastcrypto/src/secp256r1.rs
```diff
@@ -0,0 +1,492 @@
...
+impl Eq for Secp256r1PublicKey {}
+
+impl VerifyingKey for Secp256r1PublicKey {
+    type PrivKey = Secp256r1PrivateKey;
+    type Sig = Secp256r1Signature;
+    const LENGTH: usize = PUBLIC_KEY_SIZE;
+}
+
+impl Verifier<Secp256r1Signature> for Secp256r1PublicKey {
+    fn verify(&self, msg: &[u8], signature: &Secp256r1Signature) -> Result<(), signature::Error> {
```
joyqvq, line 115: for secp256k1, we have the recovery id appended to the 64 bytes signature. perhaps we can do the same for the r1 signature as well?

fastcrypto/src/tests/secp256r1_tests.rs
```diff
@@ -0,0 +1,446 @@
...
+//         assert!(key_pair_copied_2.public().verify(message, &secp_sig1).is_ok());
+//
+//         // use k256 keypair to verify sig constructed by ffi-implementation
+//         let typed_sig = k256::ecdsa::recoverable::Signature::try_from(signature.as_ref()).unwrap();
+//         assert!(pub_key_1.verify(message, &typed_sig).is_ok());
+//     }
+// }
+
+#[test]
+fn wycheproof_test() {
```
joyqvq, line 401: nice!

fastcrypto/src/secp256r1.rs
```diff
@@ -0,0 +1,492 @@
...
+    /// Recover public key(s) from signature. Either 2 or 4 potential public keys,
+    /// which could be the corresponding public key for the signature are returned.
+    /// This is based on section 4.1.6 in https://www.secg.org/sec1-v2.pdf.
+    ///
+    /// An [FastCryptoError::GeneralError] is returned if no public keys can be recovered.
+    pub fn recover(&self, msg: &[u8]) -> Result<Vec<Secp256r1PublicKey>, FastCryptoError> {
+        // This implementation is based on recover_verifying_key_from_digest_bytes in the p256 crate,
+        // but also handles the case the the x-coordinate is larger than the group order.
+
+        let (r, s) = self.sig.split_scalars();
```
joyqvq, line 425: for secp256k1, we reject signature with s in the lower half of the field. perhaps to maintain the same behavior here.

jonas-lj:
> Clarifications re ecrecover and v flag to be 100% compatible with the other ECDSA.
> 
> Note that wallets that sign with libraries that do not support outputs with ecrecover, we're going to provide converters.
> 
> We should also normalise signatures as we do for k1 curve, by reducing s value.

I'll add v flag and integrate your and Joys comments such that this implementation is 100% compatible with the k256 impl. And then I suggest that we for both ECDSAs implement both verify and ecrecover, perhaps with two different signature types. But I think we should do this in another PR.

kchalkias:
> > Clarifications re ecrecover and v flag to be 100% compatible with the other ECDSA.
> 
> > 
> 
> > Note that wallets that sign with libraries that do not support outputs with ecrecover, we're going to provide converters.
> 
> > 
> 
> > We should also normalise signatures as we do for k1 curve, by reducing s value.
> 
> 
> 
> I'll add v flag and integrate your and Joys comments such that this implementation is 100% compatible with the k256 impl. And then I suggest that we for both ECDSAs implement both verify and ecrecover, perhaps with two different signature types. But I think we should do this in another PR.

Yeah in another PR

fastcrypto/Cargo.toml
huitseeker: You could try the arithmetic generated by fiat-crypto, which is usually pretty darn fast:
https://github.com/mit-plv/fiat-crypto/tree/master/fiat-rust/src
I haven't run benches to compare to k256/p256, though.

fastcrypto/src/secp256r1.rs
```diff
@@ -0,0 +1,492 @@
...
+        }
+    }
+}
+
+impl Default for Secp256r1PublicKey {
+    fn default() -> Self {
+        Secp256r1PublicKey::from_bytes(&[0u8; PUBLIC_KEY_SIZE]).unwrap()
+    }
+}
+
```
huitseeker, line 158: This will 100% panic if called. Please fix, test. I wonder if we've made the same mistake on secp256k1?

fastcrypto/src/secp256r1.rs
```diff
@@ -0,0 +1,492 @@
...
+    type Sig = Secp256r1Signature;
+    const LENGTH: usize = PRIVATE_KEY_SIZE;
+}
+
+impl ToFromBytes for Secp256r1PrivateKey {
+    fn from_bytes(bytes: &[u8]) -> Result<Self, FastCryptoError> {
+        match ExternalSecretKey::try_from(bytes) {
+            Ok(privkey) => Ok(Secp256r1PrivateKey {
+                privkey,
+                bytes: OnceCell::new(),
```
huitseeker, line 206: Well, if the deserialization succeeded, we happen to have a copy of the bytes, maybe we could place them in the `OnceCell`?

fastcrypto/src/tests/secp256r1_tests.rs
```diff
@@ -0,0 +1,446 @@
...
+}
+
+#[test]
+fn test_public_key_recovery() {
+    let kp = keys().pop().unwrap();
+    let message: &[u8] = b"Hello, world!";
+    let signature: Secp256r1Signature = kp.sign(message);
+    let recovered_keys = signature.recover(message).unwrap();
+    assert!(recovered_keys.contains(kp.public()));
+}
```
huitseeker, line 86: Nit: Those roundtrip tests are ideal candidates for a proptest conversion, specially if we implement the [Arbitrary](https://docs.rs/proptest/latest/proptest/arbitrary/trait.Arbitrary.html) trait.

fastcrypto/src/secp256r1.rs
```diff
@@ -0,0 +1,492 @@
...
+            .expect("OnceCell invariant violated")
+    }
+}
+
+impl ToFromBytes for Secp256r1PublicKey {
+    fn from_bytes(bytes: &[u8]) -> Result<Self, FastCryptoError> {
+        match ExternalPublicKey::try_from(bytes) {
+            Ok(pubkey) => Ok(Secp256r1PublicKey {
+                pubkey,
+                bytes: OnceCell::new(),
```
huitseeker, line 146: Same note on populting the `OnceCell` as for the impl on `Secp256r1PrivateKey`.

fastcrypto/src/secp256r1.rs
```diff
@@ -0,0 +1,492 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! This module contains an implementation of the [ECDSA signature scheme](https://en.wikipedia.org/wiki/Elliptic_Curve_Digital_Signature_Algorithm) over the
+//! [secp256r1 NIST-P1 curve](https://www.secg.org/SEC2-Ver-1.0.pdf).
+//!
```
huitseeker, line 6: I think we need a bit more precision here on:
1. whether the reader can find an implementation of RFC 6979 in the below,
2. whether they should expect to find one in the future (forward guidance)

fastcrypto/src/secp256r1.rs
```diff
@@ -0,0 +1,492 @@
...
+
+impl SigningKey for Secp256r1PrivateKey {
+    type PubKey = Secp256r1PublicKey;
+    type Sig = Secp256r1Signature;
+    const LENGTH: usize = PRIVATE_KEY_SIZE;
+}
+
+impl ToFromBytes for Secp256r1PrivateKey {
+    fn from_bytes(bytes: &[u8]) -> Result<Self, FastCryptoError> {
+        match ExternalSecretKey::try_from(bytes) {
```
huitseeker, line 203: I'd love if we could make sure (either through unit testing here, or figuring out a wycheproof test that checks this) that non-canonical scalars are handled properly.

fastcrypto/src/tests/secp256r1_tests.rs
```diff
@@ -0,0 +1,446 @@
...
+//
+// #[test]
+// fn test_public_key_recovery_error() {
```
huitseeker, line 89: It would be great to at least add commentary on what this test should become if we don't have the time to fix it.

fastcrypto/src/tests/secp256r1_tests.rs
```diff
@@ -0,0 +1,446 @@
...
+        &inputs.digests,
+        &inputs.pubkeys,
+        &inputs.signatures,
+    );
+    assert!(res.is_err(), "{:?}", res);
+}
+
+// #[test]
+// fn fail_to_verify_if_upper_s() {
+//     // Test case from https://github.com/fjl/go-ethereum/blob/41c854a60fad2ad9bb732857445624c7214541db/crypto/signature_test.go#L79
```
huitseeker, line 252: Same: please add commentary offering forward guidance on what we want this test to be in the future.

huitseeker (commented):
As much as I love the awesome folks at RustCrypto, I'd urge to read their [scary warning](https://github.com/RustCrypto/elliptic-curves/tree/master/p256) on this crate:
> ## ⚠️ Security Warning
> The elliptic curve arithmetic contained in this crate has never been independently audited! (...)

I think we have several options here, which all improve security:

- we liaise with the RustCrypto folks and get this crate audited,
- we use the [point and field arithmetic from fiat-crypto](https://github.com/mit-plv/fiat-crypto/tree/master/fiat-rust/src)
- we [proptest](https://docs.rs/proptest/latest/proptest/) all the operations we use from p256, comparing the outputs to those of an audited Rust crypto lib such as [ring](https://github.com/briansmith/ring)
- we use symbolic testing with [crux-mir](https://github.com/GaloisInc/crucible/blob/master/crux-mir/README.md)
- more [conventional fuzz-testing](https://github.com/trailofbits/test-fuzz) than proptest
- ...

I don't think we need to use all of those methods in this PR, of course, but I do think we need to think about employing at least a couple of these. Until then, I would make the secp256r1 module `pub(crate)`

jonas-lj:
> Thanks a bunch for the rapid turnaround!
> 
> As much as I love the awesome folks at RustCrypto, I'd urge to read their [scary warning](https://github.com/RustCrypto/elliptic-curves/tree/master/p256) on this crate:
> 
> > ## ⚠️ Security Warning
> > The elliptic curve arithmetic contained in this crate has never been independently audited! (...)
> 
> I think we have several options here, which all improve security:
> 
> * we liaise with the RustCrypto folks and get this crate audited,
> * we use the [point and field arithmetic from fiat-crypto](https://github.com/mit-plv/fiat-crypto/tree/master/fiat-rust/src)
> * we [proptest](https://docs.rs/proptest/latest/proptest/) all the operations we use from p256, comparing the outputs to those of an audited Rust crypto lib such as [ring](https://github.com/briansmith/ring)
> * we use symbolic testing with [crux-mir](https://github.com/GaloisInc/crucible/blob/master/crux-mir/README.md)
> * more [conventional fuzz-testing](https://github.com/trailofbits/test-fuzz) than proptest
> * ...
> 
> I don't think we need to use all of those methods in this PR, of course, but I do think we need to think about employing at least a couple of these. Until then, I would make the secp256r1 module `pub(crate)`

Thanks for the review! 

The issue about lack of independent auditing is not specific for this PR since it's also the case for at least the [k256 crate](https://crates.io/crates/k256) and perhaps others that do not specifically state, that they've been audited. So I suggest we discuss how we should handle this for all crates we depend on and use the ideas you mention as inputs to that discussion. Avoiding re-export of the crate is a fine workaround for now imo.

fastcrypto/src/secp256r1.rs
jonas-lj: We did. I'll create an issue.

fastcrypto/src/secp256r1.rs
jonas-lj: To align the schemes, that certainly makes sense. It's what Ethereum does, right, but I don't think it's part of the standard? The wycheproof tests for p256 fails if signatures with s > n/2 signatures are rejected, so we'll have to get around that if we wan't to enforce this.

fastcrypto/src/secp256r1.rs
jonas-lj: It does use an implementation of RFC6979 (from the p256 crate). I've left a comment on this.

fastcrypto/src/tests/secp256r1_tests.rs
jonas-lj: The test is now fixed.

fastcrypto/src/secp256r1.rs
jonas-lj: There doesn't seem to be a wycheproof test for this. I'll add an issue.

fastcrypto/Cargo.toml
jonas-lj: With the most recent commits, we are reimplementing all of ECDSA anyway, so using a different lib could make sense. We should run some benchmarks first to see if it's worth it though.

jonas-lj:
> Clarifications re ecrecover and v flag to be 100% compatible with the other ECDSA.
> 
> Note that wallets that sign with libraries that do not support outputs with ecrecover, we're going to provide converters.
> 
> We should also normalise signatures as we do for k1 curve, by reducing s value.

This is implemented now. Verification using pub key recover, and code to sign, generate recovery id and pubkey recovery is copied from the k256 crate.

kchalkias (commented):
I think we're in a good state to merge, please resolve conflicts and will approve

jonas-lj:
> I think we're in a good state to merge, please resolve conflicts and will approve

It is done!

fastcrypto/src/secp256r1.rs
```diff
@@ -0,0 +1,566 @@
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let kp = Self::decode_base64(s).map_err(|e| eyre::eyre!("{}", e.to_string()))?;
+        Ok(kp)
+    }
+}
+
+impl Signer<Secp256r1Signature> for Secp256r1KeyPair {
+    fn try_sign(&self, msg: &[u8]) -> Result<Secp256r1Signature, signature::Error> {
+        // Code copied from Sign.rs in k256@0.11.6
```
kchalkias, line 410: Thanks!

fastcrypto/src/secp256r1.rs
```diff
@@ -0,0 +1,566 @@
...
+
+        // Hash message
+        let z = FieldBytes::from(Sha256::digest(msg).digest);
```
kchalkias, line 413: Interesting that internally the message is hashed with sha256.

fastcrypto/src/secp256r1.rs
```diff
@@ -0,0 +1,566 @@
...
+        let s = k_inv * (z + (r * x));
+
+        if s.is_zero().into() {
+            return Err(signature::Error::new());
+        }
+
+        let sig = ExternalSignature::from_scalars(r, s)?;
+
+        // Note: This line is introduced here because big_r.y is a private field.
+        let y: Scalar = Scalar::from_be_bytes_reduced(*big_r.to_encoded_point(false).y().unwrap());
```
kchalkias, line 450: Can this fail with a malformed R? maybe avoid unwrap just in case. ah it's signing, we don't really care, as it will never run on-chain + R is computed here.

fastcrypto/src/secp256r1.rs
```diff
@@ -0,0 +1,566 @@
...
+
+impl Secp256r1Signature {
+    /// Recover the public used to create this signature. This assumes the recovery id byte has been set.
+    ///
+    /// This is copied from `recover_verify_key_from_digest_bytes` in the k256@0.11.6 crate except for a few additions.
+    ///
+    /// An [FastCryptoError::GeneralError] is returned if no public keys can be recovered.
+    pub fn recover(&self, msg: &[u8]) -> Result<Secp256r1PublicKey, FastCryptoError> {
+        let (r, s) = self.sig.split_scalars();
+        let v = RecoveryId::from_byte(self.recovery_id).unwrap();
```
kchalkias, line 496: As this will run on blockchain, let's ensure it returns an error vs panic on fail, unless we know it will never fail, even under malicious signers.

fastcrypto/src/secp256r1.rs
jonas-lj: Yes, I noticed that too. However, now that we have reproduced all the code, we could easily make the hash function generic.

fastcrypto/src/secp256r1.rs
jonas-lj: The `y()` method only returns `None` if the encoded point is compressed or compact, but we specifically ask for an uncompressed encoding, so the unwrap will never fail in this case. I will make the code bit more clear.

fastcrypto/src/secp256r1.rs
jonas-lj: That's true. The deserialization doesn't check the recovery id byte, so a malicious signer could just put anything there. It's fixed now.

kchalkias (approved):
Awesome!

PR #241, opened 2022-11-21
crypto: Add Base58 encoding

joyqvq:
from benchmark result, bs58 is the winning encoding scheme

```
Base58 encoding/bs58_encode_MSG_64B_ZERO
                        time:   [141.77 ns 142.22 ns 142.72 ns]
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
Base58 encoding/base58_encode_MSG_64B_ZERO
                        time:   [157.48 ns 158.25 ns 159.08 ns]
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) high mild
  1 (1.00%) high severe
Base58 encoding/rust_base58_encode_MSG_64B_ZERO
                        time:   [287.63 ns 288.30 ns 289.05 ns]
```


jonas-lj (approved):
Lgtm!

PR #243, opened 2022-11-22
Use FastCryptoError for batched verification

jonas-lj:


jonas-lj:
The changes to BLS12381 seem very large, but it's a result of cargo fmt changing the indention after I removed a long string. It might be related to [this](https://github.com/rust-lang/rustfmt/issues/4800).

kchalkias (commented):
Totally agree w the PR's goal. Some small comments re meaningful error returns.

jonas-lj:
> Totally agree w the PR's goal. Some small comments re meaningful error returns.

Great. I totally agree with your comments regarding the errors. But since it'll require a lot more changes than what I suggest here, I'll postpone it to until after PR #240 is merged where I'll clean up the error handling. I'll add an issue.

PR #247, opened 2022-11-24
[ristretto] Derive ops for easy group definitions (PoC)

huitseeker:
This is a PoC for PR #232 which demonstrates two things:
1. how to derive a variety of [arithmetic operators](https://doc.rust-lang.org/std/ops/index.html) when the operands are NewTypes which inner types support those operations. Included are :
```
core::ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub, SubAssign};
std::iter::Sum;
```
as implemented on both references and owned instances.
2. how to define a trait using those operations. In particular, we show the "convenience trait" pattern, (here `GroupOps, GroupOpsOwned, ScalarMul, ScalarMulOwned`) and their use in the final trait definition.

(see discussion at https://github.com/MystenLabs/fastcrypto/pull/232#discussion_r1030563246)

### DIFF
--- a/fastcrypto-derive/src/lib.rs
+++ b/fastcrypto-derive/src/lib.rs
@@ -9,7 +9,7 @@
 
 use proc_macro::TokenStream;
 use quote::quote;
-use syn::DeriveInput;
+use syn::{parse_macro_input, DeriveInput};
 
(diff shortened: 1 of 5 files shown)


huitseeker:
This has served its purpose. See the results in the discussion & evolution of #232

PR #249, opened 2022-11-25
Remove support for BLS12-377 signing & bump library version

huitseeker:
# Rationale

1. while we do use the BLS12-377 curve for proving (for good reason), we do not use it for signing, and are unlikely to do so in the near to medium term (due to strict performance requirements),
2. @benr-ml is about to introduce a distinct instance of this scheme as a dependency, in the context of threshold signing (see [repo](https://github.com/benr-ml/celo-threshold-bls-rs))
3. The implementations of BLS12-377 are not as mature as other schemes, notably in the area of hashing to the curve. Indeed, BLS12-377 does not yet have a suite in the [hash-to-curve RFC](https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-16.html#suites).

### Celo-BLS-snark-rs (upstream of this implementation)

https://github.com/celo-org/celo-bls-snark-rs
- Unmaintained,
- minSig: [PK in G2](https://github.com/celo-org/celo-bls-snark-rs/blob/master/crates/bls-crypto/src/bls/public.rs#L16)
- uses [try-and-increment as hash to G1](https://github.com/celo-org/celo-bls-snark-rs/blob/master/crates/bls-crypto/src/hash_to_curve/mod.rs)

### W3f/BLS

https://github.com/w3f/bls
- More maintained
- maintenance is [not great](https://github.com/w3f/bls/pull/14)
- optionally, [minSig or minPK](https://github.com/w3f/bls/blob/master/src/engine.rs)
- [customizable](https://github.com/w3f/bls/blob/19099c038c0ee2b5fb323c55d250742a95f28bf6/src/engine.rs#L102-L105) hash-to-curve
- allows an override to Wahby-Boneh in [a PR](https://github.com/w3f/bls/pull/57)
- includes a PoP scheme, which has an [issue](https://github.com/w3f/bls/issues/32)

### Consensys/gnark

https://github.com/ConsenSys/gnark-crypto
- in Go,
- maintained,
- implements an SSWU map (see [RFC](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hash-to-curve/) section 6.6.2)
- the parameters are [undocumented](https://github.com/ConsenSys/gnark-crypto/blob/master/ecc/bls12-377/hash_to_g1.go)
- the parameters may possibly come from [this script](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-06#appendix-E.2)

# Conclusion

It seems best to spend a little bit of time thinking about BLS12-377, a scheme we don't really use, before adopting a version of it. Once we sort out the map_to_curve part, it may be suitable to improve it may be suitable to improve the
implementation we're set to receive from (2.) above (or an upstream impl) rather than continuing with this specific code.


benr-ml:
Correction re #2 - I am not planning to use BLS12-377, it is already removed from the repo I'm working on.

Still, I agree with the suggestion here.

benr-ml (approved):

kchalkias (commented):
I'd suggest we move it to experimental instead of completely deleting it. Mainly for a) tracking: even the comments here in this PR are useful for the community b) comparisons and benches (even FB came back to us for any existing comparison numbers) c) upcoming blogpost with supporting link (we also found other deserved to be mentioned issues, such as parallel implementation is actually slower cc: @jonas-lj

huitseeker:
@kchalkias I want to make super clear that this PR only removes the use of BLS12-377 *as a signing scheme*. To spell out what this means across your topics, *nothing* is lost through this PR concerning the ability to do benchmarking for curve operations or the use of the BLS12-377 curve in proofs, and *nothing* is lost concerning the ability to 1-layer recursion in proofs. Those abilities simply happen to not be supported by the code present in this PR.

kchalkias:
What I mentioned is putting this as a singing scheme in experimental (which lives in the same repo but it's a different crate), does this affect publishing as well?

huitseeker:
An experimental crate that would contain the BLS12-377 signature scheme as currently implemented would only prevent publication of that experimental crate. However, note benchmarks involving that scheme would also have to be run from that crate (since benchmarks require the dependency to be present in `dev-dependencies`).

kchalkias:
@huitseeker off course I get the point that whatever experimental crate we pick, it won't be published to crates.io, but this is fine, it's literally more important for benchmarks (we need to publish the BLS 377 to 381 sig current Rust comparison, as we found some very interesting results and it also justifies the use of 381 for Sui (people from various research groups also asked for these bench, and we need a place to maintain it).

kchalkias (approved):
Happy to accept, let's please document in a release notes the commitID that BLS12377 was removed, so someone can eventually move that code to an experimental bench crate.

PR #250, opened 2022-11-28
feat: write & use a macro for serialization to/from base64

huitseeker:
We repeatedly use the exact same serialization to/from base64, based on the `EncodeDecodeBase64` trait. This simplifies the pattern, and makes it usable from other crates (e.g. Sui).


kchalkias (approved):
Cool!

PR #265, opened 2022-11-30
Add a generic polynomial

benr-ml:


fastcrypto/src/polynomial.rs
```diff
@@ -0,0 +1,185 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// TODO: add license notice
```
kchalkias, line 4: Notice re what?

fastcrypto/src/polynomial.rs
```diff
@@ -0,0 +1,185 @@
...
+        self.0.iter_mut().zip(&other.0).for_each(|(a, b)| *a += *b)
+    }
+
+    // TODO: Some of the functions/steps below may be executed many times in practice thus cache can be
+    // used to improve efficiency (e.g., eval(i) may be called with the same index every time a partial
+    // signature from party i is verified).
+
+    /// Evaluates the polynomial at the specified value.
+    pub fn eval(&self, i: Idx) -> Eval<C> {
+        assert!(i > 0); // Never reveal the secret coefficient directly.
```
kchalkias, line 78: do we want this to panic if maliciously or accidentally invoked at index0?

fastcrypto/src/polynomial.rs
```diff
@@ -0,0 +1,185 @@
...
+            num = num * inv;
+            yi = yi * num;
+            acc += yi;
+        }
+
+        Ok(acc)
+    }
+
+    fn share_map(t: u32, mut shares: Vec<Eval<C>>) -> Result<BTreeMap<Idx, C>, FastCryptoError> {
+        if shares.len() < t.try_into().unwrap() {
```
kchalkias, line 125: indeed `unwrap` is safe here.

fastcrypto/src/polynomial.rs
```diff
@@ -0,0 +1,185 @@
...
+                };
+                // xj - 0
+                num = num * C::ScalarType::from(*j as u64);
+                // 1 / (xj - xi)
+                let tmp = C::ScalarType::from(*j as u64) - C::ScalarType::from(*i as u64);
+                den = den * tmp;
+            }
+            // Next line is safe since i != j.
+            let inv = C::ScalarType::generator() / den;
+            num = num * inv;
```
kchalkias, line 116: you can omit these operations and update `acc += yi * num * inv` directly, no?
Actually you could even avoid setting the temp `yi` and `inv` values and do `acc += *xi * num * C::ScalarType::generator() / den`

fastcrypto/src/polynomial.rs
```diff
@@ -0,0 +1,185 @@
...
+            let mut den = C::ScalarType::generator();
+
+            for j in xs.keys() {
+                if i == j {
+                    continue;
+                };
+                // xj - 0
+                num = num * C::ScalarType::from(*j as u64);
+                // 1 / (xj - xi)
+                let tmp = C::ScalarType::from(*j as u64) - C::ScalarType::from(*i as u64);
```
kchalkias, line 111: ditto, tmp value can be omitted

fastcrypto/src/polynomial.rs
benr-ml: Added a reference to https://github.com/celo-org/celo-threshold-bls-rs (also Apache-2.0).
Let me know if more clarifications are needed.

fastcrypto/src/polynomial.rs
benr-ml: changed i to be NonZeroU32 and removed that assert

fastcrypto/src/groups/ristretto255.rs
```diff
@@ -136,16 +149,29 @@ impl Mul<RistrettoScalar> for RistrettoScalar {
     }
 }
 
+impl Div<RistrettoScalar> for RistrettoScalar {
+    type Output = RistrettoScalar;
+
+    fn div(self, rhs: RistrettoScalar) -> RistrettoScalar {
+        // TODO: Is there a better way?
+        assert_ne!(rhs.0, ExternalRistrettoScalar::zero());
```
kchalkias, line 157: Did you check that? https://github.com/dalek-cryptography/curve25519-dalek/blob/90292dccf84cdc614c1a5640f2d6cb13bfff9761/src/field.rs#L106

fastcrypto/src/polynomial.rs
```diff
@@ -0,0 +1,178 @@
...
+    /// Performs polynomial addition in place.
+    pub fn add(&mut self, other: &Self) {
+        // if we have a smaller degree we should pad with zeros
+        if self.0.len() < other.0.len() {
+            self.0.resize(other.0.len(), C::zero())
+        }
+        self.0.iter_mut().zip(&other.0).for_each(|(a, b)| *a += *b)
+    }
+
+    // TODO: Some of the functions/steps below may be executed many times in practice thus cache can be
```
kchalkias, line 76: Good point, what is a good example in practice for our uses where we'll request at the same index?

fastcrypto/src/polynomial.rs
```diff
@@ -0,0 +1,178 @@
...
+/// element for the coefficients.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Poly<C>(Vec<C>);
+
+pub type PrivatePoly<C> = Poly<<C as GroupElement>::ScalarType>;
+pub type PublicPoly<C> = Poly<C>;
+
+//// Vector related operations.
+
+impl<C> Poly<C> {
```
kchalkias, line 38: Ooc, is the celo paradigm the most popular + fast lib for polynomials?

fastcrypto/src/groups/ristretto255.rs
benr-ml: The todo is actually regarding replacing the assert with something better that doesn't panic.
I don't see how to do that as long as we use the Div trait.

fastcrypto/src/polynomial.rs
benr-ml: e.g., you call it when you want to check a partial signature, so assuming some validators are always answering faster than others, you could cache their public keys, either here or on higher level (e.g., the authentication server).

fastcrypto/src/polynomial.rs
benr-ml: Not sure there is a "paradigm" as we only compute Lagrange coefficients and multiply them with group elements, it's the straightforward method (which many others use as well).
There are faster ways for some use cases, e.g., using FFTs for large values of t, but for our needs it should suffice for now.

fastcrypto/src/groups/ristretto255.rs
kchalkias: Oh I see, then do we need to even check or panic? This might be adversarial input and while we're in a cyclic group, dividing by zero is actually multiplying with the invert and we might decide to actually allow it or we always assume our RistrettoScalar serialization never serializes a zero object, so this will never happen.

benr-ml: Right, we should use it like Optional's unwrap, i.e., only when we know we are safe.
I included the assert_ne call just to make it make explicit and less dependent on the external library.

kchalkias (approved):
Approving to unblock, but will create an issue not to forget this edge case.

PR #266, opened 2022-12-01
[zkp] Add Groth16 hash preimage example

huitseeker:
This proves the knowledge of a Blakek2 hash preimage using Groth16, with a trivial (RNG'd) setup, with the preimage a hidden witness of the proof.

### DIFF
--- a/fastcrypto-zkp/examples/blake2s_demo.rs
+++ b/fastcrypto-zkp/examples/blake2s_demo.rs
@@ -1,0 +1,120 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use ark_bls12_381::{Bls12_381, Fr};
+use ark_ff::ToConstraintField;
+use ark_groth16::{create_random_proof, generate_random_parameters};
+
+use ark_crypto_primitives::{
+    prf::blake2s::constraints::Blake2sGadget, prf::blake2s::Blake2s as B2SPRF, PRFGadget, PRF,
+};
+use ark_relations::r1cs::{
+    ConstraintSynthesizer, ConstraintSystem, ConstraintSystemRef, SynthesisError,
+};
+use blake2::{digest::Digest, Blake2s256};
+
+use ark_r1cs_std::prelude::*;
+use fastcrypto_zkp::verifier::{process_vk_special, verify_with_processed_vk};
+
+#[derive(Clone, Copy, Debug)]
+struct Blake2sCircuit {
+    input: [u8; 32],
+    blake2_seed: [u8; 32],
+    pub expected_output: [u8; 32],
+}
+
+impl Blake2sCircuit {
+    fn new() -> Self {
+        // We're going to prove knowledge of the blake2 hash of this secret
+        // see https://en.wikipedia.org/wiki/Cluedo
+        let statement = b"butler in the yard with a wrench";
+        let bytes = statement.to_vec();
+
+        // as a sanity-check, we'll also compute the blake2 hash of the secret
+        // with the reference implementation
+
+        // Blake2s takes a seed parameter. We use a default seed.
+        let seed: [u8; 32] = [0u8; 32];
+
+        // Compute the hash by traditional means.
+        let mut h = Blake2s256::new_with_prefix(seed);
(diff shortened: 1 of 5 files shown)


fastcrypto-zkp/examples/blake2s_demo.rs
```diff
@@ -0,0 +1,122 @@
...
+use fastcrypto_zkp::verifier::{process_vk_special, verify_with_processed_vk};
+
+#[derive(Clone, Copy, Debug)]
+struct Blake2sCircuit {
+    input: [u8; 32],
+    blake2_seed: [u8; 32],
+    pub expected_output: [u8; 32],
+}
+
+impl Blake2sCircuit {
```
kchalkias, line 27: Awesome, 
Given `some_public_string`, `h = Blake2s(seed, some_public_string)` and a `ZKP(I know secret of h)`, this is literally a signature HMAC scheme proven in ZK, not just a hash, right?

huitseeker: It depends what you mean by "this", in that the exact code you're looking at in this PR sets the variable `some_public_string` as a (hidden) witness rather than a public input. But it would be an easy modification to do what you intend to, yes.

kchalkias: @huitseeker what's the point if both the `seed` and `some_public_string` are hidden? I'd expect two reasonable examples. One for which you prove you know the preimage (no public inputs) (not sure if the seed is required here).

And another that requires the `seed` to be the secret (witness) and `some_public_string` be a public input. 

Both examples have different applications, and hence we probably need them both.

kchalkias: Please do only if it's easy to modify. I used at least both of them in many pitching sessions in the past next to each other to explain ie how easy it is to transform (and why a small transformation is a completely different cool application)

huitseeker: No, your feedback makes total sense, I've now modified the example to present:
> seed, some_public_string (seed is witness, the other is public input)
which is closer to the objective of the example. I was just explaining why one would not want to present *all three examples* jointly in this file.

fastcrypto-zkp/examples/blake2s_demo.rs
```diff
@@ -0,0 +1,121 @@
...
+use ark_relations::r1cs::{
+    ConstraintSynthesizer, ConstraintSystem, ConstraintSystemRef, SynthesisError,
+};
+use blake2::{digest::Digest, Blake2s256};
+
+use ark_r1cs_std::prelude::*;
+use fastcrypto_zkp::verifier::{process_vk_special, verify_with_processed_vk};
+
+#[derive(Clone, Copy, Debug)]
+struct Blake2sCircuit {
```
kchalkias, line 20: I suggest that you add a short public description doc per our discussion. Otherwise the reader has to read through the inline comments to understand the logic. Pretty much what we discussed. 
```
/// A prover proves pre-image knowledge of a hash output, where the hash input is a <public_seed, secret_string>, and they create a circuit where the witness is the secret_string and the public inputs are the public_seed and the hash_output.
```

fastcrypto-zkp/examples/blake2s_demo.rs
```diff
@@ -0,0 +1,121 @@
...
+    fn new() -> Self {
+        // We're going to prove knowledge of the blake2 hash of this secret
+        // see https://en.wikipedia.org/wiki/Cluedo
+        let statement = b"butler in the yard with a wrench";
+        let bytes = statement.to_vec();
+
+        // as a sanity-check, we'll also compute the blake2 hash of the secret
+        // with the reference implementation
+
+        // Blake2s takes a seed parameter. We expect to be committed to an
```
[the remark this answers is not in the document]
huitseeker: > Seed is optional in blake2s, no? 

In Blake2s has keyed and unkeyed variants, and the gadget implements the former for generality.
The "prefix" approach is an unfortunate consequence of the RustCrypto traits, it's not how the RFC approaches it.

I can plug in the default seed.

kchalkias (approved):
Approving and we can improve on docs in upcoming work.

Issue #20, opened 2021-12-25
[crypto] Make sure the crypto API does not expose the hash of a message


huitseeker:
At the moment signature verification passes a hash as an argument:
https://github.com/MystenLabs/narwhal/blob/36968d41de1680a8d79dbc3f7c301f322ba4ad1c/crypto/src/lib.rs#L210-L230

https://github.com/SalusaSecondus/CryptoGotchas/blob/master/README.md#signatures
> An attacker mustn't be allowed to select the actual value being validated in the signature. (The hashing step in all standard signatures defends against that as they can only select the hash pre-image, not the value of the hash.) If they could then they could trivially craft a valid signature for an arbitrary public key by (essentially) generating a random signature and seeing what message would be verified by that signature and then returning that message/signature pair.

benr-ml:
The current code & APIs correctly verify signatures using messages

PR #283, opened 2022-12-06
Add pairing and hash to curve operations

benr-ml:


fastcrypto/src/groups/bls12381.rs
```diff
@@ -135,6 +138,43 @@ impl GroupElement for G1Element {
...
+        unsafe {
+            blst_p1_to_affine(&mut self_affine, &self.0);
+            blst_p2_to_affine(&mut other_affine, &other.0);
+        }
+        // The API requires DST although we don't use hash-to-curve here.
+        let unused_dst = [0u8; 3];
+        let mut pairing_blst = Pairing::new(false, &unused_dst);
+        pairing_blst.raw_aggregate(&other_affine, &self_affine);
+        Self::Output::from(pairing_blst.as_fp12())
+    }
```
huitseeker, line 157: If you are going to compute the pairing with only one set of operands involved, the following is just as efficient and unencumbered by signing APIs:
```
        blst_p1_to_affine(&mut self_a, &self);
        blst_p2_to_affine(&mut other_a, &other);

        blst_miller_loop(&mut gt_point, &other_a, &self_a);
        blst_final_exp(&mut gt_point, &gt_point);
```

benr-ml: nice, thanks for the suggestion

fastcrypto/src/groups/bls12381.rs
```diff
@@ -135,6 +139,42 @@ impl GroupElement for G1Element {
...
+        let mut self_affine = blst_p1_affine::default();
+        let mut other_affine = blst_p2_affine::default();
+        let mut res = blst_fp12::default();
+        unsafe {
+            blst_p1_to_affine(&mut self_affine, &self.0);
+            blst_p2_to_affine(&mut other_affine, &other.0);
+            blst_miller_loop(&mut res, &other_affine, &self_affine);
+            blst_final_exp(&mut res, &res);
+        }
+        Self::Output::from(res)
```
joyqvq, line 156: here it defines (G1, G2) -> GT, are there ways to disallow definitions for "invalid" pairings? i.e. this trait should not implemented for certain type alias

fastcrypto/src/groups/mod.rs
```diff
@@ -44,3 +44,14 @@ pub trait Scalar:
...
+
+pub trait Pair {
+    type Other;
+    type Output;
+
+    fn pair(&self, other: &Self::Other) -> Self::Output;
+}
+
+pub trait FromHashedMessage {
+    fn hash(msg: &[u8]) -> Self;
```
joyqvq, line 56: should this takes in a SHA3_256 type instead of raw bytes? https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-06#section-8.7.1

benr-ml: It does the hashing using the hash function defined by the RFC.
What indeed we may do here is to receive also the domain separator (DST in our code) and pass it to blst. Currently I'm using the standard ones since I don't see why we would want to use, e.g., PoP here. If we need that in the future, we can add the DST string as a parameter.

fastcrypto/src/groups/mod.rs
```diff
@@ -44,3 +44,14 @@ pub trait Scalar:
 {
     fn rand<R: AllowedRng>(rng: &mut R) -> Self;
 }
+
+pub trait Pair {
```
huitseeker, line 48: Nits:
1. does it make sense to put a group bound on `Self` or `Other`? 
2. the trait is defined with a dual `Self` assignment, i.e one could define the $G_2 \times G_1 \mapsto G_T$. operation using this same trait just as easily. 
This may or may not be something we want to resolve.

fastcrypto/src/groups/bls12381.rs
```diff
@@ -218,6 +258,24 @@ impl GroupElement for G2Element {
...
+                msg.len(),
+                DST_G2.as_ptr(),
+                DST_G2.len(),
+                ptr::null(),
+                0,
+            );
+        }
+        Self::from(res)
+    }
+}
```
huitseeker, line 277: OK, so this is the one that is actually confusing, let me explain how I think about inferring what you're trying to do here:
- in idiomatic Rust `FromXYZ` implemented for a type `Foo` indicate there's a mapping from `XYZ` to `Foo`. Here `hash` takes an arbitrary slice of `u8`, so it's unclear what `XYZ` is,
- `FromHashedMessage` would seem to indicate that you're reaching for `XYZ` == a field element, because there is indeed a setup phase in which we hash arbitrary messages to fixed-size field elements before trying to map them to points, [following the RFC-advised API](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve#section-3). The fact that we do not quite have a field trait defined (hi @jonas-lj ) would explain why you'd be taking in bytes instead of a particular type (e.g. `blst_fp2` in the case of G2). OK, so I would have expected to find a call to `map_to_g2` in the implementation (or some variant, see the details of how to make `encode_to_g2` uniform, and here I find `hash_to_g2`, which itself contains a hash to the field. How are the arguments to `hash` "Hashed", exactly?
- Now I'm thinking you may be trying to model the idiosyncratic practice of the sui code base to sign and process the application-level hashes of messages instead of the messages themselves (i.e. we sign hashes of transactions rather than the transaction bytes). This practice is one I have found problematic for a while (#20) but it is pretty solidly entrenched in the Sui (and NW) code base today. The `FromHashedMessage` name would them mean "FromApplicationMessage", which make sense. 

Is this what you were trying to model @benr-ml ? @joyqvq did you understand this as well?

(side note: when it will come to implementing this, I'll advise looking at the RFC for the details of the difference between `blst_hash_to_g2` and `blst_encode_to_g2`, but that's a more involved discussion that's premature at this stage)

fastcrypto/src/groups/mod.rs
benr-ml: 1. added it to self & other
2. let's leave it general

fastcrypto/src/groups/bls12381.rs
```diff
@@ -135,6 +139,42 @@ impl GroupElement for G1Element {
     }
 }
 
+impl Pair for G1Element {
```
kchalkias, line 142: Why didn't we use the word Pairing instead of Pair?

PR #284, opened 2022-12-08
fix : Keypair ser/de

### LINKED ISSUE #18: Keypair Serde is Broken
### Description
Keypair serializes, but does not deserialize when trying to test it out in MystenLabs/narwhal#557. Does not break Narwhal nor Sui so continuing with MystenLabs/narwhal#557 anyways.

To replicate this issue, run the following test:
```
#[test]
fn test_kp_serde() {
    let kp = keys().pop().unwrap();
    let serialized = bincode::serialize(&kp).unwrap();
    let deserialized: Ed25519KeyPair = bincode::deserialize(&serialized).unwrap();
}
```
We get this error:
```
thread 'bls12377_tests::test_kp_serde' panicked at 'called `Result::unwrap()` on an `Err` value: Custom("invalid Base64 encoding")', crypto/src/tests/bls12377_tests.rs:375:75
```

Upon further digging, it looks like removing this line fixes the issue:
```
#[serde(tag = "type")] // REMOVE THIS LINE
pub struct Ed25519KeyPair {
    name: Ed25519PublicKey,
    secret: Ed25519PrivateKey,
}
```

huitseeker:
Fixes #18
I've tested this was compatible with Sui, using the following commit on my fork:
https://github.com/huitseeker/sui/commit/c453e5222c9f523e83e1bf249f5608d33b66168a

### DIFF
--- a/fastcrypto/src/bls12381/mod.rs
+++ b/fastcrypto/src/bls12381/mod.rs
@@ -84,7 +84,6 @@ pub struct BLS12381PrivateKey {
 // There is a strong requirement for this specific impl. in Fab benchmarks.
 /// BLS 12-381 public/private keypair.
 #[derive(Debug, Serialize, Deserialize)]
-#[serde(tag = "type")] // necessary so as not to deserialize under a != type.
 pub struct BLS12381KeyPair {
     name: BLS12381PublicKey,
     secret: BLS12381PrivateKey,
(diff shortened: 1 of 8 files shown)


kchalkias (approved):

jonas-lj (approved):
Well done!

PR #285, opened 2022-12-08
Add serialize/deserialize to BLS groups

jonas-lj:


fastcrypto/src/groups/bls12381.rs
```diff
@@ -135,6 +138,42 @@ impl GroupElement for G1Element {
...
+impl Serialize for G1Element {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let mut bytes = [0u8; 48];
+        unsafe {
+            blst_p1_compress(bytes.as_mut_ptr(), &self.0);
+        }
+        serializer.serialize_bytes(&bytes)
```
huitseeker, line 150: So that's actually just a compressed variant of the point (which will work with Arkworks' `Canonical{De,}Serialize`), but will not match the RFC, which mandates the Zcash representation. If you want to use that later representation, you need to call `blst_p{1,2}_affine_serialize`. 

The reason I'm going to require changes for this PR is not that I'm adamant about one representation or another (arguably, the RFC refers to BLS points used *as keys*, and in this context of a generic trait you may want to make other choices), but that this isn't self-consistent: `blst_p{1,2}_deserialize` uses the Zcash format, while `blst_p{1,2}_compress` does not, so serialize & deserialize won't pass a roundtrip test on all inputs (e.g. infinity).

huitseeker (changes requested):

fastcrypto/src/groups/bls12381.rs
jonas-lj: I chose the compressed format to make the serialization more compact, and the more general deserialize method for deserialisation because it allows both compressed and uncompressed points, assuming the user wants to deserialise points not serialised by fastcrypto.

The roundtrip with infinity actually works as it is now, because the deserialization methods allows both uncompressed and compressed points where it [checks both the compressed and infinity-bits](https://github.com/supranational/blst/blob/f791f7a465cda8ecda74df0a60778331dde40809/src/e1.c#L328-L347). But I don't have any opinion on what format to use, and it might be more clear if we at use one of the serialize/deserialize or compress/decompress combinations.

huitseeker: - the BLS RFC (and by transitivity ZCash) allows for both compressed and uncompressed serialization, and you'll receive no criticism from me for using the compressed one.
- looking at what you're doing again, I was hasty and wrong in my comments, you're correct, there's only one library at play here: blst, which follows the RFC in all points, whether in the compress/decompress or the serialize/deserialize pair. Sorry for the confusion, I'll chalk it up to PTSD from dealing with Arkworks compatibility,
- your thinking in selecting and consistently using one pair of ser/de functions as expressed here aligns with mine:
> I don't have any opinion on what format to use here, but it would be more clear if we use one of the serialize/deserialize or compress/decompress combinations, so I've used the first in rev https://github.com/MystenLabs/fastcrypto/commit/903dee5753d2a56a06016e3ce25a136b5a1177f3.

fastcrypto/src/groups/bls12381.rs
```diff
@@ -277,6 +317,42 @@ impl HashToGroupElement for G2Element {
...
+            let mut affine = blst_p2_affine::default();
+            if blst_p2_deserialize(&mut affine, bytes.as_ptr()) != BLST_ERROR::BLST_SUCCESS {
+                return Err(de::Error::custom("Deserialization failed"));
+            }
+            blst_p2_from_affine(&mut ret, &affine);
+
+            // Verify that the deserialized element is in G2
+            if !blst_p2_in_g2(&ret) {
+                return Err(de::Error::custom("Element is not in subgroup"));
+            }
```
huitseeker, line 350: Yep.

huitseeker (approved):
Tip top, thanks!

PR #286, opened 2022-12-08
tBLS abstractions - sign and verify of full and partial signatures

benr-ml:
Simple high level functionalities for signing, verifying and aggregating partial signatures.
Implementation of this for BLS12-381 min-sig, to be used by the randomness beacon.

### DIFF
--- a/fastcrypto-tbls/Cargo.toml
+++ b/fastcrypto-tbls/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "fastcrypto-tbls"
+version = "0.1.0"
+edition = "2021"
+license = "Apache-2.0"
+authors = ["Mysten Labs <build@mystenlabs.com>"]
+readme = "README.md"
+description = "Threshold BLS and DKG protocols"
+repository = "https://github.com/MystenLabs/fastcrypto"
+
+[dependencies]
+fastcrypto = { path = "../fastcrypto" }
--- a/fastcrypto-tbls/Cargo.toml
+++ b/fastcrypto-tbls/Cargo.toml
@@ -10,3 +10,4 @@ repository = "https://github.com/MystenLabs/fastcrypto"
 
 [dependencies]
 fastcrypto = { path = "../fastcrypto" }
+rand = { version = "0.8.5", features = ["std"] }
--- a/fastcrypto-tbls/src/tbls.rs
+++ b/fastcrypto-tbls/src/tbls.rs
@@ -1,0 +1,81 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Some of the code below is based on code from https://github.com/celo-org/celo-threshold-bls-rs,
+// modified for our needs.
+
+use fastcrypto::error::FastCryptoError;
+use fastcrypto::groups::{bls12381, GroupElement, HashToGroupElement, Pairing, Scalar};
+use fastcrypto::polynomial::{Eval, IndexedValue, Poly};
+
+pub type Share<S> = IndexedValue<S>;
+pub type PartialSignature<S> = IndexedValue<S>;
+
+/// * Trait ThresholdBls provides sign & verify functions for standard and partial BLS signatures.
+pub trait ThresholdBls {
+    type Private: Scalar;
+    /// `Public` represents the group over which the public keys are represented.
+    type Public: GroupElement<ScalarType = Self::Private>;
+    /// `Signature` represents the group over which the signatures are represented.
+    type Signature: GroupElement<ScalarType = Self::Private> + HashToGroupElement;
+
+    fn verify_pairings(p: &Self::Public, sig: &Self::Signature, hm: &Self::Signature) -> bool;
+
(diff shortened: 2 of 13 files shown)


fastcrypto-tbls/src/tbls.rs
```diff
@@ -0,0 +1,77 @@
...
+    type Signature: GroupElement<ScalarType = Self::Private> + HashToGroupElement;
+
+    fn verify_pairings(p: &Self::Public, sig: &Self::Signature, hm: &Self::Signature) -> bool;
+
+    fn sign(private: &Self::Private, msg: &[u8]) -> Self::Signature {
+        let h = Self::Signature::hash_to_group_element(msg);
+        h * private
+    }
+
+    fn verify(public: &Self::Public, msg: &[u8], sig: &Self::Signature) -> bool {
```
kchalkias, line 29: For alignment with the rest of fastcrypto schemes, verification should better return an error vs `bool`

fastcrypto-tbls/src/tbls.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    fn partial_sign(share: &Share<Self::Private>, msg: &[u8]) -> PartialSignature<Self::Signature> {
+        PartialSignature {
+            index: share.index,
+            value: Self::sign(&share.value, msg),
+        }
+    }
+
+    fn partial_verify(
```
kchalkias, line 41: ditto

fastcrypto-tbls/src/tbls.rs
```diff
@@ -0,0 +1,77 @@
...
+/// Implementation of [ThresholdBls] for BLS12-381-min-sig. A variant for BLS12-381-min-pk can be
+/// defined in a similar way if needed in the future.
+pub struct ThresholdBls12381MinSig {}
+
+impl ThresholdBls for ThresholdBls12381MinSig {
+    type Private = bls12381::Scalar;
+    type Public = bls12381::G2Element;
+    type Signature = bls12381::G1Element;
+
+    fn verify_pairings(pk: &Self::Public, sig: &Self::Signature, hm: &Self::Signature) -> bool {
```
kchalkias, line 70: ditto

fastcrypto-tbls/src/tbls.rs
```diff
@@ -0,0 +1,77 @@
...
+pub trait ThresholdBls {
+    type Private: Scalar;
+    /// `Public` represents the group over which the public keys are represented.
+    type Public: GroupElement<ScalarType = Self::Private>;
+    /// `Signature` represents the group over which the signatures are represented.
+    type Signature: GroupElement<ScalarType = Self::Private> + HashToGroupElement;
+
+    fn verify_pairings(p: &Self::Public, sig: &Self::Signature, hm: &Self::Signature) -> bool;
+
+    fn sign(private: &Self::Private, msg: &[u8]) -> Self::Signature {
```
kchalkias, line 24: For signing I guess we're fine and will never fail? iirc we have `try_sign` returning a `Result` in the rest of fastcrypto.

benr-ml: I don't a case we miss here

fastcrypto-tbls/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+version = "0.1.0"
+edition = "2021"
+license = "Apache-2.0"
+authors = ["Mysten Labs <build@mystenlabs.com>"]
+readme = "README.md"
+description = "Threshold BLS and DKG protocols"
+repository = "https://github.com/MystenLabs/fastcrypto"
+
+[dependencies]
+fastcrypto = { path = "../fastcrypto" }
```
joyqvq, line 12: i think this needs to be a a published version of fastcrypto (you can try `cargo publish -p fastcrypto-tbls` --dry-run to confirm)

fastcrypto-tbls/src/tbls.rs
joyqvq: we use both Signature and Result<Signature, Error> in sui/fastcrypto. I think the Signature without error is more appropriate since there is really no valid error cases here?

fastcrypto-tbls/src/tbls.rs
```diff
@@ -0,0 +1,92 @@
...
+
+/// Trait [ThresholdBls] provides sign & verify functions for standard and partial BLS signatures.
+pub trait ThresholdBls {
+    type Private: Scalar;
+    /// `Public` represents the group over which the public keys are represented.
+    type Public: GroupElement<ScalarType = Self::Private>;
+    /// `Signature` represents the group over which the signatures are represented.
+    type Signature: GroupElement<ScalarType = Self::Private> + HashToGroupElement;
+
+    fn verify_pairings(
```
joyqvq, line 22: useful to add /// description of all functions in this trait for rust docs.

fastcrypto-tbls/src/tbls.rs
kchalkias: We're good if signing can never fail.

joyqvq (approved):
lgtm!

PR #287, opened 2022-12-09
Add secp256r1 to the readme

jonas-lj:


huitseeker (approved):
Yay! While I have you here, if you'd be inclined to remove the leftover mentions of BLS12-377, that would be great.

kchalkias (approved):
Let's specify in both secp256 k1 and r1 that it's ECDSA (this is because there are recent schemes (ie in Bitcoin) that use the same curve but for regular Schnorr, not ECDSA

PR #292, opened 2022-12-12
Add signature verification without public key recovery for ecdsa schemes

jonas-lj:


jonas-lj:
This PR is now ready for review. Note that the extension of the implementations for r1 and k1 differ a bit since k1 has a type for non-recoverable signatures, so here the signature field is replaced by an enum which is either a recoverable or non-recoverable signature. For r1, where the underlying crate does not have a recoverable signatures, so the recovery_id field is set as optional instead. This allows match-statements where each case needs to be handled separately.

Wycheproof tests for non-recoverable signatures have been added in both cases, and verification tests are now done on both types of signatures. Benchmarks for verification of a signature have also been added.

fastcrypto/src/secp256k1.rs
```diff
@@ -422,14 +460,38 @@ impl From<Secp256k1PrivateKey> for Secp256k1KeyPair {
 }
 
 impl Secp256k1Signature {
-    /// Recover public key from signature
+    /// Return true if the public key is recoverable from this signature.
+    pub fn is_recoverable(&self) -> bool {
+        match self.sig {
+            RECOVERABLE(_) => true,
+            NONRECOVERABLE(_) => false,
+        }
```
benr-ml, line 468: self.sig == RECOVERABLE(_)

fastcrypto/src/secp256k1.rs
```diff
@@ -422,14 +460,38 @@ impl From<Secp256k1PrivateKey> for Secp256k1KeyPair {
...
+
+    /// Recover public key from signature. If this signature is not recoverable, an
+    /// [FastCryptoError::GeneralError] is returned.
     pub fn recover(&self, hashed_msg: &[u8]) -> Result<Secp256k1PublicKey, FastCryptoError> {
-        match Message::from_slice(hashed_msg) {
-            Ok(message) => match self.sig.recover(&message) {
-                Ok(pubkey) => Secp256k1PublicKey::from_bytes(pubkey.serialize().as_slice()),
-                Err(_) => Err(FastCryptoError::GeneralError),
+        match self.sig {
+            NONRECOVERABLE(_) => Err(FastCryptoError::GeneralError),
```
benr-ml, line 475: InvalidInput?

benr-ml (commented):
I'm not sure I understand why we need `Secp256k1Signature` to include the recoverable option:
- if we have the public key -> let's just call verify, ignoring v
- if we don't -> let's call recover to get the public key and that's it

Should we have a new trait called `RecoverableSiganture` that basically has only `recover()-> Self::VerificationKey` function, and an implementation of `From<RecoverableSignature>` to Signature?

fastcrypto/src/secp256k1.rs
jonas-lj: That's apparently not valid rust code. But the matches macro can do it.

jonas-lj:
> I'm not sure I understand why we need `Secp256k1Signature` to include the recoverable option:
> 
> * if we have the public key -> let's just call verify, ignoring v
> * if we don't -> let's call recover to get the public key and that's it
> 
> Should we have a new trait called `RecoverableSiganture` that basically has only `recover()-> Self::VerificationKey` function, and an implementation of `From<RecoverableSignature>` to Signature?

I've now done as you have suggested. Note that this means that by default, signatures are non-recoverable and recoverable signatures do not have a verify function anymore. Instead one have to either call recover and compare with a public key or convert to a regular signature first.

kchalkias (commented):
The most important part is to see how this is used in Sui. If due to abstraction we expect that every signature verification needs to invoke .verify, then how can one use recoverable sigKeys? @joyqvq to also contribute on this discussion

jonas-lj:
> The most important part is to see how this is used in Sui. If due to abstraction we expect that every signature verification needs to invoke .verify, then how can one use recoverable sigKeys? @joyqvq to also contribute on this discussion

That's a good point. A simple fix could be to let recoverable signatures implement the Signature trait also, allowing a recoverable signature to be used as a default signature in sui also. The verify function could in this case just ignore the recovery id and verify it as Ben suggests. The problem with this is that the recovery id is not verified at all, so we could also, as was done before, recover the public key and compare with the given key.

benr-ml:
> The most important part is to see how this is used in Sui. If due to abstraction we expect that every signature verification needs to invoke .verify, then how can one use recoverable sigKeys? @joyqvq to also contribute on this discussion

joyqvq:
> The most important part is to see how this is used in Sui. If due to abstraction we expect that every signature verification needs to invoke .verify, then how can one use recoverable sigKeys? @joyqvq to also contribute on this discussion

since now we dont care about malleability, we can just ignore the recovery id when verify

benr-ml:
Sorry, clicked the wrong button :(

Today's abstraction is that public keys verify signatures, which is not the right abstraction for recoverable signatures for which we don't have the public key (but only its hash).
Let's understand the use cases and then decide which abstraction needed: public keys that check signatures, or signatures that verify if they are consistent with a function of a given public key. In both cases the change between recoverable <-> recoverable can be done in variants of try_from.

jonas-lj:
> Sorry, clicked the wrong button :(
> 
> Today's abstraction is that public keys verify signatures, which is not the right abstraction for recoverable signatures for which we don't have the public key (but only its hash). Let's understand the use cases and then decide which abstraction needed: public keys that check signatures, or signatures that verify if they are consistent with a function of a given public key. In both cases the change between recoverable <-> recoverable can be done in variants of try_from.

What if we implement a completely different signature scheme for each recoverable signatures where the public key is a sui address (or generally, some predicate on an actual ecdsa public key) instead. This allows us to call verify without having the actual public key, but lets us keep the abstraction of having the verify function on the public key.

kchalkias (commented):
Note that we don't know what the default Sui behavior will be. This will be defined by cluster stress tests. Eventually we'll need to have a default (at least on Sui's side)

jonas-lj:
> Note that we don't know what the default Sui behavior will be. This will be defined by cluster stress tests. Eventually we'll need to have a default (at least on Sui's side)

That make sense. But if we implement two different schemes, recoverable and non-recoverable, for each curve, it will definitely cover this and also handle any compatibility issues. The only issue I see is that it will bloat fastcrypto a bit, but since the main functionality (sign/verify) is different for the two modes, it shouldn't be too bad. I'll make a quick PR to show what I mean.

fastcrypto/src/secp256k1/mod.rs
```diff
@@ -378,19 +354,13 @@ impl FromStr for Secp256k1KeyPair {
 
 impl Signer<Secp256k1Signature> for Secp256k1KeyPair {
     fn try_sign(&self, msg: &[u8]) -> Result<Secp256k1Signature, signature::Error> {
-        let secp = Secp256k1::signing_only();
-        #[cfg(test)]
-        let message =
-            Message::from_slice(<sha3::Keccak256 as digest::Digest>::digest(msg).as_slice())
-                .unwrap();
+        // Sha256 is used by default
```
kchalkias, line 357: I guess this is because we go according to the RFC, right?

fastcrypto/src/secp256r1/mod.rs
```diff
@@ -117,15 +111,14 @@ impl VerifyingKey for Secp256r1PublicKey {
...
 impl Verifier<Secp256r1Signature> for Secp256r1PublicKey {
     fn verify(&self, msg: &[u8], signature: &Secp256r1Signature) -> Result<(), signature::Error> {
-        let pk = signature
-            .recover(msg)
-            .map_err(|_| signature::Error::new())?;
-
-        if pk != *self {
+        // We enforce non malleability, eg. that the s value must be low. This is aligned with
+        // the ECDSA implementation in the secp256k1 crate.
+        if signature.sig.s().is_high().into() {
```
kchalkias, line 116: Thanks, in theory now that we're segwit we don't need it, but it's a good hygiene to have it.

fastcrypto/src/secp256r1/recoverable.rs
```diff
@@ -0,0 +1,581 @@
...
+//
+// Secp256r1RecoverableSignature
+//
+impl<'a, D: PublicKeyDigest<BasePK = Secp256r1PublicKey>> From<&'a Secp256r1RecoverableSignature<D>>
+    for Secp256r1Signature
+{
+    fn from(s: &'a Secp256r1RecoverableSignature<D>) -> Self {
+        Secp256r1Signature {
+            sig: s.clone().sig,
+            bytes: OnceCell::new(), // TODO: May use the first 64 bytes of an existing serialization
```
kchalkias, line 181: Why? should `v` (the 65th byte) be part of serialization too?

fastcrypto/src/secp256r1/mod.rs
```diff
@@ -293,11 +282,10 @@ impl Authenticator for Secp256r1Signature {
 
 impl AsRef<[u8]> for Secp256r1Signature {
     fn as_ref(&self) -> &[u8] {
-        let mut bytes = [0u8; SIGNATURE_SIZE];
-        bytes[..SIGNATURE_SIZE - 1].copy_from_slice(self.sig.as_ref());
-        bytes[SIGNATURE_SIZE - 1] = self.recovery_id;
         self.bytes
-            .get_or_try_init::<_, eyre::Report>(|| Ok(bytes))
+            .get_or_try_init::<_, eyre::Report>(|| {
+                Ok(<[u8; 64]>::try_from(self.sig.as_ref()).unwrap())
```
joyqvq, line 287: replace 64 with named constant SIGNATURE_SIZE

fastcrypto/src/secp256k1/mod.rs
jonas-lj: Which rfc are you referring to? RFC 6979 doesn't specify a specific hash function but uses sha-256 for all examples. We follow "SEC 1: Elliptic Curve Cryptography", which allows one of SHA-224, SHA-256, SHA-384 and SHA-512.

fastcrypto/src/secp256r1/recoverable.rs
jonas-lj: Not for non-recoverable, but we can reuse the first 64 bytes from the serialization of a recoverable signature. It's a very small optimisation and will introduce some complexity so it's probably not worth it though.

PR #297, opened 2022-12-12
Simple mock DKG

benr-ml:
Very simple functions that "emulate" different views of the DKG outputs. To be used as an interim (insecure) solution until we finish the implementation of the DKG protocol.

### DIFF
--- a/fastcrypto-tbls/src/mocked_dkg.rs
+++ b/fastcrypto-tbls/src/mocked_dkg.rs
@@ -56,7 +56,7 @@ pub fn generate_share_and_public_keys(
     epoch: u64,
     id: ShareIndex,
 ) -> (Share, PublicBlsKey, PublicVssKey) {
-    // The private polynomial is c_0=epoch and c_i = 1.
+    // The private polynomial is c_0 = epoch and c_i = 1.
     let mut coefficients: Vec<Scalar> = (0..threshold)
         .into_iter()
         .map(|_| Scalar::generator())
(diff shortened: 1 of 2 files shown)


fastcrypto-tbls/src/mocked_dkg.rs
```diff
@@ -0,0 +1,84 @@
...
+    PrivateBlsKey::from(epoch)
+}
+
+/// Emulate the output of DKG for a given id.
+pub fn generate_share_and_public_keys(
+    threshold: u32,
+    epoch: u64,
+    id: ShareIndex,
+) -> (Share, PublicBlsKey, PublicVssKey) {
+    // The private polynomial is c_0 = epoch and c_i = 1.
```
kchalkias, line 59: yes that works for a mock

fastcrypto-tbls/src/mocked_dkg.rs
```diff
@@ -0,0 +1,84 @@
...
+    let mut coefficients: Vec<Scalar> = (0..threshold)
+        .into_iter()
+        .map(|_| Scalar::generator())
+        .collect();
+    *coefficients.get_mut(0).unwrap() = get_private_key(epoch);
```
kchalkias, line 64: I guess we're ok with unwraps for the mock, but the real implementation will return a result? This is temp api?

benr-ml: yes, it's just a temp solution until we have the dkg, it's not meant to be used for anything but a placeholder

PR #302, opened 2022-12-15
Pin GitHub Actions in workflows to commit SHA

jnaulty:
Used [mheap/pin-github-action](https://github.com/mheap/pin-github-action) to update the workfows in `.github/workflows/*.yml`

The utility updates the actions in the workflow to the specific SHA commit referenced at that particular version. It leaves the specific version pinned as a comment alongside the action.

This change follows best practices documented by:
* [GitHub](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions)
* [OpenSSF](https://github.com/ossf/scorecard/blob/7206a2bdeb7020ab45744a1a6234c9ddf2f3713c/docs/checks.md#pinned-dependencies)

### DIFF
--- a/.github/workflows/benchmarking.yml
+++ b/.github/workflows/benchmarking.yml
@@ -5,33 +5,33 @@ on:
     types: [published]
 
 jobs:
-   benchmark:
-     name: Continuous benchmarking
-     runs-on: ubuntu-ghcloud
-     steps:
-       - uses: actions/checkout@v3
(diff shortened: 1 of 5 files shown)


kchalkias (approved):
Thanks! This is equivalent to what we did for Sui in https://github.com/MystenLabs/sui/pull/6750

PR #311, opened 2022-12-20
crypto: Add Intent and IntentMessage

joyqvq:
moving this from sui to fastcrypto since both sui and narwhal depends on it. see https://github.com/MystenLabs/sui/pull/6927 for usage

this contains some basic tests, the meaty tests lives in sui and narwhal where the struct to bcs serialize is defined. putting behind feature flag since not all users of fastcrypto care about this.


joyqvq:
closing in favor of using a separate crate in sui

PR #312, opened 2022-12-21
Let serialization compress bls group elements

jonas-lj:


benr-ml:
Let's also add to Deserialize checks that the received buffers are of the right sizes?

fastcrypto/src/groups/bls12381.rs
```diff
@@ -11,9 +11,9 @@ use blst::{
     blst_fr, blst_fr_add, blst_fr_cneg, blst_fr_from_scalar, blst_fr_inverse, blst_fr_mul,
     blst_fr_rshift, blst_fr_sub, blst_hash_to_g1, blst_hash_to_g2, blst_lendian_from_scalar,
     blst_miller_loop, blst_p1, blst_p1_add_or_double, blst_p1_affine, blst_p1_cneg,
-    blst_p1_deserialize, blst_p1_from_affine, blst_p1_in_g1, blst_p1_mult, blst_p1_serialize,
+    blst_p1_compress, blst_p1_deserialize, blst_p1_from_affine, blst_p1_in_g1, blst_p1_mult,
```
kchalkias, line 14: Do we need to change the name? It's still serialization, right?

kchalkias (approved):
Approving, but let's ensure naming convention.

fastcrypto/src/groups/bls12381.rs
benr-ml: It is the standard compressed serialization (i.e., 48 bytes for G1 element)

Issue #310, opened 2022-12-20
Keypair encoding / serialization to store the private key only


kchalkias:
This is better hygiene, to also defend against the private - public key mismatching we found on ed25519 libs. The pub key can be derived during deserialization. Creating a similar issue in Sui repo.

joyqvq:
we actually already does this when decoding from base64:

in fastcrypto: 
```
    let secret = <T as KeyPair>::PrivKey::from_bytes(&bytes[..sk_length])?;
    let kp: T = secret.into();
    if kp.public().as_ref() != &bytes[sk_length..] {
        return Err(eyre::eyre!("Invalid keypair"));
    }
```
 in sui: 
```
                    let sk = Secp256k1PrivateKey::from_bytes(
                        bytes
                            .get(1 + Secp256k1PublicKey::LENGTH..)
                            .ok_or_else(|| eyre::eyre!("Invalid length"))?,
                    )?;
                    Ok(SuiKeyPair::Secp256k1(<Secp256k1KeyPair as From<
                        Secp256k1PrivateKey,
                    >>::from(sk)))
```

we can do this for encoding as well
```
            SuiKeyPair::Ed25519(kp) => {
                let kp1 = kp.copy();
                bytes.extend_from_slice(&[self.public().flag()]);
                bytes.extend_from_slice(kp.public().as_ref()); // change this to pubkey::from(kp.private())
                bytes.extend_from_slice(kp1.private().as_ref());
            }
```

PR #313, opened 2022-12-22
crypto: Rederive pubkey instead of using self.name

joyqvq:
https://github.com/MystenLabs/fastcrypto/issues/310


benr-ml (approved):

kchalkias (approved):
Thanks!

PR #315, opened 2022-12-22
Serialize BLS12-381 group elements using fixed size arrays

benr-ml:


fastcrypto/src/groups/bls12381.rs
```diff
@@ -317,6 +327,15 @@ impl HashToGroupElement for G2Element {
     }
 }
 
+// Serde treats arrays larger than 32 as variable length arrays, and then add the length as a prefix.
```
kchalkias, line 330: Good idea.

kchalkias (approved):

PR #317, opened 2022-12-25
crypto: PartialEq and Eq for kp

joyqvq:
necessary for https://github.com/MystenLabs/sui/pull/7011 to put a keypair in `OnceCell`


benr-ml (approved):

kchalkias (approved):
Thanks

PR #327, opened 2022-12-28
{PublicKeyBytes, Base64Representation} -> BytesRepresentation, and introduce serialize_deserialize_with_to_from_byte_array

benr-ml:
- Removed `PublicKeyBytes` and it dependencies (verified they are not used in `sui/`). That struct could have stored invalid "serialized" objects, thus it was not always safe to use it.
- Changed `Base64Representation` to serialize differently depending on `is_human_readable` and renamed it to `BytesRepresentation`. (I couldn't find a nicer way to support both options unfortunately).
We may use `BytesRepresentation` everywhere we intended to use `PublicKeyBytes`.
Also added a macro for deriving such type for a given type `generate_bytes_representation`.
- Added a macro `serialize_deserialize_with_to_from_byte_array` for simplifying adding ser/des for objects without caching.


fastcrypto/src/bytes_representation.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+impl<T, const N: usize> AsRef<[u8]> for BytesRepresentation<T, N> {
+    fn as_ref(&self) -> &[u8] {
+        self.bytes.as_ref()
+    }
+}
+
+// Define our own serialize/deserialize functions instead of using #[serde_as(as = "Base64")]
+// so we could serialize a flat object (i.e., "1234" instead of "{ bytes: 1234 }").
```
joyqvq, line 54: i think you can use flatten here? https://serde.rs/attr-flatten.html

fastcrypto/src/bytes_representation.rs
```diff
@@ -0,0 +1,137 @@
...
+                    Base64::decode(&s).map_err(|_| de::Error::custom("Base64 decoding failed"))?;
+                if decoded.len() != N {
+                    return Err(de::Error::custom(format!(
+                        "Invalid buffer length {}, expecting {}",
+                        decoded.len(),
+                        N
+                    )));
+                }
+                let mut bytes = [0u8; N];
+                bytes.copy_from_slice(&decoded[..N]);
```
joyqvq, line 87:
```suggestion
                decoded.try_into();
```

fastcrypto/src/bytes_representation.rs
benr-ml: it's actually an old irrelevant comment, now the serializer does more things.
removed the comments

joyqvq (approved):
lgtm!

fastcrypto/src/tests/bls12381_tests.rs
```diff
@@ -446,13 +446,13 @@ fn test_hkdf_generate_from_ikm() {
...
 }
 
-#[test]
-fn test_public_key_bytes_conversion() {
-    let kp = keys().pop().unwrap();
-    let pk_bytes: BLS12381PublicKeyBytes = kp.public().into();
-    let rebuilt_pk: BLS12381PublicKey = pk_bytes.try_into().unwrap();
-    assert_eq!(kp.public().as_bytes(), rebuilt_pk.as_bytes());
-}
+// #[test]
```
kchalkias, line 449: Do we still need these tests as comments?

fastcrypto/src/serde_helpers.rs
```diff
@@ -118,3 +121,241 @@ impl<'de> DeserializeAs<'de, ed25519_consensus::Signature> for Ed25519Signature
...
+///
+/// Every type's serialization should work as following:
+/// - Return base64 encoded bytes for is_human_readable() serializers.
+/// - Else, return the bytes as a fixed size array. This should be the canonical representation of
+///   the object (e.g., as defined in RFCs).
+///
+/// Types that do store a *cached* version of the serialized object should implement [ToFromBytes]
+/// and call [serialize_deserialize_with_to_from_bytes].
+///
+/// Types that do *not* store a cached version of the serialized object should implement
```
kchalkias, line 136: Is there any way to merge these two options? Why should a developer know if cache exists?

benr-ml: Cache as in having `pub bytes: OnceCell<Vec<u8>>` inside the type. The developer should know if its type is using that or not, and depending on that use `ToFromBytes` vs `ToFromByteArray`.

kchalkias: @benr-ml interesting that caching is public api, and because of this indeed even external devs who just use the api can (in theory) know if it uses cache or not. Not a strong fan of this knowledge requirement, but 🤷

kchalkias (approved):

PR #331, opened 2022-12-30
Update the serializer of aggregated BLS

benr-ml:
- Move the serializer of aggregated BLS to the same file with the rest of the code
- Use SerializationHelper to serialize a fixed length array (as it previously used more bytes since it added the length as a prefix)

### DIFF
--- a/fastcrypto/src/serde_helpers.rs
+++ b/fastcrypto/src/serde_helpers.rs
@@ -18,7 +18,7 @@ use crate::{
     traits::{KeyPair, SigningKey, ToFromBytes, VerifyingKey},
 };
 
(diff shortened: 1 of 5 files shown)


fastcrypto/src/serde_helpers.rs
```diff
@@ -18,61 +18,14 @@ use crate::{
     traits::{KeyPair, SigningKey, ToFromBytes, VerifyingKey},
 };
 
-fn to_custom_error<'de, D, E>(e: E) -> D::Error
+pub fn to_custom_error<'de, D, E>(e: E) -> D::Error
```
kchalkias, line 21: `pub` or `pub(crate)`?

kchalkias (approved):
Thanks

PR #332, opened 2023-01-01
Move Poly<> and add ECIES from Ristretto

benr-ml:


benr-ml:
@kchalkias please take a look. 
While there are still several TODOs I'm planning to resolve, I think all the interesting code is already in and I'd appreciate an early feedback.

fastcrypto-tbls/src/ecies.rs
```diff
@@ -0,0 +1,196 @@
...
+use fastcrypto::error::FastCryptoError;
+use fastcrypto::groups::bls12381::G1Element;
+use fastcrypto::groups::{GroupElement, HashToGroupElement, Scalar};
+use fastcrypto::hmac::{hkdf_sha3_256, HkdfIkm};
+use fastcrypto::traits::{AllowedRng, Cipher, ToFromBytes};
+use serde::{Deserialize, Serialize};
+use typenum::consts::U16;
+
+///
+/// Simple ECIES encryption using a generic group and Aes-128-counter.
```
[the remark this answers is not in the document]
benr-ml: changed to 256

fastcrypto/src/groups/ristretto255.rs
```diff
@@ -173,6 +174,16 @@ impl GroupElement for RistrettoScalar {
 
 impl Scalar for RistrettoScalar {
     fn rand<R: AllowedRng>(rng: &mut R) -> Self {
-        RistrettoScalar::from(ExternalRistrettoScalar::random(rng))
+        Self(ExternalRistrettoScalar::random(rng))
+    }
+}
+
+impl HashToGroupElement for RistrettoScalar {
```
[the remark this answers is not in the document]
benr-ml: note that the output is a scalar. I don't think there is a standard for this, you only need a hash function with an output that is large enough.

fastcrypto-tbls/src/ecies.rs
```diff
@@ -0,0 +1,221 @@
...
+use fastcrypto::error::FastCryptoError;
+use fastcrypto::groups::bls12381::G1Element;
+use fastcrypto::groups::{GroupElement, HashToGroupElement, Scalar};
+use fastcrypto::hmac::{hkdf_sha3_256, HkdfIkm};
+use fastcrypto::traits::{AllowedRng, Cipher, ToFromBytes};
+use serde::{Deserialize, Serialize};
+use typenum::consts::{U16, U32};
+
+///
+/// Simple ECIES encryption using a generic group and AES-256-counter.
```
kchalkias, line 15: thanks

fastcrypto-tbls/src/random_oracle.rs
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use digest::Digest;
+use serde::Serialize;
+use sha3::Sha3_512;
+
+/// Random Oracle from Sha3.
+/// - prefix should be globally unique.
+/// - evaluate serializes the given input and outputs SHA3-512(prefix_len as u32 | prefix | input).
```
kchalkias, line 10: why 512?

fastcrypto/src/groups/bls12381.rs
```diff
@@ -533,6 +534,19 @@ impl ScalarType for Scalar {
     }
 }
 
+impl HashToGroupElement for Scalar {
+    fn hash_to_group_element(bytes: &[u8]) -> Self {
+        let digest = Sha256::digest(bytes); // TODO: should this be 512?
```
kchalkias, line 539: Interestingly, for Ristretto we used 512 and here 256. Let's make them aligned

kchalkias (approved):
Just a short comment re hash function alignment. Great work!

fastcrypto-tbls/src/random_oracle.rs
benr-ml: Given a uniform string x of length l, x mod q has a bias of 2^-(l-|q|), and so assuming we want <2^-128 bias we need l 256+128 for all the groups we work with. 512 seems like the right size.

fastcrypto/src/groups/bls12381.rs
benr-ml: I think it should be 512 in both, but I used here 256 to be consistent with https://github.com/MystenLabs/fastcrypto/blob/main/fastcrypto/src/bls12381/min_sig/mskr.rs#L25-L26

@jonas-lj what was the reason we used 256 in mskr?

jonas-lj: It's a while back, but I think there was a discussion if it was sufficient with half length scalars for mskr, in which case SHA-256 truncated to 128 bits should be fine for that specific case.

PR #337, opened 2023-01-03
First version of DKG

benr-ml:


fastcrypto-tbls/src/dkg.rs
```diff
@@ -0,0 +1,404 @@
...
+    nodes: Nodes<EG>,
+    ecies_sk: ecies::PrivateKey<EG>,
+    ecies_pk: ecies::PublicKey<EG>,
+    vss_sk: PrivatePoly<G>,
+    vss_pk: PublicPoly<G>,
+    threshold: u32,
+    random_oracle: RandomOracle,
+}
+
+/// [EncryptedShare] holds the ECIES encryption of a share destined to the receiver.
```
kchalkias, line 47: Are you following a particular ECIES standard? ie apple uses ANSI X9.63 and ECIESEncryptionCofactorX963SHA256AESGCM etc.

fastcrypto-tbls/src/dkg.rs
```diff
@@ -0,0 +1,404 @@
...
+
+/// [DkgFirstMessage] holds all encrypted shares a dealer sends during the first phase of the
+/// protocol.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct FirstMessage<G: GroupElement, EG: GroupElement> {
+    pub sender: ShareIndex,
+    /// The encrypted shares created by the sender.
+    pub encrypted_shares: Vec<EncryptedShare<EG>>,
+    /// The commitment of the secret polynomial created by the sender.
+    // TODO: add a proof of possession/knowledge?
```
kchalkias, line 64: good point, do we know what other production ready implementations do?

fastcrypto-tbls/src/dkg.rs
```diff
@@ -0,0 +1,404 @@
...
+        let ecies_pk = ecies::PublicKey::<EG>::from_private_key(&ecies_sk);
+
+        // Check if the public key is in one of the nodes.
+        let curr_node = nodes
+            .iter()
+            .find(|n| n.pk == ecies_pk)
+            .ok_or(FastCryptoError::InvalidInput)?;
+
+        // Generate a secret polynomial and commit to it.
+        let vss_sk = PrivatePoly::<G>::rand(threshold - 1, rng);
```
kchalkias, line 131: do we need to check that threshold is in the expected range ie <= n?

fastcrypto-tbls/src/dkg.rs
```diff
@@ -0,0 +1,404 @@
...
+
+        FirstMessage {
+            sender: self.id,
+            encrypted_shares,
+            vss_pk: self.vss_pk.clone(),
+        }
+    }
+
+    /// 5. Process the first messages of exactly 'threshold' nodes and create the second message to
+    ///    be broadcast.
```
kchalkias, line 176: broadcasted

fastcrypto-tbls/src/dkg.rs
```diff
@@ -0,0 +1,404 @@
...
+            // Else, decrypt it.
+            let share = Self::decrypt_and_check_share(
+                &self.ecies_sk,
+                my_id,
+                &message.vss_pk,
+                encrypted_share.expect("checked above that is not None"),
+            );
+            match share {
+                Ok(sh) => {
+                    shares.insert(message.sender, sh);
```
kchalkias, line 229: do we need to detect messages from the same sender to avoid duplicates or we expect that the last insert always wins? Generally, as long as we use vectors vs hashmaps for to receive messages as function input, how do we deal with duplicated data at this stage (step)? in general do we perform a fail fast approach or just ignore duplicates?

kchalkias (commented):
Almost ready for approval, only a few clarification questions.

fastcrypto-tbls/src/dkg.rs
benr-ml: No, as I need specific curves and specific operations (e.g. the recovery package)

fastcrypto-tbls/src/dkg.rs
benr-ml: added a check

fastcrypto-tbls/src/dkg.rs
benr-ml: it's protocol dependent, i'm not sure we need it; will decide how to resolve it once i have a security proof for the protocol

fastcrypto-tbls/src/dkg.rs
benr-ml: added a check

kchalkias (approved):
Great job!

PR #341, opened 2023-01-04
AES encryption should not return result

jonas-lj:


fastcrypto/src/traits.rs
```diff
@@ -373,10 +373,12 @@ pub trait Nonce:
...
     type IVType: Nonce;
 
-    /// Encrypt `plaintext` and write result to `buffer` using the given IV
-    fn encrypt(&self, iv: &Self::IVType, plaintext: &[u8]) -> Result<Vec<u8>, FastCryptoError>;
+    /// Encrypt `plaintext` using the given IV and return the result.
+    fn encrypt(&self, iv: &Self::IVType, plaintext: &[u8]) -> Vec<u8>;
 
-    /// Decrypt `ciphertext` and write result to `buffer` using the given IV
+    /// Decrypt `ciphertext` using the given IV and return the result. An error may be returned in
+    /// CBC-mode if the ciphertext is not correctly padded, but in CTR mode this method always
```
benr-ml, line 380:
```suggestion
    /// CBC-mode if the ciphertext is not correctly padded, but in other modes this method always
```

kchalkias (commented):
Can we explain the motivation? What is a reason that `apply_keystream_b2b(plaintext, &mut buffer).unwrap();` would fail?
Should I guess we're safe because, an issue would only be related to mismatched IV sizes etc and our strict IV type logic prevents that?

Issue #325, opened 2022-12-28
Support ECVRF in Ristretto group


kchalkias:
Due to: https://github.com/MystenLabs/sui/issues/7038

PR #357, opened 2023-01-10
Add ECVRF

jonas-lj:
Closing #325

### DIFF
--- a/fastcrypto/src/groups/ristretto255.rs
+++ b/fastcrypto/src/groups/ristretto255.rs
@@ -134,6 +134,14 @@ impl RistrettoScalar {
     pub fn group_order() -> RistrettoScalar {
         RistrettoScalar(BASEPOINT_ORDER)
     }
+
+    /// Hash the given bytes in a uniformly random scalar.
+    pub fn hash_to_scalar<H: HashFunction<64>>(bytes: &[u8]) -> Self {
+        // Implementation copied from https://docs.rs/curve25519-dalek-ng/4.1.1/src/curve25519_dalek_ng/scalar.rs.html#629
+        RistrettoScalar(ExternalRistrettoScalar::from_bytes_mod_order_wide(
+            &H::digest(bytes).digest,
+        ))
+    }
 }
 
 impl From<u64> for RistrettoScalar {
--- a/fastcrypto/src/groups/ristretto255.rs
+++ b/fastcrypto/src/groups/ristretto255.rs
@@ -21,10 +21,11 @@ use curve25519_dalek_ng::traits::Identity;
 use derive_more::{Add, Div, From, Neg, Sub};
 use digest::Digest;
 use fastcrypto_derive::GroupOpsExtend;
-use serde::{de, Deserialize, Serialize};
+use serde::{de, Deserialize};
 use std::ops::{Div, Mul};
 
 const RISTRETTO_POINT_BYTE_LENGTH: usize = 32;
+const RISTRETTO_SCALAR_BYTE_LENGTH: usize = 32;
 
 /// Represents a point in the Ristretto group for Curve25519.
 #[derive(Default, Clone, Copy, Debug, PartialEq, Eq, From, Add, Sub, Neg, GroupOpsExtend)]
@@ -103,21 +104,7 @@ generate_bytes_representation!(
 );
 
 /// Represents a scalar.
-#[derive(
-    Clone,
-    Copy,
-    Debug,
-    Serialize,
-    Deserialize,
-    PartialEq,
-    Eq,
-    From,
-    Add,
-    Sub,
-    Neg,
-    Div,
-    GroupOpsExtend,
-)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, From, Add, Sub, Neg, Div, GroupOpsExtend)]
 pub struct RistrettoScalar(ExternalRistrettoScalar);
 
 impl RistrettoScalar {
@@ -197,3 +184,24 @@ impl HashToGroupElement for RistrettoScalar {
         Self(ExternalRistrettoScalar::from_bytes_mod_order_wide(&hash))
     }
 }
+
+impl ToFromByteArray<RISTRETTO_SCALAR_BYTE_LENGTH> for RistrettoScalar {
+    fn from_byte_array(
+        bytes: &[u8; RISTRETTO_SCALAR_BYTE_LENGTH],
+    ) -> Result<Self, FastCryptoError> {
+        Ok(RistrettoScalar(
+            ExternalRistrettoScalar::from_bytes_mod_order(*bytes),
+        ))
+    }
+
+    fn to_byte_array(&self) -> [u8; RISTRETTO_SCALAR_BYTE_LENGTH] {
+        self.0.to_bytes()
+    }
+}
+
+serialize_deserialize_with_to_from_byte_array!(RistrettoScalar);
+generate_bytes_representation!(
+    RistrettoScalar,
+    RISTRETTO_SCALAR_BYTE_LENGTH,
+    RistrettoScalarAsBytes
+);
--- a/fastcrypto/src/groups/ristretto255.rs
+++ b/fastcrypto/src/groups/ristretto255.rs
@@ -5,6 +5,7 @@
 //! prime order 2^{252} + 27742317777372353535851937790883648493 built over Curve25519.
 
 use crate::groups::{GroupElement, HashToGroupElement, Scalar};
+use crate::hash::Sha512;
 use crate::serde_helpers::BytesRepresentation;
 use crate::serde_helpers::ToFromByteArray;
 use crate::traits::AllowedRng;
@@ -19,7 +20,6 @@ use curve25519_dalek_ng::ristretto::RistrettoPoint as ExternalRistrettoPoint;
 use curve25519_dalek_ng::scalar::Scalar as ExternalRistrettoScalar;
 use curve25519_dalek_ng::traits::Identity;
 use derive_more::{Add, Div, From, Neg, Sub};
-use digest::Digest;
 use fastcrypto_derive::GroupOpsExtend;
 use serde::{de, Deserialize};
 use std::ops::{Div, Mul};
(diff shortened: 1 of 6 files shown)


jonas-lj:
I could only find one crate supporting ECVRF, but it's not maintained and it's LGPL, so I wrote an implementation my self based on the draft of the IETF specs (https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vrf-04).

fastcrypto/src/vrf.rs
```diff
@@ -180,7 +180,7 @@ pub mod ecvrf {
         fn from(c: &Challenge) -> Self {
             let mut scalar = [0u8; 32];
             scalar[..C_LEN].copy_from_slice(&c.0);
-            RistrettoScalar::from_bytes_mod_order(scalar)
+            RistrettoScalar::from_byte_array(&scalar).unwrap()
```
kchalkias, line 183: Can this fail? if not please add a comment re why

fastcrypto/src/groups/ristretto255.rs
```diff
@@ -98,42 +86,37 @@ impl TryFrom<&[u8]> for RistrettoPoint {
...
+pub struct RistrettoScalar(ExternalRistrettoScalar);
+
+impl RistrettoScalar {
     /// The order of the base point.
     pub fn group_order() -> RistrettoScalar {
         RistrettoScalar(BASEPOINT_ORDER)
     }
+
+    /// Construct a [RistrettoScalar] by reducing a 64-byte little-endian integer modulo the group order.
+    pub fn from_bytes_mod_order_wide(bytes: [u8; 64]) -> Self {
```
kchalkias, line 117: ooc, why is this not a reference? `bytes: [u8; 64]`

fastcrypto/src/groups/ristretto255.rs
```diff
@@ -184,6 +167,27 @@ impl HashToGroupElement for RistrettoScalar {
         hasher.update(bytes);
         let mut hash = [0u8; 64];
         hash.copy_from_slice(hasher.finalize().as_slice());
-        Self(ExternalRistrettoScalar::from_bytes_mod_order_wide(&hash))
+        Self::from_bytes_mod_order_wide(hash)
```
kchalkias, line 170: ditto (why not reference?)

fastcrypto/src/tests/vrf_tests.rs
```diff
@@ -0,0 +1,53 @@
...
+    let sk_serialized = bincode::serialize(&kp.sk).unwrap();
+    let sk_reconstructed = bincode::deserialize(&sk_serialized).unwrap();
+    assert_eq!(&kp.sk, &sk_reconstructed);
+
+    let input = b"Hello, world!";
+    let (output, proof) = kp.output(input);
+    let proof_serialized = bincode::serialize(&proof).unwrap();
+
+    // A proof consists of a point, a challenge (half length of field elements) and a field element.
+    assert_eq!(32 + 16 + 32, proof_serialized.len());
```
kchalkias, line 47: minor: use `assert_eq!(RISTRETTO_POINT_BYTE_LENGTH + RISTRETTO_SCALAR_BYTE_LENGTH / 2 + RISTRETTO_SCALAR_BYTE_LENGTH, proof_serialized.len());`

fastcrypto/src/vrf.rs
```diff
@@ -0,0 +1,268 @@
...
+    use crate::groups::{GroupElement, Scalar};
+    use crate::hash::{HashFunction, Sha512};
+    use crate::serde_helpers::ToFromByteArray;
+    use crate::traits::AllowedRng;
+    use crate::vrf::{VRFKeyPair, VRFPrivateKey, VRFProof, VRFPublicKey};
+    use elliptic_curve::hash2curve::{ExpandMsg, Expander};
+    use serde::{Deserialize, Serialize};
+
+    /// draft-irtf-cfrg-vrf-15 specifies suites for suite-strings 0x00-0x04 and notes that future
+    /// designs should specify a different suite_string constant, so we use 0x05 here.
```
kchalkias, line 80: it's not always clear or true that future designs won't skip a value, is there a reason for preferring a future (potentially never assigned value)?

fastcrypto/src/groups/ristretto255.rs
jonas-lj: You're right. That should definitely be a reference.

fastcrypto/src/vrf.rs
jonas-lj: draft-irtf-cfrg-vrf-15 specifies that:
> If future designs need to specify variants (e.g., additional ciphersuites) of the RSA-FDH-VRF or the ECVRF in this document, then, to avoid the possibility that an adversary can obtain a VRF output under one variant, and then claim it was obtained under another variant, they should specify a different suite_string constant.

If I understand this correctly, we just need to choose any other suite string than the ones in the draft. But we can add a note that we update the suite_string according to new releases of the specifications.

fastcrypto/src/vrf.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// the other suites uses half the length, so we do the same here.
+    const C_LEN: usize = 16;
+
+    /// Default hash function
+    type H = Sha512;
+
+    /// Domain separation tag used in ecvrf_encode_to_curve
+    const DST: &[u8; 43] = b"ECVRF_ristretto255_XMD:SHA-512_R255MAP_RO_5";
+
+    const COFACTOR: u64 = 8;
```
benr-ml, line 93: let's move this to ristretto255.rs

fastcrypto/src/vrf.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    #[derive(Serialize, Deserialize, PartialEq, Eq, Debug)]
+    pub struct ECVRFPrivateKey(RistrettoScalar);
+
+    impl VRFPrivateKey for ECVRFPrivateKey {
+        type PublicKey = ECVRFPublicKey;
+    }
+
+    impl ECVRFPrivateKey {
+        /// Generate nonce from binary string. See section 5.4.2.2. of draft-irtf-cfrg-vrf-15.
```
benr-ml, line 140: why is it called a nonce and not scalar?

jonas-lj: To follow the terminology from the rfc draft (https://www.ietf.org/archive/id/draft-irtf-cfrg-vrf-15.html#name-ecvrf-nonce-generation), but it is a scalar so we could also call it that if you think it makes more sense.

fastcrypto/src/vrf.rs
```diff
@@ -0,0 +1,268 @@
...
+    use crate::vrf::{VRFKeyPair, VRFPrivateKey, VRFProof, VRFPublicKey};
+    use elliptic_curve::hash2curve::{ExpandMsg, Expander};
+    use serde::{Deserialize, Serialize};
+
+    /// draft-irtf-cfrg-vrf-15 specifies suites for suite-strings 0x00-0x04 and notes that future
+    /// designs should specify a different suite_string constant, so we use 0x05 here.
+    const SUITE_STRING: [u8; 1] = [0x05]; // TODO: Update this with future releases of the draft or final version.
+
+    /// Length of challenge. Must not exceed the length of field elements which is 32 in this case and
+    /// the other suites uses half the length, so we do the same here.
```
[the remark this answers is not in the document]
jonas-lj: You're right, but the rfc doesn't mention a minimum length. But I'll add a note that the chosen length is aligned with the other suites in the rfc.

fastcrypto/src/vrf.rs
```diff
@@ -0,0 +1,268 @@
...
+    ) -> Result<(), FastCryptoError> {
+        self.verify(input, public_key)?;
+        if self.to_hash() != output {
+            return Err(FastCryptoError::GeneralError);
+        }
+        Ok(())
+    }
+
+    /// Compute the output of the VRF with this proof.
+    fn to_hash(&self) -> [u8; OUTPUT_SIZE];
```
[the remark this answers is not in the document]
jonas-lj: I don't think the RFC allows it to not return a hash, and it's also the case for the other construction, RSA-FDH-VRF, that it will always return. In what ways do you think it can fail?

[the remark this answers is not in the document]
benr-ml: I guess it's just a nit. I meant for general VRF functions, not only ones defined by the RFC.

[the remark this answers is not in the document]
jonas-lj: I see. I don't think it'll be too much trouble to change the output to a Result if we at some point implement one of those constructions.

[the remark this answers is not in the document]
kchalkias: Are we going to resolve this and return a Result?

fastcrypto/src/vrf.rs
```diff
@@ -0,0 +1,263 @@
...
+
+    #[derive(Serialize, Deserialize, PartialEq, Eq, Debug)]
+    pub struct ECVRFPrivateKey(RistrettoScalar);
+
+    impl VRFPrivateKey for ECVRFPrivateKey {
+        type PublicKey = ECVRFPublicKey;
+    }
+
+    impl ECVRFPrivateKey {
+        /// Generate nonce from binary string. See section 5.4.2.2. of draft-irtf-cfrg-vrf-15.
```
benr-ml, line 135:
```suggestion
        /// Generate scalar/nonce from binary string. See section 5.4.2.2. of draft-irtf-cfrg-vrf-15.
```

fastcrypto/src/vrf.rs
```diff
@@ -0,0 +1,263 @@
...
+
+            let mut bytes = [0u8; H::OUTPUT_SIZE];
+            expanded_message.fill_bytes(&mut bytes);
+            RistrettoPoint::map_to_point::<H>(&bytes)
+        }
+
+        /// Implements ECVRF_validate_key which checks the validity of a public key. See section 5.4.5
+        /// of draft-irtf-cfrg-vrf-15.
+        fn valid(&self) -> bool {
+            self.0 * RistrettoScalar::from(COFACTOR) != RistrettoPoint::zero()
```
benr-ml, line 123: Do we need this for Ristretto? (actually all cofactor related code here)

fastcrypto/src/vrf.rs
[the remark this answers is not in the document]
jonas-lj: If I understand Ben's like correctly, we're going to keep it as it is for now and only introduce the Result when needed.

fastcrypto/src/vrf.rs
jonas-lj: No, you're right! The ristretto group already factors out the cofactor, so we just need to check that it's not the identity.

kchalkias (approved):
I believe we can approve to unblock and then come back again for hardening tests and full test edge cases coverage.

PR #362, opened 2023-01-11
crypto: PartialEq compares bytes representation

joyqvq:
make sure the equality is checked on bytes. otherewise the struct may not have oncecell initialized and fails equality check.


kchalkias (approved):
Approving, but what is the rationale re comparing against the bytes?

joyqvq:
> Approving, but what is the rationale re comparing against the bytes?

alternatively we can compare every other field of the struct withouts the bytes. because two objects are considered equal regardless of `bytes: OnceCell` is initialized or not

PR #363, opened 2023-01-12
crypto: use macro to derive ser/de from ToFromBytes

joyqvq:


fastcrypto/src/serde_helpers.rs
```diff
@@ -168,16 +168,24 @@ macro_rules! serialize_deserialize_with_to_from_bytes {
     ($type:ty) => {
         impl ::serde::Serialize for $type {
             fn serialize<S: ::serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
-                serializer.serialize_str(&self.encode_base64())
+                match serializer.is_human_readable() {
+                    true => serializer.serialize_str(&self.encode_base64()),
+                    false => serializer.serialize_bytes(&self.as_bytes()),
```
benr-ml, line 173: pls use SerializationHelper as above
also in the deserializer

joyqvq: came across a limitation when using this in sui. since the length as to be known for SerializationHelper, this does not work with enums where each member has variable length. 

```
pub enum Signature {
    Ed25519SuiSignature,
    Secp256k1SuiSignature,
    Secp256r1SuiSignature,
}
```

`serializer.serialize_bytes(&self.as_bytes())` would be one way to work around this. Ideally I want to use this macro to derive ser/de for enum as well.

benr-ml (approved):
1. Just a thought, your call - maybe it'd nice to add the length type to trait ToFromBytes (like we have in ToFromByteArray) and use it in the macro instead of adding the length as a parameter.
2. Please check locally that Sui is not trivially broken because of this change before merging.

fastcrypto/src/serde_helpers.rs
benr-ml: You mean that the macro won't work for this enum, right? but it will work for each of its types.
That's ok, the goal of the macro is to have a consistent serde for the crypto types, ones that have canonical representations.

joyqvq: :+1 i will have a separate macro customized for sui then. agreed on having serde in fastcrypto to be the most concise one with known length.

kchalkias (approved):
just resolve conflict re secp

kchalkias (approved):

PR #365, opened 2023-01-16
Update Ed25519 to be used for authority signatures

jonas-lj:


kchalkias (commented):
is this still required?

kchalkias (commented):
@jonas-lj is this mergeable or just a reference commitID used when switching algorithms during benchmarks?

PR #369, opened 2023-01-19
Move recoverable secp256k1 signature scheme to submodule

jonas-lj:
This PR moves the Secp256k1 signature scheme to a new secp256k1::recoverable module and renames relevant types. There are no functional changes.

This is done to prepare for the addition of non-recoverable signatures.


kchalkias (approved):

kchalkias:
Please ensure that when this and the other PRs merge that we update Sui as well. Note that Sui accounts will only support non-recoverable sigs, and I guess at contract level we want to support both?

PR #373, opened 2023-01-23
Add recover_hashed to secp256_k1

jonas-lj:
Also ensure that the right hash function is used internally in tests.


fastcrypto/src/secp256k1/recoverable.rs
```diff
@@ -12,7 +12,7 @@
 //! let kp = Secp256k1RecoverableKeyPair::generate(&mut thread_rng());
 //! let message: &[u8] = b"Hello, world!";
 //! let signature = kp.sign(message);
-//! assert!(kp.public().verify(message, &signature).is_ok());
+//! assert_eq!(&signature.recover(message).unwrap(), kp.public());
```
jonas-lj, line 15: The doctest is also changed to show a public key recovery rather than a verification.

joyqvq (approved):

PR #379, opened 2023-01-25
Add ecdsa over secp256r1 with nonrecoverable signatures

jonas-lj:


fastcrypto/src/secp256r1/mod.rs
```diff
@@ -1,4 +1,452 @@
...
+    type Sig = Secp256r1Signature;
+    const LENGTH: usize = PRIVATE_KEY_SIZE;
+}
+
+impl ToFromBytes for Secp256r1PrivateKey {
+    fn from_bytes(bytes: &[u8]) -> Result<Self, FastCryptoError> {
+        match ExternalSecretKey::try_from(bytes) {
+            Ok(privkey) => Ok(Secp256r1PrivateKey {
+                privkey,
+                bytes: OnceCell::with_value(<[u8; 32]>::try_from(bytes).unwrap()),
```
joyqvq, line 208: replace 32 with PRIVATE_KEY_SIZE

fastcrypto/src/secp256r1/mod.rs
```diff
@@ -1,4 +1,452 @@
...
+impl<'de> Deserialize<'de> for Secp256r1Signature {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let data: Vec<u8> = Vec::deserialize(deserializer)?;
+        <Secp256r1Signature as Signature>::from_bytes(&data)
+            .map_err(|e| de::Error::custom(e.to_string()))
+    }
+}
```
joyqvq, line 276: the ser/de impl here is OK here - i will clean up with the macro for all crates in a separate pr

fastcrypto/src/secp256r1/mod.rs
```diff
@@ -1,4 +1,452 @@
...
+
+impl Eq for Secp256r1Signature {}
+
+impl Display for Secp256r1Signature {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        write!(f, "{}", Base64::encode(self.as_ref()))
+    }
+}
+
+impl Default for Secp256r1Signature {
```
joyqvq, line 329: do we really use the default anywhere?

fastcrypto/src/secp256r1/mod.rs
```diff
@@ -1,4 +1,452 @@
...
+        }
+    }
+
+    fn generate<R: AllowedRng>(rng: &mut R) -> Self {
+        let privkey = ExternalSecretKey::random(rng);
+        let pubkey = ExternalPublicKey::from(&privkey);
+
+        Secp256r1KeyPair {
+            name: Secp256r1PublicKey {
+                pubkey,
```
joyqvq, line 402: can just use `privkey.into()` here?

joyqvq (approved):
some nits and lgtm

fastcrypto/src/secp256r1/mod.rs
jonas-lj: Yea, it's a bit annoying because it's a very odd trait to impl for both signatures, but it is used in sui, and I'm not sure whether it's possible to change the Sui code to not use default, but it would be great. The same is also true for public keys.

Issue #360, opened 2023-01-10
Helper fun to convert ECDSA recoverable to non recoverable and vice versa


kchalkias:
This has been requested multiple times, especially from non-recoverable to recoverable, ie iPhone enclaves are only outputting non-recoverable.
any of @joyqvq and @jonas-lj who are familiar with that repo part.

jonas-lj:
Recoverable to non-recoverable is already part of #292, but I'll make sure to add the non-recoverable to recoverable also.

jonas-lj:
Won't the easiest way to do this be to try the different possibilities for recovery id and then compare with a public key or an address? This is at least the way it's done in https://www.secg.org/sec1-v2.pdf sec 4.1.6.

Also, I'll add this in another PR. The PR of #292 is large enough as it is...

joyqvq:
agreed, trying all 4 (but more like 3) recovery id should work

PR #389, opened 2023-02-01
Add converters for recoverable and nonrecoverable k1 signature

jonas-lj:
Part of #360

### DIFF
--- a/fastcrypto/src/secp256k1/recoverable.rs
+++ b/fastcrypto/src/secp256k1/recoverable.rs
@@ -15,6 +15,7 @@
 //! assert_eq!(&signature.recover(message).unwrap(), kp.public());
 //! ```
 
+use crate::secp256k1::{Secp256k1PublicKey, Secp256k1Signature};
 use crate::{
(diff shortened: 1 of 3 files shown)


fastcrypto/src/secp256k1/recoverable.rs
```diff
@@ -295,6 +305,35 @@ impl Default for Secp256k1RecoverableSignature {
     }
 }
 
+impl Secp256k1RecoverableSignature {
```
benr-ml, line 308: it's the same code as for R1, can it be a generic or a trait like
TryFromNonRecoverable {
type sig;
type pk;
 pub fn try_from_nonrecoverable () ...
 }

jonas-lj: There's a lot of duplicated code between the two already, so I'd rather fix that in another PR.

jonas-lj: I've added a new issue #393

benr-ml (approved):

PR #392, opened 2023-02-01
Add optimised multiscalar multiplication to the ristretto group

jonas-lj:


fastcrypto/src/vrf.rs
```diff
@@ -239,8 +239,12 @@ pub mod ecvrf {
             let h = public_key.ecvrf_encode_to_curve(alpha_string);
 
             let challenge = RistrettoScalar::from(&self.c);
-            let u = RistrettoPoint::generator() * self.s - public_key.0 * challenge;
-            let v = h * self.s - self.gamma * challenge;
+            let u = RistrettoPoint::multiscalar_mul(
+                [self.s, -challenge],
+                [RistrettoPoint::generator(), public_key.0],
+            )
+            .unwrap();
```
benr-ml, line 246: the code above
`.map(RistrettoPoint).ok_or(FastCryptoError::GeneralError)`
implies that multiscalar_mul can fail in some cases other than if the arguments have different lengths.

jonas-lj: Yea, that probably need a comment. It'll happen if some of the elements are None.

jonas-lj: Actually there's also a non-optional version that we can use instead.

fastcrypto/src/groups/ristretto255.rs
```diff
@@ -51,6 +51,27 @@ impl RistrettoPoint {
     pub fn decompress(bytes: &[u8; 32]) -> Result<Self, FastCryptoError> {
         RistrettoPoint::try_from(bytes.as_slice())
     }
+
+    /// Compute the linear combination of the given scalars and points. An error will be returned if
+    /// the sizes does not match.
```
kchalkias, line 56: do~es~ not match

fastcrypto/src/vrf.rs
```diff
@@ -239,8 +239,12 @@ pub mod ecvrf {
...
+            let v = RistrettoPoint::multiscalar_mul([self.s, -challenge], [h, self.gamma]).unwrap();
```
kchalkias, line 247: As we're planning to remove or justify any unwraps across our codebase (unless they can absolutely never happen) we generally prefer Result<> to avoid panics, no?

jonas-lj: That makes sense. I've added ?'s to make sure any errors are propagated, but since an error is only returned if the sizes of the iterables do not match, it will never happen here anyway.

kchalkias (approved):

Issue #345, opened 2023-01-08
[serialization] BLS12381AggregateSignature should not have an option<signature>


benr-ml:
It results in a non standard serialized representation, which is actually returned also to users via APIs.

Alternatives:
- use inf as the default/empty signature
- do not implement `Default` and remove `Option<sig>`. Instead, modify NW's structs to use `Option<BLS12381AggregateSignature>`.

kchalkias:
fixed

PR #396, opened 2023-02-02
crypto: Use infinity for bls sig and aggr sig without option

joyqvq:


kchalkias (approved):
What is our stance if someone tries to verify against an infinity signature? we fail?

fastcrypto/src/bls12381/mod.rs
```diff
@@ -350,19 +349,11 @@ impl Signature for BLS12381Signature {
...
-        // TODO: improve this!
-        let ikm: [u8; 32] = [
-            0x93, 0xad, 0x7e, 0x65, 0xde, 0xad, 0x05, 0x2a, 0x08, 0x3a, 0x91, 0x0c, 0x8b, 0x72,
-            0x85, 0x91, 0x46, 0x4c, 0xca, 0x56, 0x60, 0x5b, 0xb0, 0x56, 0xed, 0xfe, 0x2b, 0x60,
-            0xa6, 0x3c, 0x48, 0x99,
-        ];
-
-        let sk = blst::SecretKey::key_gen(&ikm, &[]).unwrap();
+        let mut infinity: [u8; $sig_length] = [0; $sig_length];
+        infinity[0] = 0xc0;
```
benr-ml, line 353: what is this value, is that based on the rfc?
let's add a comment explaining that

benr-ml (approved):
Awesome!

fastcrypto/src/bls12381/mod.rs
joyqvq: > What is our stance if someone tries to verify against an infinity signature? we fail?

yes if it verifies against any list of pks it will fail.

joyqvq:
> What is our stance if someone tries to verify against an infinity signature? we fail?

yes if it verifies against any list of pks it will fail.

kchalkias (approved):

benr-ml:
Fixes #345

PR #398, opened 2023-02-07
Ed25519 speccheck tests

### LINKED ISSUE #17: [crypto] Ed25519 security
Feedback from writing MystenLabs/narwhal#32 on our ed25519 implementation:
- the "master" record of our keys is now a ed25519_dalek type, which is zeroized properly on drop => can we ensure this sort of thing is kept as a convention for future impls? perhaps with a unit test?
- we do not perform any verifications on deserialization of a key, including:
    * canonical scalar representation for PrivKey,
    * small order element for Public keys and keypairs.
- we should probably add warnings about small order components in keys, as they will disagree with single-signature verification.

Inspirations for accomplishing the above:
- https://github.com/diem/diem/blob/main/crates/diem-crypto/src/ed25519.rs
- https://eprint.iacr.org/2020/1244.pdf
- https://github.com/novifinancial/ed25519-speccheck (contains test vectors!)

jonas-lj:
Fixes #17. But I'm not sure it's relevant anymore. What do you think @kchalkias?

### DIFF
--- a/fastcrypto/src/tests/ed25519_tests.rs
+++ b/fastcrypto/src/tests/ed25519_tests.rs
@@ -716,3 +716,65 @@ proptest! {
         assert_eq!(kp.public(), deserialized.public());
     }
 }
+
+#[test]
+fn ed25519_speccheck() {
+    let msgs = [
+        "8c93255d71dcab10e8f379c26200f3c7bd5f09d9bc3068d3ef4edeb4853022b6",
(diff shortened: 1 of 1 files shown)


kchalkias (commented):
Is this ready for review?

kchalkias (commented):
Hey @jonas-lj Can you please post the output flags to the PR description? Then I can properly review the outcome and if it conforms to the canonical expectations.

kchalkias (approved):
Accepting and will update with the theory analysis re expected results in an upcoming comments PR.

kchalkias (approved):

PR #405, opened 2023-02-07
Add ECVRF cli

jonas-lj:
Closing [#8069](https://github.com/MystenLabs/sui/issues/8069) in sui


cryptotool/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "cryptotool"
```
benr-ml, line 2: perhaps let's put this under `fastcrypto-cli`, and name the tool `ecvrf-cli`?

fastcrypto-cli/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "fastcrypto-cli"
+version = "0.1.0"
+license = "Apache-2.0"
+authors = ["Mysten Labs <build@mystenlabs.com>"]
+edition = "2021"
+publish = ["crates-io"]
+description = "zk-proof library used at Mysten Labs"
```
benr-ml, line 8: pls update

fastcrypto-cli/src/ecvrf.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// Hex-encoded Sha512 hash of the proof. Must be 64 bytes.
+    #[clap(short, long)]
+    output: String,
+
+    /// Encoding of the proof to verify. Must be 80 bytes.
+    #[clap(short, long)]
+    proof: String,
+
+    /// Hex encoding of the input string used to generate the proof. May be of arbitrary length but
+    /// should ideally be at least 16 bytes.
```
benr-ml, line 46: why is that preferred?

fastcrypto-cli/src/ecvrf.rs
```diff
@@ -0,0 +1,146 @@
...
+                    Error::new(ErrorKind::Other, "Failed to serialize secret key.")
+                })?);
+            let pk_string =
+                hex::encode(bincode::serialize(&keypair.pk).map_err(|_| {
+                    Error::new(ErrorKind::Other, "Failed to serialize public key.")
+                })?);
+
+            let mut result = "Secret key: ".to_string();
+            result.push_str(&sk_string);
+            result.push_str("\nPublic key: ");
```
benr-ml, line 83: Verification Key (like the name used below)

fastcrypto-cli/src/ecvrf.rs
jonas-lj: If the prover can predict the input or it has very low entropy, the prover can create a keypair giving a specific output. @kchalkias had a tweet about it last week. But this warning should either be omitted be more clearly written.

fastcrypto-cli/src/ecvrf.rs
jonas-lj: I've used public key everywhere instead to align with the fastcrypto api.

fastcrypto-cli/src/ecvrf.rs
jonas-lj: I've removed it. This warning is use case specific and should be somewhere else.

benr-ml (approved):

PR #411, opened 2023-02-08
Refactor recoverable r1

### LINKED ISSUE #393: Refactor code in R1 and K1 implementations
it's the same code as for R1, can it be a generic or a trait like
```
TryFromNonRecoverable {
type sig;
type pk;
 pub fn try_from_nonrecoverable () ...
 }
```
_Originally posted by @benr-ml in https://github.com/MystenLabs/fastcrypto/pull/389#discussion_r1093180715_

---
There's a lot of redundant code in the two recoverable schemes since the keys are the same. This should be removed.

jonas-lj:
Closes #393. Depends on #409 which should be merged first.

### DIFF
--- a/fastcrypto/src/traits.rs
+++ b/fastcrypto/src/traits.rs
@@ -219,13 +219,7 @@ pub trait RecoverableSigner {
     type Sig: RecoverableSignature<Signer = Self, PubKey = Self::PubKey>;
 
     /// Sign as a recoverable signature.
-    fn try_sign_recoverable(&self, msg: &[u8]) -> Result<Self::Sig, FastCryptoError>;
-
-    /// Sign as a recoverable signature.
-    fn sign_recoverable(&self, msg: &[u8]) -> Self::Sig {
-        self.try_sign_recoverable(msg)
-            .expect("Signature operation failed")
-    }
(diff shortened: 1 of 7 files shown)


fastcrypto/src/traits.rs
```diff
@@ -219,7 +219,25 @@ pub trait RecoverableSigner {
     type Sig: RecoverableSignature<Signer = Self, PubKey = Self::PubKey>;
 
     /// Sign as a recoverable signature.
-    fn sign_recoverable(&self, msg: &[u8]) -> Self::Sig;
+    fn try_sign_recoverable(&self, msg: &[u8]) -> Result<Self::Sig, FastCryptoError>;
```
benr-ml, line 222: Let's start using FastCryptoResult<_> :)

benr-ml, line 222: when can this function fail? (or, why is it try_?)

jonas-lj: I was just wondering that myself. The answer is that it can't fail and that the function should just be sign_recoverable. The reason the error was there in the first place was to handle the situations where k = 0 or s = 0 during the ecdsa signing, both which have negligible probability since we use RFC6979 to generate the k.

benr-ml (approved):

PR #415, opened 2023-02-09
Remove more external traits

jonas-lj:
This removes the external `signature::Signer` and `signature::Signature` traits. The `Signer` trait is replaced with a local trait with the same name, and the `Signature` trait is incorporated into the `Authenticator` trait.

Besides some refactoring, the only change is that the `sign` and `try_sign` functions in `signature::Signer` is replaced with just a `sign` function. However, `sign` in the `signature::Signer` trait defaults to calling `try_sign` and panic on fail and none of our signature schemes can error during signing anyway, so this is doesn't really change anything.

This also gets rid of `signature::Error` for all signature schemes. It is still used in the bulletproofs though.

### DIFF
--- a/fastcrypto/src/traits.rs
+++ b/fastcrypto/src/traits.rs
@@ -190,6 +190,7 @@ pub trait Authenticator:
 /// Trait impl'd by a key/keypair that can create signatures.
 ///
 pub trait Signer<Sig: Authenticator> {
+    /// Create a new signature over a message.
     fn sign(&self, msg: &[u8]) -> Sig;
 }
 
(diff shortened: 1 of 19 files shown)


fastcrypto/src/traits.rs
```diff
@@ -188,13 +180,20 @@ pub trait SigningKey: ToFromBytes + Serialize + DeserializeOwned + Send + Sync +
 /// to the ones on its associated types for private key and public key material.
 ///
 pub trait Authenticator:
-    signature::Signature + Display + Serialize + DeserializeOwned + Send + Sync + 'static + Clone
+    ToFromBytes + Display + Serialize + DeserializeOwned + Send + Sync + 'static + Clone
```
benr-ml, line 183: do we really need this in the trait?

jonas-lj: Do you mean the `ToFromBytes` trait? The `as_bytes` function (which calls `as_ref`) is definitely being called a lot of places, so I think we need it until we introduce a cached object of similar. Or what do you suggest?

benr-ml: let's leave it for another pr

benr-ml (approved):

PR #418, opened 2023-02-09
Ecdsa batch benchmarks

jonas-lj:


benr-ml (approved):

fastcrypto/benches/crypto.rs
```diff
@@ -236,6 +236,16 @@ mod signature_benches {
         static BATCH_SIZES: [usize; 5] = [4, 8, 16, 32, 64];
         let mut group: BenchmarkGroup<_> = c.benchmark_group("Verify batch");
         for size in BATCH_SIZES.iter() {
+            verify_batch_signatures_single::<Secp256k1KeyPair, _>(
+                "Secp256k1_batch",
+                *size,
+                &mut group,
+            );
+            verify_batch_signatures_single::<Secp256r1KeyPair, _>(
+                "Secp256r1_batch",
```
kchalkias, line 245: I guess ECDSA batch is in practice a for loop between N ECDSA sigs

kchalkias (approved):

fastcrypto/benches/crypto.rs
jonas-lj: Yes, that's true.

PR #419, opened 2023-02-09
Remove redundant benchmarks

jonas-lj:
Only benchmark batched verification for BLS (which both aggregates and verifies) for more fair comparison with the other schemes.

### DIFF
--- a/fastcrypto/benches/crypto.rs
+++ b/fastcrypto/benches/crypto.rs
@@ -132,28 +132,6 @@ mod signature_benches {
         );
     }
 
-    fn verify_aggregate_signatures_single<
-        KP: KeyPair,
-        A: AggregateAuthenticator<Sig = KP::Sig, PrivKey = KP::PrivKey, PubKey = KP::PubKey>,
-        M: measurement::Measurement,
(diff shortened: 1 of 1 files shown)


kchalkias (commented):
I want us to reconsider this, typically another entity is aggregating (ie the tx submitter forms a cert from receiving signatures from each validator) and another verifies the aggregated (the validators or blockchain users). Thus just verifying an already aggregated sig is still important.

jonas-lj:
> I want us to reconsider this, typically another entity is aggregating (ie the tx submitter forms a cert from receiving signatures from each validator) and another verifies the aggregated (the validators or blockchain users). Thus just verifying an already aggregated sig is still important.

That definitely makes sense. I was just confused when revisiting the benchmarks after some time because it was a bit unclear to me, what the difference between "batched" and "aggregated" was here, but this could be solved with documentation.

kchalkias (commented):
what's the verdict eventually? IMO we should still keep these benchmarks for reference.

jonas-lj:
Sure, let's keep them.

Issue #407, opened 2023-02-07
Add code coverage GH action


benr-ml:
Some options:
1/ Like https://github.com/MystenLabs/sui/blob/ac7a5685ea8274c9304b139a973e6a16e9b0446d/.github/workflows/codecov.yml
(though currently disabled)
2/ `lvm-cov` with `--hide-instantiations`, see for example https://github.com/nextest-rs/nextest/blob/main/.github/workflows/coverage.yml (can work with nextest - https://nexte.st/book/test-coverage.html)

PR #421, opened 2023-02-10
Create code coverage action

jonas-lj:
Closing #407


benr-ml:
Excellent,
- Can you add a few words to the GH issue on why this option was preferred (as opposed to the one used in Sui)?
- Does the upload to codecov work? Can you share an example of the output?

jonas-lj:
> Excellent,
> 
> * Can you add a few words to the GH issue on why this option was preferred (as opposed to the one used in Sui)?
> * Does the upload to codecov work? Can you share an example of the output?

I don't think there was much of a difference. Both approaches used codecov.io to generate the reports, but taiki (which is used here) was a bit easier to configure. The output is, as mentioned, uploaded to [codecov.io](https://app.codecov.io/gh/MystenLabs/fastcrypto/tree/codecov). The workflow works, and we have an 87% codecov in fastcrypto.

benr-ml (approved):

PR #450, opened 2023-02-23
Use blst single threaded

jonas-lj:
Parallelisation in sui has been made complicated by blst spawning threads, so I suggest we disable it altogether. Ideally, we should be able to enable multi-threading by setting a feature flag, but since the feature in blst is "negative" (`no-threads`) this is not possible. If there's a need for multithreaded verification later we should be able to come up with a solution.

Benchmarking shows that using multiple threads in blst may speed-up verification ~25% but this comes at the cost of using multiple threads.


kchalkias (approved):
did we do any profiling to ensure that this change works as expected.

PR #465, opened 2023-02-28
Small cleanups of ECDSA

benr-ml:


fastcrypto/src/secp256r1/mod.rs
```diff
@@ -408,21 +408,7 @@ impl Signer<Secp256r1Signature> for Secp256r1KeyPair {
 
 impl From<Secp256r1PrivateKey> for Secp256r1KeyPair {
     fn from(secret: Secp256r1PrivateKey) -> Self {
-        let name = Secp256r1PublicKey::from(&secret);
-        Secp256r1KeyPair { name, secret }
-    }
-}
-
-impl zeroize::Zeroize for Secp256r1KeyPair {
```
kchalkias, line 416: What is the main reason of not zeroizing the keypair anymore?

jonas-lj (approved):
Looks good to me (if @kchalkias comment is addressed)

fastcrypto/src/secp256r1/mod.rs
benr-ml: Drop of the private key takes care of that already

kchalkias (commented):
@benr-ml so if I load a KeyPair object I can invoke zeroization in its private key part element?

benr-ml:
> @benr-ml so if I load a KeyPair object I can invoke zeroization in its private key part element? Indeed it works, assuming we document and advertise it properly.

Maybe I don't understand your point - The ZeroizeOnDrop takes care of that for you, do you mean what if someone wants to zero it explicitly in other cases? if so, where do you think it's needed?

PR #469, opened 2023-03-01
Allow multiple public inputs in verify api

jonas-lj:
The current api for the Groth16 verifier implicitly expected only a single public input. This PR fixes this.


fastcrypto-zkp/src/api.rs
```diff
@@ -31,8 +33,13 @@ pub fn verify_groth16_in_bytes(
     proof_public_inputs_as_bytes: &[u8],
     proof_points_as_bytes: &[u8],
 ) -> Result<bool, FastCryptoError> {
-    let x = BlsFr::deserialize_compressed(proof_public_inputs_as_bytes)
-        .map_err(|_| FastCryptoError::InvalidInput)?;
+    let mut x = Vec::new();
+    for chunk in proof_public_inputs_as_bytes.chunks(SCALAR_LENGTH) {
```
kchalkias, line 37: Can we document that we get a concatenated bytes, then split the bytes into SCALAR_LENGTH chuncks?

fastcrypto-zkp/src/api.rs
```diff
@@ -31,8 +33,13 @@ pub fn verify_groth16_in_bytes(
...
+        if chunk.len() != SCALAR_LENGTH {
```
kchalkias, line 38: you can check earlier if the `proof_public_inputs_as_bytes.len mod SCALAR_LENGTH` = 0 before parsing anything, to fail fast

fastcrypto-zkp/src/api.rs
```diff
@@ -31,8 +33,13 @@ pub fn verify_groth16_in_bytes(
     proof_public_inputs_as_bytes: &[u8],
     proof_points_as_bytes: &[u8],
 ) -> Result<bool, FastCryptoError> {
-    let x = BlsFr::deserialize_compressed(proof_public_inputs_as_bytes)
-        .map_err(|_| FastCryptoError::InvalidInput)?;
+    let mut x = Vec::new();
```
kchalkias, line 36: We could also return fast if it's an empty public inputs. We won't fail, its possible for a circuit to not have public inputs, we need a test for that too.

kchalkias (approved):
Approving, but we need to follow with upcoming PRs:
a) test against NO public inputs (empty), @benr-ml wdyt?
b) @benr-ml what about the reported memory leak from auditor?

fastcrypto-zkp/src/api.rs
```diff
@@ -44,6 +51,5 @@ pub fn verify_groth16_in_bytes(
         delta_g2_neg_pc_bytes,
     )?;
 
-    verify_with_processed_vk(&blst_pvk, &[x], &proof)
-        .map_err(|_| FastCryptoError::GeneralOpaqueError)
+    verify_with_processed_vk(&blst_pvk, &x, &proof).map_err(|_| FastCryptoError::GeneralOpaqueError)
```
benr-ml, line 54: since there is nothing secret here, we should & can use a more explicit error (e.g., GeneralError(msg))

benr-ml:
> Approving, but we need to follow with upcoming PRs: a) test against NO public inputs (empty), @benr-ml wdyt? b) @benr-ml what about the reported memory leak from auditor?

Yes and yes, both should be done

jonas-lj:
> > Approving, but we need to follow with upcoming PRs: a) test against NO public inputs (empty), @benr-ml wdyt? b) @benr-ml what about the reported memory leak from auditor?
> 
> 
> 
> Yes and yes, both should be done

I've added a new issue.

fastcrypto-zkp/src/api.rs
jonas-lj: That's fixed in #1623

PR #470, opened 2023-03-02
Add Groth16 verifier on BN254

jonas-lj:
This PR adds an api for verifying Groth16 proofs on BN254.

The API differes from corresponding BLS12-381 api in that it takes a `VerifyingKey` and not a `PreparedVerifyingKey` as input. This is done because the `PreparedVerifyingKey` [does not](https://docs.rs/ark-groth16/0.3.0/ark_groth16/data_structures/struct.PreparedVerifyingKey.html) implement `CanonicalSerialize` and `CanonicalDeserialize` in arkworks 0.3.0, which is needed to parse the binary input values from Move, so we need to use the `VerifyingKey` instead. This also makes this api slower and more expensive to call from Move. However, this [is fixed](https://docs.rs/ark-groth16/0.4.0/ark_groth16/data_structures/struct.PreparedVerifyingKey.html) in arkworks 0.4.0, so when #460 is merged, we can change the API.


kchalkias (commented):
Let's update now that https://github.com/MystenLabs/fastcrypto/pull/460 is in

jonas-lj:
> Let's update now that #460 is in

The api is now updated such that the apis for bn254 and bls12831 are completely aligned. Note that the serialization of prepared keys is not similar to arkworks api. We could align this with arkworks, but that would mean that the prepared verification key should be ~ 35.000 bytes long, which is not practical for usage with move. So we stick to the prepared keys as designed in the bls12381 verifier. 

A method for computing the prepared key is exposed in move, but we should probably also create a cli tool to do this.

jonas-lj:
The merge is currently blocked by a [vulnerability](https://rustsec.org/advisories/RUSTSEC-2023-0018) in a dependency to the proptest crate which is not related to this pr.

kchalkias (commented):
A small conflict on verifier.rs and qq re vulnerability, how are we handling that?

kchalkias:
> > Let's update now that #460 is in
> 
> The api is now updated such that the apis for bn254 and bls12831 are completely aligned. Note that the serialization of prepared keys is not similar to arkworks api for neither of them. We could align with arkworks, but that would mean that the prepared verification key will be ~ 35.000 bytes long, which is not practical for usage with move. So I say we stick to the prepared keys as designed in the bls12381 verifier.
> 
> A method for computing the prepared key in the correct form is exposed in move, but we should probably also create a cli tool to do this.

Is this ~ 35.000 bytes long verKey now fixed after updating arkworks?

jonas-lj:
> A small conflict on verifier.rs and qq re vulnerability, is this solved?

I'll rebase. The DoS attack with a large number of public inputs will be fixed in Sui, but in fastcrypto, I think we should allow an arbitrary number of inputs.

fastcrypto-zkp/src/bn254/verifier.rs
```diff
@@ -0,0 +1,141 @@
...
+use fastcrypto::error::FastCryptoError;
+
+#[cfg(test)]
+#[path = "unit_tests/verifier_tests.rs"]
+mod verifier_tests;
+
+/// This is a helper function to store a pre-processed version of the verifying key.
+/// This is roughly homologous to [`ark_groth16::data_structures::PreparedVerifyingKey`].
+/// Note that contrary to Arkworks, we don't store a "prepared" version of the gamma_g2_neg_pc,
+/// delta_g2_neg_pc fields because they are very large and unpractical to use in the binary api.
```
kchalkias, line 21: impractical

kchalkias (approved):
Thanks @jonas-lj amazing

PR #487, opened 2023-03-08
Re-revert batch_verify impl and allow empty batches

jonas-lj:


fastcrypto/src/bls12381/mod.rs
```diff
@@ -682,25 +682,34 @@ impl AggregateAuthenticator for BLS12381AggregateSignature {
...
+             );
+         }
+
+        let result = blst::Signature::verify_multiple_aggregate_signatures(
+            &messages,
+            $dst_string,
+            &agg_pks.iter().map(|m| m.borrow()).collect::<Vec<_>>(),
+            false,
+            &signatures.iter().map(|agg_sig| &agg_sig.sig).collect::<Vec<_>>(),
+            false,
```
benr-ml, line 704: true

benr-ml (approved):

fastcrypto/src/bls12381/mod.rs
```diff
@@ -682,25 +682,34 @@ impl AggregateAuthenticator for BLS12381AggregateSignature {
...
-                        .unwrap()
-                        .map(|x| &x.pubkey)
-                        .collect::<Vec<_>>()[..],
-                );
-            if result != BLST_ERROR::BLST_SUCCESS {
-                return Err(FastCryptoError::GeneralOpaqueError);
-            }
+
+        if signatures.is_empty() {
+            // verify_multiple_aggregate_signatures fails on empty input, but we accept here.
```
kchalkias, line 687: We should be very careful with that, ideally we shouldn't return true, but catch empty sigs at the dapp layer. This has been used as an attack vector in the past.

PR #494, opened 2023-03-12
Use 96 bits for the nonces in batch verify

benr-ml:
Saves about 5% for batches of 8-64 signatures, each with ~80 public keys.

### DIFF
--- a/fastcrypto/src/bls12381/mod.rs
+++ b/fastcrypto/src/bls12381/mod.rs
@@ -273,11 +273,13 @@ impl VerifyingKey for BLS12381PublicKey {
 }
 
 fn get_128bit_scalar<Rng: AllowedRng>(rng: &mut Rng) -> blst_scalar {
-    assert!(BLS_BATCH_RANDOM_SCALAR_LENGTH <= 128);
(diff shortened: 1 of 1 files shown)


fastcrypto/src/bls12381/mod.rs
```diff
@@ -273,7 +273,7 @@ impl VerifyingKey for BLS12381PublicKey {
 }
 
 fn get_128bit_scalar<Rng: AllowedRng>(rng: &mut Rng) -> blst_scalar {
-    assert!(BLS_BATCH_RANDOM_SCALAR_LENGTH == 128);
+    assert!(BLS_BATCH_RANDOM_SCALAR_LENGTH <= 128);
```
kchalkias, line 276: I'd use some const value that we can control in the future

kchalkias (approved):

fastcrypto/src/bls12381/mod.rs
kchalkias: do we even need that assertion? IMO it's obsolete

kchalkias (approved):

fastcrypto/src/bls12381/mod.rs
benr-ml: changed to debug assert

PR #500, opened 2023-03-13
Impl ordering for BytesRepresentation

jonas-lj:
This is needed to be able to use `BytesRepresentation` as keys in BTreeMap in Sui.


kchalkias (approved):

kchalkias (commented):
is this required for the recent publickeybytes indexes etc or for another purpose?

jonas-lj:
> is this required for the recent publickeybytes indexes etc or for another purpose?

It's required for that.

kchalkias (approved):
Then let's merge it

PR #516, opened 2023-03-18
Bump ecdsa and signature crates

jonas-lj:
This PR attempts to bump the version of the ecdsa and signature crates.

Note that once the implementation of ecdsa signature recovery is complete in the signatures crate, eg. that [this PR](https://github.com/RustCrypto/signatures/pull/680) (or something equivalent) is merged, we could use their implementation, but until then we'll have to use our own.

### DIFF
--- a/fastcrypto/src/lib.rs
+++ b/fastcrypto/src/lib.rs
@@ -8,8 +8,6 @@
     rust_2021_compatibility
 )]
 
-pub use signature::Signature as _;
-
 #[cfg(test)]
 #[path = "tests/ed25519_tests.rs"]
 pub mod ed25519_tests;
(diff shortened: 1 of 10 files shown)


kchalkias (commented):
This might soon be required to be merged in Sui as well. Let's rebase and re-review asap.

jonas-lj:
> This might soon be required to be merged in Sui as well. Let's rebase and re-review asap.

Got it. It should be straight-forward to do, but we'll have to keep using our own implementation of pk recovery until the `signature` crate is updated on crates.io.

fastcrypto/src/lib.rs
```diff
@@ -8,8 +8,6 @@
     rust_2021_compatibility
 )]
 
-pub use signature::Signature as _;
```
jonas-lj, line 11: It's safe to remove this once https://github.com/MystenLabs/sui/pull/9914 is merged

kchalkias (approved):

PR #522, opened 2023-03-25
Add benches & unit tests for 14 public input circuit

cryptopapi997:
Title says it all - 
Slight redundancy in the duplicate utils file for parsing the vkey from its string representation in both the unit-tests dir and benches dir. Can either 

1) Leave it like this since it's only used for benches & tests
2) Put the utils file in bn254 directly so both can access (didn't do this initially because assumed you guys don't want that mixed in with the "real" code)
3) Have the vkey in-lined as bytes (didn't do this initially for the sake of better readability)


kchalkias (approved):
Approving this for use in our test cases. The long term plan is to transfer external use cases in a different bench file for better management, but this suffices for now. Thanks!

kchalkias (approved):

PR #523, opened 2023-03-26
Small cleanups of EdDSA

benr-ml:


fastcrypto/src/bls12381/mod.rs
```diff
@@ -272,17 +272,18 @@ impl VerifyingKey for BLS12381PublicKey {
     }
 }
 
-fn get_128bit_scalar<Rng: AllowedRng>(rng: &mut Rng) -> blst_scalar {
-    debug_assert!(BLS_BATCH_RANDOM_SCALAR_LENGTH <= 128);
+fn get_random_scalar<Rng: AllowedRng>(rng: &mut Rng) -> blst_scalar {
+    static_assertions::const_assert!(64 <= BLS_BATCH_RANDOM_SCALAR_LENGTH);
+    static_assertions::const_assert!(BLS_BATCH_RANDOM_SCALAR_LENGTH <= 128);
```
kchalkias, line 277: Why did we prefer this from a single AND clause?

fastcrypto/src/bls12381/mod.rs
```diff
@@ -272,17 +272,18 @@ impl VerifyingKey for BLS12381PublicKey {
...
     let mut vals = [0u64; 4];
     loop {
         vals[0] = rng.next_u64();
         vals[1] = rng.next_u64();
 
-        // Check this ^^
-
         // Reject zero as it is used for multiplication.
-        if vals[0] | vals[1] != 0 {
+        let vals1_lsb = vals[1] & ((1 << (BLS_BATCH_RANDOM_SCALAR_LENGTH - 64)) - 1);
```
kchalkias, line 285: I guess in practice we don't even need to sample `vals[1]` in the first place when requesting for 64bits. That would need two if statements (one for sampling one for checking values are zero), but I'm not sure which is one is better for readability. I'd personally do the two `if`.

fastcrypto/src/ed25519.rs
```diff
@@ -368,6 +448,8 @@ impl PartialEq for Ed25519AggregateSignature {
 
 impl Eq for Ed25519AggregateSignature {}
 
+// Ed25519AggregateSignature is experimental and here we disable it by default.
+#[cfg(any(test, feature = "experimental"))]
```
kchalkias, line 452: thanks

PR #526, opened 2023-03-27
crypto: Expose Bn254 and useful APIs

joyqvq:
useful for sui to use construct public inputs without deserialization. public inputs such as txid, max_epoch should be converted to field elements and pass directly to fastcrypto without needing to ser/de again


kchalkias (approved):

kchalkias (approved):
@joyqvq do we need to merge this and also bump fastcrypto on Sui?

joyqvq:
> @joyqvq do we need to merge this and also bump fastcrypto on Sui?

correct, merging now to unblock in sui

PR #529, opened 2023-03-27
Bump elliptic-curve dependency

jonas-lj:


fastcrypto/src/secp256r1/mod.rs
```diff
@@ -333,17 +330,12 @@ impl Secp256r1KeyPair {
...
-                *FieldBytes::<NistP256>::from_slice(H::digest(msg).as_ref()),
-                &[],
-            )
-            .unwrap()
-            .0;
+
+        let sig: Signature = self
+            .secret
+            .privkey
+            .sign_prehash(H::digest(msg).as_ref())
```
kchalkias, line 337: is this the deterministic rfc6979 signature?

kchalkias (commented):
@jonas-lj please rebase and we can merge

fastcrypto/src/secp256r1/mod.rs
jonas-lj: Yes, it is. I'll add that to the docs to clarify.

PR #542, opened 2023-04-03
Remove redundant hash in ecvrf

jonas-lj:


benr-ml:
1. Let's wait for @kchalkias approval on changing this api
2. the fact that such a meaningful change didn't break any test is a bit scary - let's add regression tests

PR #543, opened 2023-04-04
Code improvements

petrosagg:
This PR is split into multiple commits so that part of it can be merged individually if needed. The last commit which adds the precomputed multiplication needs a bit more work to nicely integrate into the `ristretto` module as in its current form it exposes the internals to the rest o the codebase. This was done for benchmarking purposes.

The performance optimizations yield a speedup of ~27% on the verification benchmark. That gain is primarly attributed to the variable time multiscalar multiplication, which applies to any setting, and secondarily to the caching of the compressed public key which applies to the setting where repeated verifications of proofs with the same public key are performed.

### DIFF
--- a/fastcrypto/Cargo.toml
+++ b/fastcrypto/Cargo.toml
@@ -106,3 +106,7 @@ radix64 = "0.6.2"
 base58 = "0.2.0"
 rust-base58 = "0.0.4"
 bcs.workspace = true
+
+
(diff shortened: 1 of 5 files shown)


kchalkias (commented):
cc: @benr-ml I guess you have better visibility re audit outcomes and what should go in.

PR #544, opened 2023-04-05
ristretto: fix scalar deserialization

jonas-lj:


kchalkias (approved):
did we check backwards compatibility with Sui? Otherwise it seems fine.

jonas-lj:
> did we check backwards compatibility with Sui? Otherwise it seems fine.

Unfortunately, it's not as simple as this. In ecvrf we need to generate ristretto scalars from a pure byte array (as we do now). But we might also need `from_canonical_bytes`. I suggest we have both, use the canonical representation for serde, and fix the naming of the first (from `from_bytes` to `from_bytes_mod_order` or similar).

jonas-lj:
Fixed in #556

PR #546, opened 2023-04-05
Code clean-ups in the Groth16 verifier

jonas-lj:
Misc clean-ups in fastcrypto-zkp. The main contribution is that it introduces wrapper types for the types exposed in public api's for both the Bn254 and Bls12381 verifiers. Note that the binary apis which are used in Sui are _not_ changed by this PR.

### DIFF
--- a/fastcrypto-zkp/src/bn254/api.rs
+++ b/fastcrypto-zkp/src/bn254/api.rs
@@ -1,10 +1,9 @@
 // Copyright (c) 2022, Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::bn254::verifier::{process_vk_special, PreparedVerifyingKey};
+use crate::bn254::verifier::{process_vk_special, verify_with_processed_vk, PreparedVerifyingKey};
 pub use ark_bn254::{Bn254, Fr as Bn254Fr};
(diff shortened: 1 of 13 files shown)


fastcrypto-zkp/src/bn254/api.rs
```diff
@@ -42,40 +44,22 @@ pub fn verify_groth16_in_bytes(
...
-        delta_g2_neg_pc_bytes,
-        &x,
-        proof_points_as_bytes,
-    )
-}
-
-/// Verify Groth16 proof using the serialized form of the prepared verifying key (see more at
-/// [`crate::bn254::verifier::PreparedVerifyingKey`]), a vector of proof public inputs and
-/// serialized proof points.
-pub fn verify_groth16(
```
jonas-lj, line 61: This has been moved to bn254/verifier.rs

kchalkias (commented):
is this compatible with Sui Groth verifiers?

jonas-lj:
> is this compatible with Sui Groth verifiers?

Yes, it doesn't require any changes to the move interface or anywhere else in Sui.

kchalkias (approved):

benr-ml (approved):

PR #548, opened 2023-04-05
crypto: Add Poseidon and expose Base64UrlUnpadded

joyqvq:
experimental feature

### DIFF
--- a/fastcrypto-zkp/Cargo.toml
+++ b/fastcrypto-zkp/Cargo.toml
@@ -24,6 +24,7 @@ ark-serialize = "0.4.1"
 blst = "0.3.10"
 byte-slice-cast = "1.2.2"
 fastcrypto = { path = "../fastcrypto" }
+light-poseidon = "0.0.4"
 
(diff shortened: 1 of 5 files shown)


fastcrypto-zkp/Cargo.toml
```diff
@@ -24,6 +24,7 @@ ark-serialize = "0.4.1"
 blst = "0.3.10"
 byte-slice-cast = "1.2.2"
 fastcrypto = { path = "../fastcrypto" }
+light-poseidon = "0.0.4"
```
kchalkias, line 27: noticed that this is a 4-star repo without forks, do we feel confident about its robustness? any test vectors that we could use (or fuzzers)?

joyqvq: this is the only poseidon + bn254 rust library i can find, agreed we should find a better library and/or add more tests. merged for now on openid-zkp branch

joyqvq:
closing since this is merged to openid-zkp branch

PR #563, opened 2023-04-21
Unit and integration tests for sigs-cli

jonas-lj:


fastcrypto-cli/src/sigs_cli.rs
```diff
@@ -261,3 +261,265 @@ fn execute(cmd: Command) -> Result<(), FastCryptoError> {
...
+
+#[cfg(test)]
+mod tests {
+    use crate::{execute, Command, KeygenArguments, SigningArguments, VerifiyingArguments};
+    use fastcrypto::error::FastCryptoError;
+
+    // Test vectors
+    const MSG: &str = "00010203";
+    const SEED: &str = "0101010101010101010101010101010101010101010101010101010101010101";
+    const ED25519: &str = "ed25519";
```
benr-ml, line 273: let's write those as 
```
struct TestParams {
 name: string,
 private: string,
 public: string,
 sig: string
 }
```

and make the test run on a list of those

kchalkias (commented):
Yes what Ben asked, chatGPT4 is also very helpful for these recommendations, try it and let us know if by providing the code and Ben's suggestion it made it with one shot, I'm curious as an experiment :)

fastcrypto-cli/tests/sigs_cli.rs
```diff
@@ -0,0 +1,216 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use assert_cmd::Command;
+use regex::Regex;
+
+// Test vectors
+const MSG: &str = "00010203";
+const SEED: &str = "0101010101010101010101010101010101010101010101010101010101010101";
+const ED25519: &str = "ed25519";
```
benr-ml, line 10: should we do the same here?

jonas-lj: Yea, should be done now.

PR #565, opened 2023-04-26
Expose secp256k1 Signature

chargarlic:
This is for re-use in sui repo where we use AWS KMS as a signer for secp256k1

### DIFF
--- a/fastcrypto/src/secp256k1/recoverable.rs
+++ b/fastcrypto/src/secp256k1/recoverable.rs
@@ -18,7 +18,7 @@
 //! ```
 
 use crate::hash::HashFunction;
-pub use crate::secp256k1::{DefaultHash, Secp256k1KeyPair, Secp256k1PublicKey, Secp256k1Signature};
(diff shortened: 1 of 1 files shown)


joyqvq:
fix lint with: cargo fmt && cargo xclippy

lgtm

fastcrypto/src/secp256k1/recoverable.rs
```diff
@@ -33,6 +33,7 @@ use rust_secp256k1::{
     All, Message, Secp256k1,
 };
 use std::fmt::{self, Debug, Display};
+pub use rust_secp256k1::ecdsa::Signature as secpSig;
```
kchalkias, line 36: `as secp256k1Sig` not to be confused with `secp256r1` that we also support

kchalkias (changes requested):

fastcrypto/src/secp256k1/recoverable.rs
chargarlic: makes sense, updated it now

kchalkias (approved):

PR #567, opened 2023-04-26
chore: prepare for release and update README

joyqvq:
should we consider enforcing semver? 

```
      # - name: Check semver
      #   # Only run the semver script if the version changed, otherwise it errors out
      #   if: steps.check.outputs.is_new_version == 'yes'
      #   uses: obi1kenobi/cargo-semver-checks-action@v1
      #   with:
      #     crate-name: ${{ matrix.package }}
      #     version-tag-prefix: ${{ matrix.package }}-v
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1505,7 +1505,7 @@ checksum = "4443176a9f2c162692bd3d352d745ef9413eec5782a80d8fd6f8a1ac692a07f7"
 
 [[package]]
 name = "fastcrypto"
-version = "0.1.4"
+version = "0.1.5"
 dependencies = [
  "aes",
  "aes-gcm",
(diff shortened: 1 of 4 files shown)


benr-ml (approved):

benr-ml:
> should we consider enforcing semver?
> 
> ```
>       # - name: Check semver
>       #   # Only run the semver script if the version changed, otherwise it errors out
>       #   if: steps.check.outputs.is_new_version == 'yes'
>       #   uses: obi1kenobi/cargo-semver-checks-action@v1
>       #   with:
>       #     crate-name: ${{ matrix.package }}
>       #     version-tag-prefix: ${{ matrix.package }}-v
> ```

I'd prefer not to since it's more restricting than helping at this stage, but if we do want to enable it, let's run it only for non "experimental" code.

PR #572, opened 2023-04-27
Add BLS-12-381 raw pairing

benr-ml:


kchalkias (commented):
Having discussed w Arnab, we should come up with a stable api for all group operations. we need a design phase for all Ristretto and BLS12-381 (and their related scalar) primitives.

PR #576, opened 2023-05-01
Add tests for r or s zero in p256 ecdsa

jonas-lj:
In the two secp256r1 ECDSA schemes it is not possible to deserialize a signature where either s or r are zero, so the implementation is working but it was a bit unclear in the code how this is handled. This PR adds tests and a few comments to clarify.

### DIFF
--- a/fastcrypto/src/secp256r1/recoverable.rs
+++ b/fastcrypto/src/secp256r1/recoverable.rs
@@ -71,6 +71,8 @@ impl ToFromBytes for Secp256r1RecoverableSignature {
                 SECP256R1_RECOVERABLE_SIGNATURE_LENGTH,
             ));
         }
+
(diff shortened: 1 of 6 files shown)


fastcrypto/src/secp256r1/recoverable.rs
```diff
@@ -206,30 +212,31 @@ impl RecoverableSignature for Secp256r1RecoverableSignature {
...
-            );
-
-            Ok(Secp256r1PublicKey {
-                pubkey: VerifyingKey::from_affine(pk)
-                    .map_err(|_| FastCryptoError::GeneralOpaqueError)?,
-                bytes: OnceCell::new(),
-            })
-        } else {
-            Err(FastCryptoError::GeneralOpaqueError)
+        if big_r.is_none().into() {
```
jonas-lj, line 215: Move the return closer to the check to make the code more readable.

benr-ml (approved):
thanks!

kchalkias (approved):

PR #594, opened 2023-05-23
Add multi scalar mult to BLS + trait

benr-ml:
- Added a trait for groups with multi scalar mul, updated ristretto to use it
- Added msm for BLS' G1 & G2
- cleaned up some tests

### DIFF
--- a/fastcrypto/src/groups/bls12381.rs
+++ b/fastcrypto/src/groups/bls12381.rs
@@ -3,8 +3,10 @@
 
 use crate::bls12381::min_pk::DST_G2;
 use crate::bls12381::min_sig::DST_G1;
-use crate::error::FastCryptoError;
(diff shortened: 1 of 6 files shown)


kchalkias (approved):
LGTM
- @benr-ml do you see any performance optimizations in both BLS and ECVRF? We might need to adjust gas fee in Sui if it's statistically significant.
- @jonas-lj is this a trait that you were planning to use as well (working on r1 and co)?

jonas-lj:
> LGTM
> 
> * @benr-ml do you see any performance optimizations in both BLS and ECVRF? We might need to adjust gas fee in Sui if it's statistically significant.
> * @jonas-lj is this a trait that you were planning to use as well (working on r1 and co)?

I've only implemented it for two multiplications because that's all we need for ECDSA, but it could easily be extended to this trait if we need id.

jonas-lj (approved):

Issue #590, opened 2023-05-12
Design EC apis that can take as inputs precomputed tables of public keys


kchalkias:
This is useful for known/popular public keys. Similar with generator precomputed tables (but for keys)

benr-ml:
@kchalkias where will this improve performance? Sui doesn't check individual signatures in the happy flow

kchalkias:
Not referring to validator BLS, but for popular Sui accounts (ie EdDSA or ECDSA sigs from game studio and custodians/exchanges etc).

benr-ml:
Oh, I see. Let's wait and see if we have such heavy keys in practice.

Issue #589, opened 2023-05-12
Introduce precomputed tables in Ristretto Group generator


kchalkias:
First check if they do exist already. This will improve scalar to point by far.

Issue #573, opened 2023-04-30
Extract secp256r1 from ring /openssl


kchalkias:
In order to avoid importing the whole "heavy" openssl lib, let's extract the secp256r1 part only.

PR #595, opened 2023-05-24
Use pre-computation to speed-up Secp256r1 verification

### LINKED ISSUE #588: Introduce precomputed tables (over generator) in ECDSA r1

jonas-lj:
Fixes #588 and #573 (because the construction is similar to the one uses by openssl)

This PR made to speed up signature verification in Secp256r1 (gives about ~50% speed-up), but it is implemented for generic groups, so it may also be used for other constructions, eg Ristretto (#589) so it's also relevant for #590. It uses a sliding window method (which is variable time) for verification and fixed window (which is constant time) for signing.

The BGMW algorithm is not used for ecdsa, but is faster for single scalar multiplications at the cost of larger precomputation tables. It has been added under the experimental feature flag.


fastcrypto/src/groups/mod.rs
```diff
@@ -37,6 +41,11 @@ pub trait GroupElement:
 
     /// Return an instance of the generator for this group.
     fn generator() -> Self;
+
+    /// Compute 2 * Self. May be overwritten by implementations that has a fast doubling operation.
```
kchalkias, line 45: that ha~s~ve

fastcrypto/src/groups/multiplier/comb_method.rs
```diff
@@ -0,0 +1,140 @@
...
+use crate::serde_helpers::ToFromByteArray;
+
+/// Performs scalar multiplication using a comb method. We must have HEIGHT >= ceil(SCALAR_SIZE * 8 / ceil(log2(WIDTH))
+/// and the precomputation tables will be of size WIDTH x HEIGHT. Once pre-computation has been done,
+/// a scalar multiplication requires HEIGHT additions. Both `mul` and `double_mul` are constant time
+/// assuming the group operations for `G` are constant time.
+///
+/// This method is faster than the fixed window for a single multiplication, but it requires a larger
+/// number of precomputed points.
+pub struct CombMultiplier<
```
kchalkias, line 16: @jonas-lj is there an official name for this algorithm?

fastcrypto/src/groups/multiplier/integer_utils.rs
```diff
@@ -0,0 +1,155 @@
...
+pub(crate) fn div_ceil(numerator: usize, denominator: usize) -> usize {
+    (numerator + denominator - 1) / denominator
+}
+
+/// Get the integer represented by a given range of bits of a an integer represented by a little-endian
+/// byte array from start to end (exclusive). The `end` argument may be arbitrarily large, but if it
+/// is larger than 8*N, the remaining bits of the byte array will be assumed to be zero.
+#[inline]
+pub fn get_bits_from_bytes<const N: usize>(bytes: &[u8; N], start: usize, end: usize) -> usize {
+    assert!(start < end && start < 8 * N);
```
kchalkias, line 46: Is this something that in Sui can be called arbitrarily by a malicious sender? or we should return an error instead of panicking?

fastcrypto/src/groups/secp256r1.rs
```diff
@@ -0,0 +1,121 @@
...
+}
+
+impl ToFromByteArray<SCALAR_SIZE_IN_BYTES> for Scalar {
+    fn from_byte_array(bytes: &[u8; SCALAR_SIZE_IN_BYTES]) -> Result<Self, FastCryptoError> {
+        Ok(Scalar(Fr::from_le_bytes_mod_order(bytes)))
+    }
+
+    fn to_byte_array(&self) -> [u8; SCALAR_SIZE_IN_BYTES] {
+        let mut bytes = [0u8; SCALAR_SIZE_IN_BYTES];
+        self.0.serialize_uncompressed(&mut bytes[..]).unwrap();
```
kchalkias, line 118: ditto re panicking, can a malicious sender invoke this with a wrong scalar bytes in Sui?

fastcrypto/src/groups/multiplier/integer_utils.rs
jonas-lj: All values are determined by constants (window_width and scalar_size) which is set in the secp256r1 file, so a caller has no influence on them.

fastcrypto/src/groups/secp256r1.rs
jonas-lj: This will only fail if the array isn't large enough, and the internal type is an arkworks scalar, so this check already happened in deserialization.

fastcrypto/src/groups/multiplier/bgmw.rs
```diff
@@ -0,0 +1,145 @@
...
+    use super::*;
+    use crate::groups::ristretto255::{RistrettoPoint, RistrettoScalar};
+    use crate::groups::secp256r1::{ProjectivePoint, Scalar};
+
+    #[test]
+    fn test_scalar_multiplication_ristretto() {
+        let multiplier = BGMWScalarMultiplier::<RistrettoPoint, RistrettoScalar, 16, 64, 32>::new(
+            RistrettoPoint::generator(),
+        );
+        let scalar = RistrettoScalar::from(12345423);
```
kchalkias, line 108: Shall we check against edge cases? ie scalar = 0 or 1 or close to the order?

kchalkias (approved):
Amazing work!

PR #597, opened 2023-05-26
zk-login: Add APIs and structs

joyqvq:
api.rs and zk_login.rs is ready for review. (committing an example test files too.)


fastcrypto-zkp/src/bn254/zk_login.rs
```diff
@@ -0,0 +1,753 @@
...
+        vec![SupportedKeyClaim::Sub, SupportedKeyClaim::Email];
+
+    supported_claims
+        .iter()
+        .map(|claim| claim.to_string())
+        .collect()
+}
+
+/// Verify proof using fixed verifying key.
+pub fn verify_groth16_with_fixed_vk(
```
kchalkias, line 134: Do we plan to invoke this from Sui? I'm currently ok to have special functions and fixed configs for the time being here, maybe in the future we should put them in a different "gadgets" folder in fastcrypto or even in Sui. Good to proceed as it is atm.

fastcrypto-zkp/src/bn254/zk_login.rs
```diff
@@ -0,0 +1,753 @@
...
+    }
+
+    /// Get the key claim name string.
+    pub fn get_claim_name(&self) -> &str {
+        &self.key_claim_name
+    }
+
+    /// Calculate the poseidon hash from 10 selected fields in the aux inputs.
+    pub fn calculate_all_inputs_hash(&self) -> String {
+        // Safe to unwrap here all fields are converted to string from valid BigInt.
```
kchalkias, line 344: <3

kchalkias (approved):

PR #599, opened 2023-05-30
zk-login: Add APIs and structs

joyqvq:


kchalkias (commented):
@joyqvq do we know why the test fails with json not found?

joyqvq:
> @joyqvq do we know why the test fails with json not found?

updated to use read from raw string instead of file path

.gitignore
```diff
@@ -6,4 +6,7 @@ fastcrypto/Cargo.lock
 fastcrypto-derive/Cargo.lock
 
 # .DS_Store files
-.DS_Store
\ No newline at end of file
+.DS_Store
+
+# JS
```
mskd12, line 11: No JS files in the lib

fastcrypto-zkp/src/bn254/zk_login.rs
```diff
@@ -0,0 +1,749 @@
...
+
+/// Calculate the merklized hash of the given bytes after 0 paddings.
+pub fn calculate_merklized_hash(bytes: &[u8]) -> String {
+    let mut bitarray = bytearray_to_bits(bytes);
+    pad_bitarray(&mut bitarray, 248);
+    let bigints = convert_to_bigints(&bitarray, 248);
+    let mut poseidon1 = PoseidonWrapper::new(15);
+    let hash1 = poseidon1.hash(&bigints[0..15]);
+
+    let mut poseidon2 = PoseidonWrapper::new(bigints.len() - 15);
```
mskd12, line 703: What happens if bigints.len() < 15? 

Minor: You are combining two different operations (base change and poseidon hash) into one function (which is fine but perhaps warrants different function?).

fastcrypto-zkp/src/bn254/poseidon.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+#[cfg(test)]
+mod test {
+    use crate::bn254::zk_login::calculate_merklized_hash;
+
+    use super::PoseidonWrapper;
+    use ark_bn254::Fr;
+    use std::str::FromStr;
+    #[test]
+    fn poseidon_test() {
```
mskd12, line 41: I think it might be a good idea to be more robust with our Poseidon tests since we are using new libraries. 

We could generate 15 test vectors, one for each vector length, i.e., 1, 2, 3, ..., 15, and use these to make sure that all the libraries (Rust's light_poseidon, circom's internal one and JS's poseidon-light) conform with each other.

fastcrypto-zkp/src/bn254/zk_login.rs
```diff
@@ -0,0 +1,749 @@
...
+impl fmt::Display for SupportedKeyClaim {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            SupportedKeyClaim::Email => write!(f, "email"),
+            SupportedKeyClaim::Sub => write!(f, "sub"),
+        }
+    }
+}
+
+pub fn get_supported_claims() -> Vec<String> {
```
mskd12, line 122: Perhaps remove email until we decide to enable it?

fastcrypto-zkp/src/circom.rs
mskd12, line 1: Are the functions in this file being used somewhere?

fastcrypto-zkp/src/circom.rs
```diff
@@ -0,0 +1,117 @@
...
+    nPublic: u8,
+    vk_alpha_1: CircomG1,
+    vk_beta_2: CircomG2,
+    vk_gamma_2: CircomG2,
+    vk_delta_2: CircomG2,
+    IC: Vec<CircomICElement>,
+}
+
+// Read a Circom vkey file and return the arkworks verification key
+pub fn read_vkey(path: &str) -> VerifyingKey<Bn254> {
```
mskd12, line 51: Is `read_vkey` used somewhere?

fastcrypto-zkp/src/bn254/zk_login.rs
```diff
@@ -0,0 +1,749 @@
...
+        if jwt_length != 8 * (payload_start_index + payload_len) {
+            return Err(FastCryptoError::GeneralError(
+                "Incorrect jwt length".to_string(),
+            ));
+        }
+
+        // Parse sha2 pad into a bit array.
+        let sha_2_pad = bytearray_to_bits(&masked_content_tmp[payload_start_index + payload_len..]);
+
+        // Verify that the first bit of the bit array of sha2 pad is 1.
```
mskd12, line 449: Move code from lines 449-458 into a function titled `validate_sha2_pad`? The logic in `validate_zeros_count` can be a part of this.

fastcrypto-zkp/src/bn254/zk_login.rs
```diff
@@ -0,0 +1,749 @@
...
+    BigInt::from_str(value)
+        .expect("Invalid big int string")
+        .to_bytes_be()
+        .1
+}
+
+/// Calculate the integer value from the bytearray.
+fn calculate_value_from_bytearray(arr: &[u8]) -> usize {
+    let sized: [u8; 8] = arr.try_into().expect("Invalid byte array");
+    ((sized[7] as u16) | (sized[6] as u16) << 8).into()
```
mskd12, line 674: Is this assuming that the length will fit into the last two bytes? That seems fine, but we should at least check that the remaining 6 bytes are zeroes. Also, if you're only using this function at one place: why not take a fixed-length array of size 8 as input?

fastcrypto-zkp/src/bn254/zk_login.rs
```diff
@@ -0,0 +1,749 @@
...
+fn convert_to_bigints(bitarray: &[bool], segment_size: usize) -> Vec<Bn254Fr> {
+    let chunks = bitarray.chunks(segment_size);
+    chunks
+        .map(|chunk| {
+            let mut bytes = vec![0; (segment_size + 7) / 8];
+            for (i, &bit) in chunk.iter().enumerate() {
+                bytes[i / 8] |= (bit as u8) << (7 - i % 8);
+            }
+            let f = bitarray_to_string(chunk);
+            let st = BigInt::from_radix_be(Sign::Plus, &f, 2)
```
mskd12, line 743: Nit: I suspect `BigUint::from_radix_be(&f, 2)` should work

fastcrypto-zkp/src/bn254/poseidon.rs
joyqvq: adding todo for now

fastcrypto-zkp/src/bn254/zk_login.rs
joyqvq: removed

mskd12 (approved):
Looks mostly good!

PR #607, opened 2023-06-15
Add test for recoverable signatures with id 2 or 3

shresthagrawal:
Signatures with ids 2 or 3 only occur when their $r$ value is small enough that it is feasible for either $r$ or $r+q$ to be the x-coordinate of an intermediate curve point, where $q$ is the curve order. Due to the relation between the curve order $q$ and field modulus $p$, such cases can only occur rarely during normal usage as the fraction $p-q \over p$ is negligible. However, it's simple and tractable to construct them on purpose. For this reason, it is important that they be treated consistently by both branches of the implementation (i.e., converting between recoverable and non-recoverable forms of the same signature whilst keeping track of the related public key should never change the output of the verification function). We have added testing to confirm this is indeed the case.


benr-ml (approved):
Thanks for this!

@jonas-lj fyi

benr-ml:
Thanks! Closed in https://github.com/MystenLabs/fastcrypto/pull/617

PR #612, opened 2023-06-21
Merkle trees first commit

jonas-lj:


fastcrypto-data/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+authors = ["Mysten Labs <build@mystenlabs.com>"]
+readme = "../README.md"
+description = "Collection of useful cryptographic data structures"
+repository = "https://github.com/MystenLabs/fastcrypto"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+rs_merkle = "1.4.1"
+fastcrypto = "0.1.5"
```
benr-ml, line 15: let's use fastcrypto = { path = "../fastcrypto" }

fastcrypto-data/src/merkle_tree.rs
```diff
@@ -0,0 +1,298 @@
...
+//! tree.insert_all(elements.iter());
+//!
+//! let index = 1;
+//! let proof = tree.prove(index);
+//!
+//! let verifier = tree.create_verifier().unwrap();
+//! assert!(verifier.verify(index, &elements[index], &proof));
+//! ```
+
+use std::borrow::Borrow;
```
benr-ml, line 25: What is the use case for this data structure? there are many variants of merkle trees, e.g., key-value, with deletions, etc.
Also, what is the expected number of objects, and the expected insertion/deletion pattern?

kchalkias (commented):
Let's resurrect that PR and get some timings as well (part of our light client story)

PR #615, opened 2023-06-29
fix: update verifier logic

joyqvq:
reflect all verifier changes based on circuit updates.


fastcrypto-zkp/src/bn254/zk_login_api.rs
```diff
@@ -0,0 +1,202 @@
...
+        beta_g2: vk_beta_2,
+        gamma_g2: vk_gamma_2,
+        delta_g2: vk_delta_2,
+        gamma_abc_g1: vk_gamma_abc_g1,
+    };
+    process_vk_special(&Bn254VerifyingKey(vk))
+}
+
+/// A whitelist of client_ids (i.e. the value of "aud" in claims) for each provider.
+pub static DEFAULT_WHITELIST: Lazy<HashMap<&str, Vec<&str>>> = Lazy::new(|| {
```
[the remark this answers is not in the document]
joyqvq: yeah forgot to remove this

fastcrypto-zkp/src/bn254/zk_login.rs
```diff
@@ -1,84 +1,140 @@
...
+
+impl OAuthProviderContent {
+    /// Get the kid string.
+    pub fn kid(&self) -> &str {
+        &self.kid
+    }
+
+    /// Parse OAuthProviderContent from the reader struct.
+    pub fn from_reader(reader: OAuthProviderContentReader) -> FastCryptoResult<Self> {
+        if reader.alg != "RS256" || reader.my_use != "sig" || reader.kty != "RSA" {
```
mskd12, line 86: I'd like us to be careful with this check. After George's exploration of JWTs is done, let's see if all the headers actually contain these fields..

fastcrypto-zkp/src/bn254/zk_login.rs
```diff
@@ -123,380 +170,227 @@ impl fmt::Display for SupportedKeyClaim {
...
-        )?;
+        let mut poseidon = PoseidonWrapper::new();
+        let addr_seed = to_field(&self.address_seed)?;
 
-        // Parse the jwt length from the last 8 bytes of the masked content.
-        let jwt_length_bytes = masked_content_tmp
-            .get(masked_content_tmp.len() - 8..)
-            .ok_or_else(|| FastCryptoError::GeneralError("Invalid last 8 bytes".to_string()))?;
-        let jwt_length = calculate_value_from_bytearray(jwt_length_bytes);
+        let (first_half, second_half) = eph_pubkey_bytes.split_at(eph_pubkey_bytes.len() / 2);
```
mskd12, line 333: There is a mismatch between the logic in TS code and this. In the TS code, the splitting is deterministic, i.e., we compute `eph_pubkey_bytes / 2^128` and `eph_pubkey_bytes % 2^128`. If the public key is bigger than 256 bits, then I think we have a bug at hand. 

Could you update this to always split at 128 bits, i.e., 16 bytes?

joyqvq: revised - it should be 17 since the index here is exclusive!

fastcrypto-zkp/src/bn254/zk_login.rs
joyqvq: happy to remove when new providers appear to not have this. for now, the three providers we want to launch with all satisfy this.

fastcrypto-zkp/src/bn254/zk_login.rs
```diff
@@ -1,86 +1,144 @@
...
+            kty: reader.kty,
+            kid: reader.kid,
+            e: trim(reader.e),
+            n: trim(reader.n),
+            alg: reader.alg,
+        })
+    }
+
+    /// Parse OAuthProviderContent from the reader struct.
+    pub fn validate(&self) -> FastCryptoResult<()> {
```
mskd12, line 99: Where is this used? It seems like some validation is being done in `from_reader` and some here. Can we consolidate so that validation only happens at one place?

fastcrypto-zkp/src/bn254/zk_login.rs
```diff
@@ -1,86 +1,144 @@
...
+        if self.alg != "RS256" || self.kty != "RSA" || self.e != "AQAB" {
+            return Err(FastCryptoError::InvalidInput);
+        }
+        Ok(())
+    }
+}
+
+/// Trim trailing '=' so that it is considered a valid base64 url encoding string by base64ct library.
+fn trim(str: String) -> String {
+    str.trim_end_matches(|c: char| c == '=').to_owned()
```
jonas-lj, line 109: I think `trim_end_matches('=')` should work here.

fastcrypto-zkp/src/bn254/zk_login.rs
```diff
@@ -626,21 +471,21 @@ fn base64_to_bitarray(input: &str) -> Vec<bool> {
...
+                bits.push(u8::from((index >> (5 - i)) & 1 == 1));
             }
             bits
         })
         .collect()
 }
 
-/// Convert a bitarray to a bytearray.
-fn bits_to_bytes(bits: &[bool]) -> Vec<u8> {
+/// Convert a bitarray to a bytearray by taking each 8 bits as a byte.
```
jonas-lj, line 481: Does each u8 represent a bit? That's not clear from the doc.

jonas-lj (changes requested):

fastcrypto-zkp/src/bn254/zk_login.rs
```diff
@@ -1,86 +1,144 @@
...
+const MAX_HEADER_LEN: u16 = 500;
+const PACK_WIDTH: u16 = 248;
+const ISS: &str = "iss";
+const AUD: &str = "aud";
+const NUM_EXTRACTABLE_STRINGS: u8 = 5;
+const MAX_EXTRACTABLE_STR_LEN: u16 = 150;
+const MAX_EXTRACTABLE_STR_LEN_B64: u16 = 4 * (1 + MAX_EXTRACTABLE_STR_LEN / 3);
 
 /// Supported OAuth providers. Must contain "openid" in "scopes_supported"
 /// and "public" for "subject_types_supported" instead of "pairwise".
```
mskd12, line 36: The statement about "public" vs "pairwise" is not true anymore. We support both.

fastcrypto-zkp/src/bn254/zk_login.rs
```diff
@@ -658,99 +503,154 @@ fn bits_to_bytes(bits: &[bool]) -> Vec<u8> {
...
-    bytearray
-        .iter()
-        .flat_map(|&byte| (0..8).rev().map(move |i| (byte >> i) & 1 == 1))
-        .collect()
+
+/// Maps a stream of bigints to a single field element. First we convert the base from
+/// inWidth to packWidth. Then we compute the poseidon hash of the "packed" input.
+/// input is the input vector containing equal-width big ints. inWidth is the width of
+/// each input element.
+fn map_to_field(input: &[BigUint], in_width: u16) -> Result<Bn254Fr, FastCryptoError> {
```
mskd12, line 563: We've decided to rename `map_to_field` to `hash_to_field`..

fastcrypto-zkp/src/bn254/zk_login.rs
```diff
@@ -658,99 +503,154 @@ fn bits_to_bytes(bits: &[bool]) -> Vec<u8> {
...
-        let mut poseidon1: PoseidonWrapper = PoseidonWrapper::new();
-        Ok(poseidon1.hash(inputs)?.to_string())
-    } else if inputs.len() <= 30 {
-        let mut poseidon1: PoseidonWrapper = PoseidonWrapper::new();
-        let hash1 = poseidon1.hash(inputs[0..15].to_vec())?;
-
-        let mut poseidon2 = PoseidonWrapper::new();
-        let hash2 = poseidon2.hash(inputs[15..].to_vec())?;
+/// Pads a stream of bytes and maps it to a field element
+fn map_bytes_to_field(str: &str, max_size: u16) -> Result<Bn254Fr, FastCryptoError> {
```
mskd12, line 534: We've decided to rename map_bytes_to_field to hash_bytes_to_field..

fastcrypto-zkp/src/bn254/zk_login.rs
```diff
@@ -1,86 +1,144 @@
...
+    /// See https://www.facebook.com/.well-known/openid-configuration/
+    Facebook,
+}
+
+/// Struct that contains all the OAuth provider information. A list of them can
+/// be retrieved from the JWK endpoint (e.g. <https://www.googleapis.com/oauth2/v3/certs>)
+/// and published on the bulletin along with a trusted party's signature.
+// #[derive(Debug, Clone, PartialEq, Eq, JsonSchema, Hash, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct OAuthProviderContent {
```
mskd12, line 52: Rename `OAuthProviderContent` to `JWK`?

fastcrypto-zkp/src/bn254/zk_login.rs
```diff
@@ -1,86 +1,144 @@
...
-use num_bigint::{BigInt, BigUint};
-use once_cell::sync::Lazy;
-use regex::Regex;
+use num_bigint::{BigInt, BigUint, Sign};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
-use serde_json::Value;
 use std::str::FromStr;
 
+type ParsedJWKs = Vec<((String, String), OAuthProviderContent)>;
```
mskd12, line 22: Seems like there is some redundancy here.. E.g., `kid` is both inside `OAuthProviderContent` and in the first tuple. Can this be avoided?

fastcrypto-zkp/src/bn254/zk_login.rs
joyqvq: whats "hash" here? the list of big ints?

fastcrypto-zkp/src/bn254/zk_login.rs
joyqvq: removed

fastcrypto-zkp/src/bn254/utils.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+/// Calculate the nonce for the given parameters. Nonce is defined as the Base64Url encoded of the poseidon hash of 4 inputs:
+/// first half of eph_pk_bytes in BigInt, second half of eph_pk_bytes in BigInt, max_epoch and jwt_randomness.
+pub fn get_nonce(eph_pk_bytes: &[u8], max_epoch: u64, jwt_randomness: &str) -> String {
+    let (first_half, second_half) = eph_pk_bytes.split_at(17);
+    let first_bigint = BigInt::from_bytes_be(Sign::Plus, first_half);
+    let second_bigint = BigInt::from_bytes_be(Sign::Plus, second_half);
+    let mut poseidon = PoseidonWrapper::new();
+    let first = Bn254Fr::from_str(&first_bigint.to_string()).unwrap();
```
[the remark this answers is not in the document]
joyqvq: turns out there is a conversion from biguint to Fr. revised.

fastcrypto-zkp/src/bn254/utils.rs
```diff
@@ -0,0 +1,85 @@
...
+    Ok(Base64UrlUnpadded::encode(truncated, &mut buf)
+        .unwrap()
+        .to_string())
+}
+
+/// Given a 33-byte public key bytes (flag || pk_bytes), returns the two Bn254Fr split at the 128 bit index.
+pub fn split_to_two_frs(eph_pk_bytes: &[u8]) -> Result<(Bn254Fr, Bn254Fr), FastCryptoError> {
+    // Split the bytes deterministically such that the first element contains the first 128
+    // bits of the hash, and the second element contains the latter ones.
+    let (first_half, second_half) = eph_pk_bytes.split_at(17);
```
mskd12, line 71: While this works, I recommend aligning with the logic elsewhere (in case we move from 33 to 34 in the future). Could we instead split at `(len - 16)` so that the first and second chunks are always guaranteed to be `eph_pk / 2^128` and `eph_pk % 2^128` respectively

fastcrypto-zkp/src/bn254/zk_login_api.rs
```diff
@@ -0,0 +1,164 @@
...
+fn verify_zk_login_proof_with_fixed_vk(
+    usage: Environment,
+    proof: Proof<Bn254>,
+    public_inputs: &[Bn254Fr],
+) -> Result<bool, FastCryptoError> {
+    let pvk = match usage {
+        Environment::Production => &GLOBAL_VERIFYING_KEY,
+        Environment::Test => &INSECURE_GLOBAL_VERIFYING_KEY,
+    };
+    Groth16::<Bn254>::verify_with_processed_vk(pvk, public_inputs, &proof)
```
jonas-lj, line 162: You can avoid the map_err because the verify function also returns a FastCryptoError.

jonas-lj (approved):

fastcrypto-zkp/src/bn254/zk_login_api.rs
joyqvq: this returns a arkworks error enum so needs to be mapped

PR #618, opened 2023-07-24
Misc fixes of secp256r1

jonas-lj:
Some fixes of the Secp256r1 implementation, including:
* Ensure all parameters for the multiplier works correctly.
* Follow RFC 6979 correctly: The old implementation computed nonce as HMAC(sk, H(m)) but it should be HMAC(sk, H(m) mod q). This is not a security issue since the nonces are not correlated and no nonces will be reused.
* Ensure that sampling of random scalars are done uniformly.
* Implement code path for recovery id's 2 and 3.


fastcrypto/src/secp256r1/mod.rs
```diff
@@ -299,14 +299,16 @@ impl AsRef<[u8]> for Secp256r1PrivateKey {
 
 impl zeroize::Zeroize for Secp256r1PrivateKey {
     fn zeroize(&mut self) {
+        // Unwrap is safe here because we are using a constant and it has been tested
+        // (see fastcrypto/src/tests/secp256r1_tests::test_sk_zeroization_on_drop)
+        self.privkey = ExternalSecretKey::from_bytes(&Scalar::ONE.to_bytes()).unwrap();
```
benr-ml, line 304: why is this needed? (was the old comment wrong?)

jonas-lj: No, I don't think it was. I just added this line to align the implementation with secp256k1.

benr-ml (approved):

PR #625, opened 2023-08-14
DKG - Support reducing weights given an allowed gap in the threshold

benr-ml:


fastcrypto-tbls/src/nodes.rs
```diff
@@ -76,4 +76,35 @@ impl<G: GroupElement> Nodes<G> {
...
+    /// - The new threshold is ceil(t / d)
+    /// - The new weights are all divided by d (floor division)
+    /// - The precision loss, counted as the sum of the remainders of the division by d, is at most
+    ///   the allowed delta
+    /// In practice, allowed delta will be the extra liveness we would assume above 2f+1.
+    pub fn reduce(&self, t: u16, allowed_delta: u16) -> (Self, u16) {
+        let mut max_d = 1;
+        for d in 2..=50 {
+            let sum = self.nodes.iter().map(|n| n.weight % d).sum::<u16>();
+            if sum <= allowed_delta {
```
jonas-lj, line 91: Maybe I'm misunderstanding this, but even after sum > allowed_delta for some d, there could still be a larger d which makes the sum <= allowed_delta, so I don't think you can quit the loop early here.

fastcrypto-tbls/src/nodes.rs
```diff
@@ -74,4 +74,33 @@ impl<G: GroupElement> Nodes<G> {
...
+    /// In practice, allowed delta will be the extra liveness we would assume above 2f+1.
+    pub fn reduce(&self, t: u16, allowed_delta: u16) -> (Self, u16) {
+        let mut max_d = 1;
+        for d in 2..=40 {
+            let sum = self.nodes.iter().map(|n| n.weight % d).sum::<u16>();
+            if sum <= allowed_delta {
+                max_d = d;
+            }
+        }
+        let nodes = self
```
jonas-lj, line 93: I don't know if you win a lot by doing so, but you could exit early here if max_d == 1.

jonas-lj (approved):

PR #631, opened 2023-08-16
fix: move supported providers outside fastcrypto

joyqvq:
the supported providers list should be maintained by sui, since any updates there changes the transaction validation hence needs to go through protocol upgrades properly, otherwise it will cause a consensus fork

also migrate to use maintained version of poseidon-ark relicensed to apache

also updated tests to include facebook

be explicit of the vk in test/prod are different

### DIFF
--- a/fastcrypto-zkp/src/bn254/zk_login_api.rs
+++ b/fastcrypto-zkp/src/bn254/zk_login_api.rs
@@ -18,7 +18,7 @@ use fastcrypto::error::FastCryptoError;
 use once_cell::sync::Lazy;
 
 /// Enum to specify the environment to use for verifying keys.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub enum ZkLoginEnv {
     /// Use the secure global verifying key derived from ceremony.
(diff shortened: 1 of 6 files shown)


mskd12 (approved):

fastcrypto-zkp/src/bn254/zk_login_api.rs
```diff
@@ -32,10 +32,99 @@ impl Default for ZkLoginEnv {
...
+            "1".to_string(),
+        ],
+        vec![
+            "6267760579143073538587735682191258967573139158461221609828687320377758856284"
+                .to_string(),
+            "18672820669757254021555424652581702101071897282778751499312181111578447239911"
+                .to_string(),
+            "1".to_string(),
+        ],
+    ] {
```
joyqvq, line 111: this is an old (so incorrect) vk here - i want to use explicitly as a different one (to be replaced after the ceremony) so we don't accidentally release the feature in prod cc @mskd12

jonas-lj (approved):

PR #635, opened 2023-08-18
move get_proof get_salt and style fixes

joyqvq:


fastcrypto-zkp/src/bn254/utils.rs
```diff
@@ -64,6 +71,66 @@ pub fn get_nonce(
...
 
+/// A response struct for the salt server.
+#[derive(Deserialize, Debug)]
+pub struct GetSaltResponse {
+    /// The salt in BigInt string.
+    salt: String,
+}
+
+/// Call the salt server for the given jwt_token and return the salt.
+pub async fn get_salt(jwt_token: &str) -> Result<String, FastCryptoError> {
```
joyqvq, line 82: moving them from sui to here for better organization. this utils.rs will become "enoki-sdk" in the future

mskd12 (approved):

fastcrypto-zkp/src/bn254/utils.rs
mskd12: Any reason to keep the SDK in Rust? If not, perhaps we should keep a TS equivalent of this file in https://github.com/MystenLabs/sui/tree/main/sdk/zklogin?

joyqvq: because im already using these functions in rust, i found it useful just to extract them to get a rust sdk for free. the ts sdk can be developed as well and at some point they should be in parity, now its too immature to do that + i did not know it exists till now

PR #641, opened 2023-08-28
Allow larger discriminants and strong fiat-shamir

jonas-lj:
Allow discriminants of arbitrary sizes while maintaining compatibility with chiavdf if the size is 1024 bits. Also add the possibility to use a string Fiat-Shamir construction in the VDF implementation.

### DIFF
--- a/fastcrypto/src/vdf/wesolowski.rs
+++ b/fastcrypto/src/vdf/wesolowski.rs
@@ -155,7 +155,7 @@ impl<const CHALLENGE_SIZE: usize> FiatShamir<QuadraticForm> for StrongFiatShamir
         let mut seed = vec![];
         seed.extend_from_slice(&input.as_bytes());
         seed.extend_from_slice(&output.as_bytes());
-        seed.extend_from_slice(&BigInt::from(&vdf.group_parameter).to_bytes_be().1);
+        seed.extend_from_slice(&vdf.group_parameter.to_bytes());
         seed.extend_from_slice(&vdf.iterations.to_be_bytes());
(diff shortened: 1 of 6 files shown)


fastcrypto/src/vdf/wesolowski.rs
```diff
@@ -107,13 +114,53 @@ impl WesolowskiVDF<QuadraticForm> {
...
+pub struct StrongFiatShamir<const CHALLENGE_SIZE: usize>;
+
+impl<const CHALLENGE_SIZE: usize> FiatShamir<QuadraticForm> for StrongFiatShamir<CHALLENGE_SIZE> {
+    fn compute_challenge<F>(
+        vdf: &WesolowskiVDF<QuadraticForm, F>,
+        input: &QuadraticForm,
+        output: &QuadraticForm,
+    ) -> BigInt {
+        let mut seed = vec![];
+        seed.extend_from_slice(&input.as_bytes());
```
benr-ml, line 157: let's make sure that the input to the hash has 1:1 encoding, either by prefixing vectors with their length (fixed u32) or by fixed size inputs (per discr.)

jonas-lj: Actually, the length is fixed size per discriminant size. The input and output has length as computed in `serialized_length` which depends only (and is non-decreasing) on the discriminant length, and the number of iterations is [always 8 bytes](https://doc.rust-lang.org/std/primitive.u64.html#method.to_be_bytes).

benr-ml: sg, let's document this in the function

benr-ml (approved):

PR #644, opened 2023-08-29
tBLS - batch sign and verify

benr-ml:


fastcrypto-tbls/src/dl_verification.rs
```diff
@@ -3,47 +3,53 @@
...
 /// Helper functions for checking relations between scalars and group elements.
 
+fn dot<S: Scalar>(v1: &[S], v2: &[S]) -> S {
+    assert_eq!(v1.len(), v2.len());
+    v1.iter()
+        .zip(v2.iter())
+        .fold(S::zero(), |acc, (a, b)| acc + *a * *b)
+}
+
+pub fn batch_coefficients<S: Scalar>(r: &[S], coeff: &[S], degree: u32) -> Vec<S> {
```
jonas-lj, line 18: I think we should add a method doc to this public function.

jonas-lj (approved):

PR #645, opened 2023-08-30
ZK - Small fixes

benr-ml:


fastcrypto-zkp/src/bn254/utils.rs
```diff
@@ -137,6 +144,7 @@ pub fn split_to_two_frs(eph_pk_bytes: &[u8]) -> Result<(Bn254Fr, Bn254Fr), FastC
     // bits of the hash, and the second element contains the latter ones.
     let (first_half, second_half) = eph_pk_bytes.split_at(eph_pk_bytes.len() - 16);
     let first_bigint = BigUint::from_bytes_be(first_half);
+    // TODO: this is not safe if the buffer is large. Can we use a fixed size array for eph_pk_bytes?
     let second_bigint = BigUint::from_bytes_be(second_half);
```
joyqvq, line 148: not sure if we can fix the size bc eph_pubkey here can be 33 or 34 bytes depends on the scheme 🤔

joyqvq (approved):

fastcrypto-zkp/src/bn254/utils.rs
benr-ml: ok, so maybe we should just verify that eph_pubkey.len()<=34?

PR #647, opened 2023-09-06
VDF with rug/gmp

jonas-lj:
This PR adds a new implementation of class groups to be used with VDF's. The new implementation uses the rug crate which is a wrapper around the GMP C++ library, so it'll require the GMP library to be installed.

The new implementation is a lot faster (10x) or so, and is enabled by setting the `gmp` feature flag. A doubling in a group with 1024 bit discriminant is around 9 μs per iteration or 110k iterations per second compared to ~160k iterations with chiavdf on my laptop.

It is possible to install GMP on windows, but it's a bit of a hassle (requires MinGW) so the new feature is only tested in CI on ubuntu.


benr-ml (approved):
Let's make the vdf module under feature `experimental`

PR #649, opened 2023-09-08
Add poseidon/BN254 API

jonas-lj:


fastcrypto-zkp/src/bn254/poseidon.rs
```diff
@@ -60,17 +68,69 @@ pub fn to_poseidon_hash(inputs: Vec<Fr>) -> Result<Fr, FastCryptoError> {
     }
 }
 
+/// Given a binary representation of a BN254 field element as an integer in little-endian encoding,
+/// this function returns the corresponding field element. If the field element is not canonical (is
+/// larger than the field size as an integer), an `FastCryptoError::InvalidInput` is returned.
+///
+/// If more than 32 bytes is given, an `FastCryptoError::InputTooLong` is returned.
+fn from_canonical_le_bytes_to_field_element(bytes: &[u8]) -> Result<Fr, FastCryptoError> {
```
joyqvq, line 76: whats the rationale we use little endian? i think in fc we usually use BE

jonas-lj: It's because the u256 type in the sui framework only have a `to_le_bytes` [method](https://github.com/MystenLabs/sui/blob/92bd33895502c52f7933c15c6575d6aeb6f83efb/external-crates/move/move-core/types/src/u256.rs#L307-L309), so using LE makes this implementation slightly easier, but I don't have a preference for either.

joyqvq (approved):

PR #650, opened 2023-09-14
Add verify_zk_login_id function

jonas-lj:


fastcrypto-zkp/src/bn254/unit_tests/zk_login_tests.rs
```diff
@@ -539,3 +544,25 @@ async fn test_end_to_end_twitch() {
     );
     assert!(res.is_ok());
 }
+
+#[test]
+fn test_verify_zk_login() {
```
benr-ml, line 549: test also failures

fastcrypto-zkp/src/bn254/zk_login_api.rs
```diff
@@ -169,3 +170,27 @@ fn verify_zk_login_proof_with_fixed_vk(
...
+    aud: &str,
+    iss: &str,
+    salt_hash: &str,
+) -> FastCryptoResult<bool> {
+    let address_seed = gen_address_seed_with_salt_hash(salt_hash, name, value, aud)?;
+    verify_zk_login_iss(address, &address_seed, iss)
+}
+
+/// Verify that the given parameters (address_seed and iss) were used to generate the given address.
+pub fn verify_zk_login_iss(
```
benr-ml, line 189: do we need it pub?

jonas-lj: Yes, since this is also going to be called from sui-framework.

fastcrypto-zkp/src/bn254/zk_login_api.rs
```diff
@@ -169,3 +170,27 @@ fn verify_zk_login_proof_with_fixed_vk(
...
+/// Verify that the given parameters (name, value, aud, iss and salt_hash) were used to generate the
+/// given address.
+pub fn verify_zk_login_id(
+    address: &[u8],
+    name: &str,
+    value: &str,
+    aud: &str,
+    iss: &str,
+    salt_hash: &str,
+) -> FastCryptoResult<bool> {
```
benr-ml, line 183: why not FastCryptoResult<()>?
also below

jonas-lj: Why do you prefer that? I think it makes sense to distinguish between an error and a mismatch, but we've used `FastCryptoResult<()>` for the signatures so we can also use it here for consistency.

benr-ml: yes, we use () for verifications, and in this case you can return InvalidProof

PR #651, opened 2023-09-15
fix: use finalized vk

joyqvq:
prepare for finalized vk


fastcrypto-zkp/src/bn254/zk_login_api.rs
```diff
@@ -256,14 +162,10 @@ pub fn verify_zk_login(
 
 /// Verify a proof against its public inputs using the fixed verifying key.
 fn verify_zk_login_proof_with_fixed_vk(
-    usage: &ZkLoginEnv,
+    _usage: &ZkLoginEnv,
```
joyqvq, line 165: leaving the flag here as a noop for future use, if we need to change migrate vk in anyway, having a flag here would be useful.

PR #657, opened 2023-09-30
Reduce the number subtractions in `get_lagrange_coefficients` by half

jonas-lj:
It's possible to cache the differences computed in `get_lagrange_coefficients` and use this to reduce the number of subtractions by half.

I'm not sure this is a bottleneck as it is, but I did some profiling with 1000 indices and here, about 20-25% of the time is spent on the subtractions.

@benr-ml Feel free to use this if you think it's needed, otherwise you can just close the PR.


benr-ml:
Great idea!
Since in practice we compute those coefficients for the same set of parties many times, I wonder if we should generate the cache in a preprocessing stage and just pass it as an optional input. Let me think about that...

PR #672, opened 2023-10-18
DKG: clean interfaces

benr-ml:
- Introduce `complete` that combines `process_confirmations` and `aggregate`
- `merge` returns `NotEnoughInputs` in case the threshold is not yet met
- `process_message` returns an error for invalid messages (which could be skipped)
- `process_confirmations` returns `NotEnoughInputs` in case the threshold is not yet met
- `aggregate` returns `None` as the shares if the party doesn't have valid shares from the DKG (and thus cannot participate in the randomness generation)
- Fix TODOs

### DIFF
--- a/fastcrypto-tbls/src/dkg.rs
+++ b/fastcrypto-tbls/src/dkg.rs
@@ -111,6 +111,8 @@ impl<G: GroupElement, EG: GroupElement> VerifiedProcessedMessages<G, EG> {
 
 /// [Output] is the final output of the DKG protocol in case it runs
 /// successfully. It can be used later with [ThresholdBls], see examples in tests.
+///
+/// If shares is None, the object can only be used for verifying (partial and full) signatures.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Output<G: GroupElement, EG: GroupElement> {
     pub nodes: Nodes<EG>,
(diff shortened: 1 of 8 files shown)


fastcrypto-tbls/src/dkg.rs
```diff
@@ -73,13 +75,47 @@ pub struct ProcessedMessage<G: GroupElement, EG: GroupElement> {
...
+}
+
 /// [Output] is the final output of the DKG protocol in case it runs
 /// successfully. It can be used later with [ThresholdBls], see examples in tests.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Output<G: GroupElement, EG: GroupElement> {
     pub nodes: Nodes<EG>,
     pub vss_pk: Poly<G>,
-    pub shares: Vec<Share<G::ScalarType>>,
+    pub shares: Option<Vec<Share<G::ScalarType>>>, // None if some shares are missing.
```
aschran, line 118: Maybe add a bit more to the doc comment about what to do in case this is None? I guess then you can still use this to verify full sigs, but can't generate partial sigs?

fastcrypto-tbls/src/dkg.rs
```diff
@@ -404,28 +447,49 @@ where
...
+
         Output {
             nodes: self.nodes.clone(),
             vss_pk,
-            shares: final_shares.values().cloned().collect(),
+            shares,
         }
     }
 
+    /// Execute the previous two steps together.
```
aschran, line 480: Would you ever want to do only one of the previous two steps a time, or could the separate parts be removed from the public interface and only have this one be public?

aschran (approved):

aschran (approved):
Thank you, this is awesome!

aschran (approved):
Thank you, this is awesome!

PR #681, opened 2023-10-30
Fix bugs in bls12381, and add more logs to DKG

benr-ml:


fastcrypto-tbls/src/dkg.rs
```diff
@@ -155,10 +155,16 @@ where
...
         let vss_sk = PrivatePoly::<G>::rand(t - 1, rng);
 
-        debug!("Creating party {my_id}, {enc_pk:?} with threshold {t}, {random_oracle:?}");
-        if tracing::enabled!(tracing::Level::DEBUG) {
-            nodes.iter().for_each(|n| debug!("{n:?}"));
-        }
+        let vss_pk = vss_sk.commit::<G>();
+        info!(
+            "DKG: Creating party {}, nodes hash {:?}, t {}, ro {:?}, enc pk {:?}, vss pk c0 {:?}",
+            my_id,
```
aschran, line 161: optional nit - move args inline to format string where possible for better readability (e.g. `my_id`, `t`, etc.)

fastcrypto-tbls/src/dkg.rs
```diff
@@ -176,7 +182,15 @@ where
 
     /// 4. Create the first message to be broadcasted.
     pub fn create_message<R: AllowedRng>(&self, rng: &mut R) -> Message<G, EG> {
-        debug!("Creating message for party {}", self.id);
+        let vss_pk = self.vss_sk.commit();
```
aschran, line 185: not sure how expensive these calls are but consider using 

`if tracing::enabled!(tracing::Level::INFO)` or equivalent if you want to avoid the work unless the logs are turned on

aschran (approved):
ty for this painstaking debugging!

fastcrypto-tbls/src/dkg.rs
benr-ml: We need that value, I moved it from the end of the function

fastcrypto-tbls/src/dkg.rs
benr-ml: agreed, but since i also had here a value that cannot be inlined, i left everything outside (not sure if better though)

PR #689, opened 2023-11-07
DKG: More tests (nodes, nizk, ro, polynomials, ecies)

benr-ml:


fastcrypto-tbls/src/polynomial.rs
```diff
@@ -132,10 +126,14 @@ impl<C: GroupElement> Poly<C> {
     }
 
     /// Checks if a given share is valid.
-    pub fn is_valid_share(&self, idx: ShareIndex, share: &C::ScalarType) -> bool {
+    pub fn is_valid_share(&self, idx: ShareIndex, share: &C::ScalarType) -> FastCryptoResult<()> {
```
jonas-lj, line 129: I think the method should be renamed to `verify_share` or something like that if it doesn't return a boolean.

jonas-lj (approved):

PR #693, opened 2023-11-14
fix: accepts alternative iss for Google

joyqvq:


fastcrypto-zkp/src/bn254/zk_login_api.rs
```diff
@@ -227,9 +227,16 @@ pub fn verify_zk_login(
     eph_pubkey_bytes: &[u8],
     all_jwk: &ImHashMap<JwkId, JWK>,
     env: &ZkLoginEnv,
+    should_sanitize_iss: bool,
```
mskd12, line 230: When do we not sanitize iss?

joyqvq: see https://github.com/MystenLabs/sui/pull/14826. this involves a feature flag behind protocol upgrade.

mskd12 (approved):

PR #702, opened 2023-11-22
Change the multi recipient enc to a CCA secure one

benr-ml:


benr-ml:
All the interfaces are in. I didn't complete the tests yet but wanted to receive the first round of feedbacks before I add snapshots, etc.

@dariorussi , please let me know if the design makes sense, and whether any of the functions better be implemented as native ones.
@jonas-lj , please review the different crypto "conversions".

@kchalkias , @aschran - CCed in case you have comments as well.

PR #704, opened 2023-11-28
chore: suppress rsa cve

joyqvq:


benr-ml (approved):

jnaulty (approved):
> e.g. local use on a non-compromised computer is fine.

This info, and the info that we don't sign anything with this crate seems like we're in the clear here

PR #712, opened 2023-12-14
fastcrypto: impl bech32 encoding

joyqvq:
to be added as dep to sui as the standard private key encoding

### DIFF
--- a/fastcrypto/src/encoding.rs
+++ b/fastcrypto/src/encoding.rs
@@ -13,6 +13,7 @@
 //! ```
 
 use base64ct::Encoding as _;
+use bech32::{FromBase32, Variant};
 use eyre::{eyre, Result};
 use schemars::JsonSchema;
 use serde;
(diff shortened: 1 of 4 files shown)


fastcrypto/src/encoding.rs
```diff
@@ -266,3 +267,22 @@ impl<'de, const N: usize> DeserializeAs<'de, [u8; N]> for Base58 {
         Ok(array)
     }
 }
+
+/// Bech32 encoding
+pub struct Bech32(String);
+
+impl Bech32 {
+    pub fn decode(s: &str, hrp: &str) -> Result<Vec<u8>, eyre::Report> {
```
benr-ml, line 275: I'm guessing that hrp=Human Readable Part.
Let's add a comment about that, with examples.

benr-ml, line 275: encode receives hrp as an input, why not return it here as an output and let the caller verify consistency?

fastcrypto/src/encoding.rs
```diff
@@ -266,3 +267,22 @@ impl<'de, const N: usize> DeserializeAs<'de, [u8; N]> for Base58 {
...
+        let (parsed, data, variant) = bech32::decode(s).map_err(|e| eyre::eyre!(e))?;
+        if parsed != hrp || variant != Variant::Bech32 {
+            Err(eyre!("invalid hrp or variant"))
+        } else {
+            Vec::<u8>::from_base32(&data).map_err(|e| eyre::eyre!(e))
+        }
+    }
+
+    pub fn encode<T: AsRef<[u8]>>(data: T, hrp: &str) -> Result<String> {
```
benr-ml, line 284: Since it uses AsRef, data can be a ref.
(in general we should reduce dependency on AsRef for fastcrypt types, and move to non ref byte arrays / serialized objects, but we can do that later)

benr-ml (commented):
please rebase

fastcrypto/src/encoding.rs
joyqvq: revised

joyqvq: i went back and forth on this, https://github.com/MystenLabs/sui/pull/15415/files#diff-ccd99b30fee16b246ae935ee7b8c4d336530cc6962dcd611695d73f9493e8876R223 here is the usage in sui

benr-ml (approved):

PR #714, opened 2023-12-27
BLS12-381 group elements - make the code safer

benr-ml:
- Safer lifetimes
- To guarantee a unique bytes representation of GT elements, deserializing them should not use mod p. This is checked indirectly by serializing back and comparing that with the input.

### DIFF
--- a/fastcrypto/src/groups/bls12381.rs
+++ b/fastcrypto/src/groups/bls12381.rs
@@ -559,7 +559,7 @@ impl ToFromByteArray<GT_ELEMENT_BYTE_LENGTH> for GTElement {
                     let slice = &bytes[current..current + FP_BYTE_LENGTH];
                     // We compare with P_AS_BYTES to ensure that we process a canonical representation
                     // which is uses mod p elements.
-                    if slice >= &P_AS_BYTES {
+                    if *slice >= P_AS_BYTES[..] {
                         return Err(FastCryptoError::InvalidInput);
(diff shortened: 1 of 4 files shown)


fastcrypto/src/groups/bls12381.rs
```diff
@@ -554,9 +560,14 @@ impl ToFromByteArray<GT_ELEMENT_BYTE_LENGTH> for GTElement {
...
             }
         }
-        match gt.in_group() {
-            true => Ok(Self::from(gt)),
-            false => Err(FastCryptoError::InvalidInput),
+
+        // We compare with gt.to_bendian() to ensure that we process a canonical representation
+        // which is uses mod p elements.
+        // TODO: Is there a more efficient way?
+        if gt.in_group() && gt.to_bendian() == *bytes {
```
jonas-lj, line 567: Checking that the bytes is a canonical representation of a field element may be done by comparing the bytes to a big-endian representation of the field size (1a0111ea397fe69a4b1ba7b6434bacd764774b84f38512bf6730d2a0f6b0f6241eabfffeb153ffffb9feffffffffaaab) in the loop, and if one of the coefficients is not canonical the function can fail fast. This works because the ordering on byte arrays is the lexicographical ordering which is equivalent with the natural ordering on natural when using big-endian representation.

jonas-lj (changes requested):

PR #723, opened 2024-01-29
Add a simple tlock CLI

benr-ml:


fastcrypto-cli/src/tlock.rs
```diff
@@ -0,0 +1,284 @@
...
+    /// drand round number
+    #[clap(short, long)]
+    round: u64,
+
+    /// drand signature (hex, 48 bytes)
+    #[clap(short, long)]
+    signature: String,
+}
+
+/// An encryption of 32 bytes message following https://eprint.iacr.org/2023/189.pdf.
```
kchalkias, line 58: How do we encrypt shorter and larger msgs? Do we need to enable AES decryption in Sui?

kchalkias (approved):

fastcrypto-cli/src/tlock.rs
benr-ml: Today you can use the 32 bytes as an HMAC key, and impl an encryption of any size using HMAC(key, i) xor m_i where m_i is the i-th block of the message

benr-ml: (for shorter messages, pad it if needed)

PR #729, opened 2024-02-10
Faster multiplication during interpolation

benr-ml:
About %36-47 faster interpolation of 600-1100 shares.


fastcrypto-tbls/src/polynomial.rs
```diff
@@ -93,23 +103,43 @@ impl<C: GroupElement> Poly<C> {
...
-            let denominator = indices
-                .iter()
-                .filter(|j| *j != i)
-                .fold(*i, |acc, j| acc * (*j - i));
+            let mut sign = 1;
+            let (mut denominator, remaining) = indices.iter().filter(|j| *j != i).fold(
+                (C::ScalarType::from(*i), 1u128),
+                |(prev_acc, remaining), j| {
+                    let diff = if i > j {
+                        sign *= -1;
```
jonas-lj, line 125: Using a boolean for the sign is a bit cleaner, imo.

jonas-lj (approved):

PR #733, opened 2024-02-25
Add total_weight_lower_bound to Nodes::reduce

benr-ml:


fastcrypto-tbls/src/nodes.rs
```diff
@@ -146,13 +147,44 @@ impl<G: GroupElement + Serialize> Nodes<G> {
     ///   the allowed delta
     /// In practice, allowed delta will be the extra liveness we would assume above 2f+1.
     pub fn reduce(&self, t: u16, allowed_delta: u16) -> (Self, u16) {
+        self.internal_reduce(t, allowed_delta, 1) // ignores total_weight_lower_bound
+    }
+
+    /// This function should be used for e2e benchmarks only, allows limiting the level of reduction.
+    pub fn reduce_for_benchmarks(
```
aschran, line 154: this PR looks totally fine in a self-contained sense, but where do you anticipate this being called from? I don't think we are running DKG in any microbenchmarks right now, and testing fully deployed sui binaries runs the real code that would not be calling benchmark-specific functions.

benr-ml: changed to one pub API

aschran (approved):

PR #737, opened 2024-02-28
add config for new providers

joyqvq:
add - aws (test tenant), microsoft, credenza3, karrierone


benr-ml (approved):

fastcrypto-zkp/src/bn254/zklogin_test_vectors.json
```diff
@@ -0,0 +1,20 @@
+[
```
mskd12, line 1: As discussed, we need to add JWKs corresponding to these JWTs in https://github.com/MystenLabs/zklogin-circuits/blob/main/src/pastJWKs.ts.

mskd12 (approved):

PR #740, opened 2024-03-06
Add the option to deser group elements without checking if they are in the group

benr-ml:


fastcrypto/src/tests/bls12381_group_tests.rs
```diff
@@ -470,4 +481,9 @@ fn test_serialization_gt() {
     assert_eq!(carry, 0);
     bytes[0..48].copy_from_slice(&target);
     assert!(GTElement::from_byte_array(&bytes).is_err());
+
+    // Test FromTrustedByteArray.
```
jonas-lj, line 485: Could you add a test where you try to deserialize a point not in the subgroup and check that it succeeds with `from_trusted_byte_array` but fails with `from_byte_array`?

jonas-lj (changes requested):

jonas-lj (approved):

PR #744, opened 2024-03-20
DKG: Use u16 everywhere + other small fixes

benr-ml:
- use u16 instead of unsafe conversions
- more documentation
- create message for 0 weight should fail, and similarly sanity check of that message
- fix the expected threshold for process confirmations (instead of a caller supplied argument)


fastcrypto-tbls/benches/dkg.rs
```diff
@@ -67,7 +67,7 @@ mod dkg_benches {
                     |b| b.iter(|| d0.create_message(&mut thread_rng())),
                 );
 
-                let message = d0.create_message(&mut thread_rng());
+                let message = d0.create_message(&mut thread_rng()).unwrap();
```
jonas-lj, line 70: Add a note why this `unwrap` is justified (or use `except`). I think this only happens if the weight is zero.

Same for line 88 and in nidkg.rs.

jonas-lj (approved):

PR #745, opened 2024-03-20
Faster subgroup checks for BN254

jonas-lj:


jonas-lj:
This has been added to arkworks (PR #808 on arkworks-rs), so it is not needed here anyway.

PR #751, opened 2024-03-22
fix: add strong typing to Bn254Point

joyqvq:


fastcrypto-zkp/src/circom.rs
```diff
@@ -3,20 +3,100 @@
...
+    }
+}
+impl std::str::FromStr for Bn254Point {
+    type Err = Bn254PointParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let big_int = <num_bigint::BigUint as num_traits::Num>::from_str_radix(s, 10)?;
+        let be_bytes = big_int.to_bytes_be();
+        let len = be_bytes.len();
+        let mut buf = [0; 32];
```
bmwill, line 47: Can we also ensure that `s` only contains the characters 0-9 in order to ensure that ser/de roundtrips

fastcrypto-zkp/src/circom.rs
```diff
@@ -3,20 +3,100 @@
...
+
+    pub fn padded(&self) -> &[u8] {
+        &self.0
+    }
+}
+impl std::str::FromStr for Bn254Point {
+    type Err = Bn254PointParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let big_int = <num_bigint::BigUint as num_traits::Num>::from_str_radix(s, 10)?;
```
bmwill, line 44: This will work for now, but it may be worthwhile to look into using a different type from BigUint to ensure that we have fixed precision

fastcrypto-zkp/src/circom.rs
```diff
@@ -3,20 +3,100 @@
...
+pub type CircomG1 = Vec<Bn254Point>;
 
 /// A G2 point in BN254 serialized as a vector of three vectors each being a vector of two strings
 /// which are the canonical decimal representation of the coefficients of the projective coordinates
 /// in Fq2.
-pub type CircomG2 = Vec<Vec<String>>;
+pub type CircomG2 = Vec<Vec<Bn254Point>>;
+
+#[derive(Debug, Clone, JsonSchema, Eq, PartialEq)]
+pub struct Bn254Point([u8; 32]);
```
jonas-lj, line 18: This should be called Bn254FieldElement or Bn254FqElement or something like that.

fastcrypto-zkp/src/circom.rs
```diff
@@ -3,20 +3,100 @@
...
+    }
+}
+
+impl<'de> Deserialize<'de> for Bn254Point {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let s = std::borrow::Cow::<'de, str>::deserialize(deserializer)?;
+        std::str::FromStr::from_str(&s).map_err(serde::de::Error::custom)
```
jonas-lj, line 82: It should be verified, that the input is a valid field element here (e.g. that it's an integer smaller than the field size).

fastcrypto-zkp/src/circom.rs
jonas-lj: Also, I think it could be a wrapper around a `Fq` element instead of a byte array.

fastcrypto-zkp/src/circom.rs
```diff
@@ -3,20 +3,106 @@
...
 /// in Fq2.
-pub type CircomG2 = Vec<Vec<String>>;
+pub type CircomG2 = Vec<Vec<Bn254FieldElement>>;
+
+/// A struct that can represent a Bn254 point serialized as a radix10 encoded string.
+#[derive(Debug, Clone, JsonSchema, Eq, PartialEq)]
+pub struct Bn254FieldElement([u8; 32]);
+
+impl Bn254FieldElement {
+    /// Returns the unpadded version of the point. This is the point with leading zeros removed.
```
mskd12, line 23: point -> field element? (both here and below)

fastcrypto-zkp/src/circom.rs
```diff
@@ -3,20 +3,106 @@
...
-pub type CircomG1 = Vec<String>;
+pub type CircomG1 = Vec<Bn254FieldElement>;
 
 /// A G2 point in BN254 serialized as a vector of three vectors each being a vector of two strings
 /// which are the canonical decimal representation of the coefficients of the projective coordinates
 /// in Fq2.
-pub type CircomG2 = Vec<Vec<String>>;
+pub type CircomG2 = Vec<Vec<Bn254FieldElement>>;
+
+/// A struct that can represent a Bn254 point serialized as a radix10 encoded string.
```
jonas-lj, line 18: The documentation is no longer correct.

fastcrypto-zkp/src/circom.rs
```diff
@@ -3,20 +3,106 @@
...
+    }
+}
+impl std::str::FromStr for Bn254FieldElement {
+    type Err = Bn254FieldElementParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let big_int = Fq::from_str(s).map_err(|_| Bn254FieldElementParseError::InvalidInput)?;
+        let be_bytes = BigUint::from_str(&big_int.to_string())?.to_bytes_be();
+        let len = be_bytes.len();
+        let mut buf = [0; 32];
```
jonas-lj, line 51: Nit: Allocate `buf` after the size check.

fastcrypto-zkp/src/circom.rs
```diff
@@ -3,20 +3,106 @@
...
+    pub fn padded(&self) -> &[u8] {
+        &self.0
+    }
+}
+impl std::str::FromStr for Bn254FieldElement {
+    type Err = Bn254FieldElementParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let big_int = Fq::from_str(s).map_err(|_| Bn254FieldElementParseError::InvalidInput)?;
+        let be_bytes = BigUint::from_str(&big_int.to_string())?.to_bytes_be();
```
jonas-lj, line 49: Use `big_int.0.to_byte_be()` instead of the slower `from_str`.

jonas-lj: This is, btw, guaranteed to be padded to 32 bytes, so the extra length check below is not needed.

fastcrypto-zkp/src/circom.rs
```diff
@@ -3,20 +3,106 @@
...
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let s = std::borrow::Cow::<'de, str>::deserialize(deserializer)?;
+        std::str::FromStr::from_str(&s).map_err(serde::de::Error::custom)
+    }
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum Bn254FieldElementParseError {
```
jonas-lj, line 91: A public api should use `FastCryptoError`.

fastcrypto-zkp/src/circom.rs
jonas-lj: Something like this should work:
```
        let field_element = Fq::from_str(s).map_err(|_| Bn254FieldElementParseError::InvalidInput)?;
        let be_bytes = field_element.0.to_bytes_be();
        be_bytes.try_into().map_err(|_| Bn254FieldElementParseError::InvalidInput).map(Bn254FieldElement)
```
As mentioned above, the error in `try_into` will never fail, but I think it's alright to handle it anyway.

fastcrypto-zkp/src/circom.rs
```diff
@@ -3,20 +3,106 @@
...
+    #[error("Invalid input")]
+    InvalidInput,
+}
 
 /// Parse a string as a field element in BN254. Return an `FastCryptoError::InvalidInput` error if
 /// the parsing fails.
-fn parse_field_element(s: &str) -> Result<Fq, FastCryptoError> {
-    s.parse::<Fq>().map_err(|_| FastCryptoError::InvalidInput)
+fn parse_field_element(s: &Bn254FieldElement) -> Result<Fq, FastCryptoError> {
+    s.to_string()
```
jonas-lj, line 103: Avoid the string conversion and do something like `Fq::from_be_bytes_mod_order(&s.0)` instead. This never fails now, so the return value doesn't have to be a `Result` anymore.

fastcrypto-zkp/src/circom.rs
```diff
@@ -1,22 +1,142 @@
...
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let s = std::borrow::Cow::<'de, str>::deserialize(deserializer)?;
+        std::str::FromStr::from_str(&s).map_err(serde::de::Error::custom)
+    }
+}
+
+/// Convert Bn254FqElement type to arkworks' Fq.
+fn parse_fq_field_element(s: &Bn254FqElement) -> Result<Fq, FastCryptoError> {
```
jonas-lj, line 133: The `Result` is not needed here or below. It could also be implemented as `Into<Fq>`, but this works too.

jonas-lj (approved):

fastcrypto-zkp/src/circom.rs
```diff
@@ -1,22 +1,142 @@
...
+    Ok(Fq::from_be_bytes_mod_order(&s.0))
+}
+
+/// Convert Bn254FrElement type to arkworks' Fr.
+pub fn parse_fr_field_element(s: &Bn254FrElement) -> Result<Fr, FastCryptoError> {
```
mskd12, line 138: `to_arkworks_fr` seems like a better name? (same for fq)

PR #757, opened 2024-04-01
feat: add test issuers for jwt

joyqvq:


fastcrypto-zkp/src/bn254/unit_tests/zk_login_e2e_tests.rs
```diff
@@ -215,3 +216,37 @@ async fn get_test_inputs(parsed_token: &str) -> (u64, Vec<u8>, ZkLoginInputs) {
     let zk_login_inputs = ZkLoginInputs::from_reader(reader, &address_seed.to_string()).unwrap();
     (max_epoch, eph_pubkey, zk_login_inputs)
 }
+
+#[tokio::test]
+async fn test_end_to_end_test_issuer() {
+    // Nonce derived based on max_epoch = 10, kp generated from seed = [0; 32], and jwt_randomness 100681567828351849884072155819400689117.
+    let client = reqwest::Client::new();
+    let parsed_token = get_test_issuer_jwt_token(
```
mskd12, line 224: We can select a random eph pub key and generate a nonce from it to make this test more realistic, right?

mskd12 (approved):
A minor comment. Looks good otherwise.

Is there a plan to enable it in devnet?

fastcrypto-zkp/src/bn254/unit_tests/zk_login_e2e_tests.rs
joyqvq: revised to proptest

joyqvq:
> A minor comment. Looks good otherwise.
> 
> Is there a plan to enable it in devnet?

yes will be available in devnet only

PR #758, opened 2024-04-04
Make generic Groth16 implementation

jonas-lj:
This PR replaces the BLS12-381 Groth16 implementation with a generic implementation that uses the `GroupElement` and `Pairing` abstractions. This is much simpler and wraps calls to external libraries.

We can also use this for the BN254 Groth16 implementation, but it requires that we first add the BN254 construction to the supported groups.

Benchmarks have demonstrated, that the performance is unchanged.


fastcrypto-zkp/src/groth16/generic_api.rs
```diff
@@ -0,0 +1,210 @@
...
+    const GT_SIZE: usize,
+    const FR_SIZE: usize,
+>(
+    vk_gamma_abc_g1_bytes: &[u8],
+    alpha_g1_beta_g2_bytes: &[u8],
+    gamma_g2_neg_pc_bytes: &[u8],
+    delta_g2_neg_pc_bytes: &[u8],
+    proof_public_inputs_as_bytes: &[u8],
+    proof_points_as_bytes: &[u8],
+) -> Result<bool, FastCryptoError>
```
benr-ml, line 49: we usually don't return bool - why not return InvalidProof in case the proof is invalid?

fastcrypto-zkp/src/groth16/generic_api.rs
```diff
@@ -0,0 +1,210 @@
...
+        i += G2_SIZE;
+
+        let delta = G1::Other::from_byte_array(
+            bytes[i..i + G2_SIZE]
+                .try_into()
+                .map_err(|_| FastCryptoError::InvalidInput)?,
+        )?;
+        i += G2_SIZE;
+
+        let n = u64::from_le_bytes(
```
benr-ml, line 121: should we verify that n>0?

fastcrypto-zkp/src/groth16/generic_api.rs
```diff
@@ -0,0 +1,210 @@
...
+    alpha_g1_beta_g2_bytes: &[u8],
+    gamma_g2_neg_pc_bytes: &[u8],
+    delta_g2_neg_pc_bytes: &[u8],
+    proof_public_inputs_as_bytes: &[u8],
+    proof_points_as_bytes: &[u8],
+) -> Result<bool, FastCryptoError>
+where
+    G1: Pairing + ToFromByteArray<G1_SIZE> + for<'a> Deserialize<'a>,
+    <G1 as Pairing>::Other: ToFromByteArray<G2_SIZE> + for<'a> Deserialize<'a>,
+    <G1 as Pairing>::Output: GroupElement + ToFromByteArray<GT_SIZE>,
```
benr-ml, line 53: where do you use GT here?

fastcrypto-zkp/src/groth16/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+
+use fastcrypto::error::{FastCryptoError, FastCryptoResult};
+use serde::Deserialize;
+
+use fastcrypto::groups::{GroupElement, Pairing};
+
+pub(crate) mod generic_api;
+
+#[derive(Debug, Deserialize)]
+pub struct Proof<G1: Pairing>
```
benr-ml, line 14: while it's ok to how we use it today, isn't in general this is symmetric in the sense that a proof can be G2, G1, G2, where pairing is a trait of G1?

fastcrypto-zkp/src/groth16/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+
+/// This is a helper function to store a pre-processed version of the verifying key.
+/// This is roughly homologous to [`ark_groth16::data_structures::PreparedVerifyingKey`].
+/// Note that contrary to Arkworks, we don't store a "prepared" version of the `gamma_neg` and
+/// `delta_neg` fields because they are very large and unpractical to use in the binary API.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct PreparedVerifyingKey<G1>
+where
+    G1: Pairing,
+    <G1 as Pairing>::Output: Clone + Debug + PartialEq + Eq,
```
benr-ml, line 43: are those needed here?

fastcrypto/src/serde_helpers.rs
```diff
@@ -267,6 +267,34 @@ impl<const N: usize> Display for BytesRepresentation<N> {
...
+    if bytes.len() % SIZE_IN_BYTES != 0 {
+        return Err(FastCryptoError::InvalidInput);
+    }
+    bytes
+        .chunks_exact(SIZE_IN_BYTES)
+        .map(|chunk| {
+            T::from_byte_array(
+                &chunk
+                    .try_into()
+                    .map_err(|_| FastCryptoError::InvalidInput)?, // This will never fail
```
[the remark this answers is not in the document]
jonas-lj: Because the try_into succeeds if `chunk` has the right length, which is guaranteed.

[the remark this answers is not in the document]
benr-ml: i see, if so, it should not return an error but panic with expect (as something fundamental is broken in the function)

fastcrypto-zkp/src/groth16/generic_api.rs
jonas-lj: Not here, only in verify. I'll make the bounds stricter.

fastcrypto-zkp/src/groth16/mod.rs
jonas-lj: I'm not sure I follow. You'd prefer a proof to be (G1, G2, G1) because it's smaller, assuming that we've always put G1 to be the smallest, right? But yes, the pairing is symmetric, so there's nothing stopping us from implementing `Pairing` as a trait for `G1::Other` for all `G1: Pairing`.

fastcrypto-zkp/src/groth16/generic_api.rs
jonas-lj: To add the results of the pairings to do the verification.

benr-ml (commented):
Can we add regression tests in a different PR (with fixed bytes of vk & proofs that worked with the previous version, including of edge cases) and only afterwards submit this one? I want to make sure we are not breaking the current APIs

fastcrypto-zkp/src/bls12381/api.rs
```diff
@@ -1,42 +1,234 @@
...
+fn arkworks_to_gt_element(bytes: &[u8; GT_ELEMENT_BYTE_LENGTH]) -> [u8; GT_ELEMENT_BYTE_LENGTH] {
+    const PERMUTATION: [usize; 6] = [0, 2, 4, 1, 3, 5];
+    generic_convert(bytes, &PERMUTATION)
+}
+
+/// Given a serialization of a [`GTElement`], this function returns a
+/// serialization of the corresponding element as a arkworks [`PairingOutput`] type. It is _not_
+/// verified whether the input is a valid serialization..
+fn gt_element_to_arkworks(bytes: &[u8; GT_ELEMENT_BYTE_LENGTH]) -> [u8; GT_ELEMENT_BYTE_LENGTH] {
+    const PERMUTATION: [usize; 6] = [0, 3, 1, 4, 2, 5];
```
benr-ml, line 104: add a comment saying it's the opposite of the above permutation

fastcrypto-zkp/src/groth16/generic_api.rs
benr-ml: let's double check that this is backward compatible

jonas-lj: That's checked in `test_prepare_pvk_bytes` in api_tests.rs.

jonas-lj:
> Can we add regression tests in a different PR (with fixed bytes of vk & proofs that worked with the previous version, including of edge cases) and only afterwards submit this one? I want to make sure we are not breaking the current APIs

Added in #772.

benr-ml (approved):

PR #762, opened 2024-04-07
DKG: more cleanups

benr-ml:
1. Fix a possible panic/overflow in case a message with larger than MAX_U16 polynomial is sent
2. Change Nodes::reduce to new_reduced so reduction can only be applied once
3. Add more comments

### DIFF
--- a/fastcrypto-tbls/src/nodes.rs
+++ b/fastcrypto-tbls/src/nodes.rs
@@ -94,7 +94,7 @@ impl<G: GroupElement + Serialize> Nodes<G> {
     }
 
     /// Get an iterator on the share ids.
-    pub fn share_ids_iter(&self) -> impl Iterator<Item=ShareIndex> {
+    pub fn share_ids_iter(&self) -> impl Iterator<Item = ShareIndex> {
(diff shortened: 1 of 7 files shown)


fastcrypto-tbls/src/nodes.rs
```diff
@@ -135,41 +135,47 @@ impl<G: GroupElement + Serialize> Nodes<G> {
...
+    /// get the best results, set it to 1.
+    pub fn new_reduced(
+        nodes_vec: Vec<Node<G>>,
+        t: u16,
+        allowed_delta: u16,
+        total_weight_lower_bound: u16,
+    ) -> FastCryptoResult<(Self, u16)> {
+        let n = Self::new(nodes_vec)?; // checks the input, etc
+
+        assert!(total_weight_lower_bound <= n.total_weight && total_weight_lower_bound > 0);
```
jonas-lj, line 157: Is it safe to do this as an assertion?

jonas-lj (approved):

fastcrypto-tbls/src/nodes.rs
benr-ml: yes, those parameters should never not satisfy those conditions

PR #763, opened 2024-04-09
Add Bn254 `GroupElement` implementations

jonas-lj:
Implementation of GroupElement + Pairing for the BN254 curve construction.

Under the hood it's wrapping Arkworks' implementation from the `ark_bn254` crate. This is already used elsewhere, eg. in zkLogin.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -95,21 +95,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "android-tzdata"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
-
-[[package]]
-name = "android_system_properties"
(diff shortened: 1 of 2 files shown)


fastcrypto/src/groups/bn254/scalar.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    fn inverse(&self) -> FastCryptoResult<Self> {
+        Ok(Self(self.0.inverse().ok_or(FastCryptoError::InvalidInput)?))
+    }
+}
+
+impl ToFromByteArray<SCALAR_LENGTH> for Scalar {
+    fn from_byte_array(bytes: &[u8; SCALAR_LENGTH]) -> Result<Self, FastCryptoError> {
+        // Note that arkworks uses little-endian byte order for serialization here.
```
jonas-lj, line 66: The other groups uses big-endian, so perhaps we should also do it here, but the groth16 api expects little-endian representations, so it's not entirely consistent.

PR #786, opened 2024-05-13
Validate discriminants by default and add fixed discriminant

jonas-lj:
In Sui, we only plan to use a fixed discriminant, so we add that here. All other discriminants have to be verified before they are used.


fastcrypto-vdf/src/class_group/discriminant.rs
```diff
@@ -14,16 +15,29 @@ use std::ops::Neg;
...
     type Error = FastCryptoError;
 
-    /// A valid discriminant should be a negative prime congruent to 1 mod 8. The sign and
-    /// congruency are checked here but the primality is _not_. See also [Discriminant::check_primality].
+    /// A valid discriminant should be a negative prime congruent to 1 mod 8. For large discriminants,
+    /// this is very slow.
     fn try_from(value: BigInt) -> FastCryptoResult<Self> {
-        if !value.is_negative() || value.mod_floor(&BigInt::from(8)) != BigInt::from(1) {
+        if !value.is_negative()
+            || value.mod_floor(&BigInt::from(8)) != BigInt::from(1)
```
kchalkias, line 38: better use `.one()` for `BigInt::from(1)`

fastcrypto-vdf/src/class_group/mod.rs
```diff
@@ -101,8 +101,7 @@ impl QuadraticForm {
 
     /// Compute the discriminant `b^2 - 4ac` for this quadratic form.
     pub fn discriminant(&self) -> Discriminant {
-        Discriminant::try_from(self.b.pow(2) - ((&self.a * &self.c) << 2))
-            .expect("The discriminant is checked in the constructors")
+        Discriminant::from_trusted_bigint(self.b.pow(2) - ((&self.a * &self.c) << 2))
```
kchalkias, line 104: how do we ensure this will always be invoked from trusted?

fastcrypto-vdf/src/class_group/discriminant.rs
```diff
@@ -41,18 +55,25 @@ impl Discriminant {
...
-                .expect("Absolute value is non-negative"),
-        ) {
-            true => Ok(()),
-            false => Err(InvalidInput),
-        }
+    /// Create a discriminant from the given bytes. The bytes must be an BCS encoding of the
+    /// twos-complement big-endian encoding of the discriminant. Returns an error if the encoding is
+    /// invalid.
+    ///
+    /// It is assumed that the discriminant is a negative prime which is 1 mod 8. If this is not the
```
kchalkias, line 62: Do we ensure the real use case on the blockchain cannot panic? I guess we don't provide access to that, right?

jonas-lj: No, this is only called with a fixed, safe discriminant. Maybe it makes more sense to store the discriminant in fastcrypto though and just call a simple interface from the blockchain.

fastcrypto-vdf/src/class_group/mod.rs
jonas-lj: This was fixed in #787

fastcrypto-vdf/src/class_group/discriminant.rs
jonas-lj: I've updated the PR such that it is not possible to create a discriminant without verifying it. I've also added a fixed precomputed discriminant of the size, I expect we're going to use on Sui, but this will have to be updated to a discriminant generated from a seed, we don't control.

fastcrypto-cli/src/vdf.rs
```diff
@@ -133,14 +116,14 @@ fn execute(cmd: Command) -> Result<String, Error> {
...
-            if !g.has_parameter(&discriminant) {
+            if !g.has_parameter(&DISCRIMINANT_3072) {
                 return Err(Error::new(
                     ErrorKind::InvalidInput,
                     "Input point does not match discriminant.",
                 ));
             }
 
-            let vdf = DefaultVDF::new(discriminant, arguments.iterations);
+            let vdf = DefaultVDF::new(DISCRIMINANT_3072.clone(), arguments.iterations);
```
kchalkias, line 126: is there a reason this is not a reference?

fastcrypto-vdf/src/class_group/discriminant.rs
```diff
@@ -1,29 +1,51 @@
...
+    // TODO: Generate this using a seed that we provably cannot influence.
+    pub static ref DISCRIMINANT_3072: Discriminant = Discriminant(BigInt::from_str("-4080390101490206102067801750685552291425412528983716161454985565795560716833845004659207152503580931176637478422335625954692628868126419714053340412299850300602673802493259771830686596468801304317015718872352674945215883546019961626928140286675493693757393881479657605888983279619347902770789061953207866325747708864327315769009839190765716943013935708854055658243676903245686125751909996824976354309908771869043784640567352757672203749399825983258156684652782580603170228640173640869773628592618889352385821753919281706169861276929330689892675986265846043432389737049521845230769417696140636288030698887830215613149485135897148487896368642774768920061430225392365148291796645740474628778185683682893521776342856643134668770656709308404166182149870849376649591338267281149794078240401323227967073641261327798339424740171219484355109588337730742391198073121589465833677609362668436116144203312494461735357918360857667357985711").unwrap());
```
kchalkias, line 26: Awesome that works for devnet and tests

fastcrypto-vdf/src/class_group/discriminant.rs
```diff
@@ -1,29 +1,51 @@
...
 
-    /// A valid discriminant should be a negative prime congruent to 1 mod 8. The sign and
-    /// congruency are checked here but the primality is _not_. See also [Discriminant::check_primality].
+    /// A valid discriminant should be a negative prime congruent to 1 mod 8. For large discriminants,
+    /// this is very slow to check.
     fn try_from(value: BigInt) -> FastCryptoResult<Self> {
-        if !value.is_negative() || value.mod_floor(&BigInt::from(8)) != BigInt::from(1) {
+        if !value.is_negative()
+            || value.mod_floor(&BigInt::from(8)) != BigInt::one()
+            || !is_probable_prime(value.abs().magnitude())
```
kchalkias, line 47: Another thing to be careful when sampling is "non determinism". We generally avoid probabilistic functions that could have non deterministic behavior. I guess this function won't be exposed to the blockchain api. If it does, we need to ensure someone cannot cause a fork.

jonas-lj: I've actually disabled the random trials for the Miller-Rabin test (see the `is_probable_prime` functions) to ensure this, but I'll go over it again to be ceratin.

fastcrypto-cli/src/vdf.rs
jonas-lj: It's because the VDF must store the discriminant for future use, so it'll be cloned in the constructor anyway.

kchalkias (approved):

PR #787, opened 2024-05-13
Clean up trait bounds + safer group checks

jonas-lj:
The `Mul` trait is not needed for ParameterizedGroupElement's nor for the windowed multiplier.

This PR also changes the `discriminant` method for class group elements to return a BigInt instead of a wrapped `Discriminant`. This is to give a stronger guarantee that `Discriminant` objects have been validated at some point.


fastcrypto-vdf/benches/class_group.rs
```diff
@@ -3,23 +3,25 @@
...
 ) {
     let discriminant =
         Discriminant::try_from(BigInt::from_str_radix(discriminant_string, 10).unwrap()).unwrap();
     let discriminant_size = discriminant.bits();
-    let x = QuadraticForm::generator(&discriminant).mul(&BigInt::from(1234));
-    let y = QuadraticForm::generator(&discriminant).mul(&BigInt::from(4321));
+    let x =
+        QuadraticForm::hash_to_group_with_default_parameters(&[0, 1, 2], &discriminant).unwrap();
+    let y =
+        QuadraticForm::hash_to_group_with_default_parameters(&[3, 4, 5], &discriminant).unwrap();
```
kchalkias, line 24: ah previously it was just a product of generator and now they are random? Nice, that's better.

kchalkias (approved):
LGTM

PR #792, opened 2024-05-22
Misc fixes in hash module

jonas-lj:


fastcrypto/src/hash.rs
```diff
@@ -101,52 +102,51 @@ pub trait HashFunction<const DIGEST_LENGTH: usize>: Default {
...
         self.0.update(data);
     }
 
     fn finalize(self) -> Digest<DIGEST_LEN> {
-        Digest {
-            digest: self.0.finalize().as_slice().try_into().unwrap(),
-        }
+        let mut digest = [0u8; DIGEST_LEN];
+        self.0
+            .finalize_into(GenericArray::from_mut_slice(&mut digest));
```
jonas-lj, line 143: I think this does exactly what the previous code did, but I think it's a bit clearer like this

benr-ml (approved):

PR #794, opened 2024-05-23
Misc fixes and refactoring in bulletproof module

jonas-lj:


fastcrypto/src/bulletproofs.rs
```diff
@@ -51,99 +56,22 @@ impl PedersenCommitment {
...
-    fn eq(&self, other: &Self) -> bool {
-        self.point == other.point
+    fn to_byte_array(&self) -> [u8; PEDERSEN_COMMITMENT_LENGTH] {
+        self.point.compress().to_bytes()
     }
 }
 
-impl Eq for PedersenCommitment {}
-
-impl PartialOrd for PedersenCommitment {
```
jonas-lj, line 136: Are these (`PartialOrd` and `Ord`) really needed?

joyqvq: is this for sui later? we dont have to add now

fastcrypto/src/bulletproofs.rs
```diff
@@ -15,29 +15,36 @@
...
+};
+
+use serde::de;
+use serde::Deserialize;
 
 //
 // Pedersen commitments
 //
 
+// TODO: The scalars (value and blinding) are created from the lower 255 bits of a 256 bit scalar, so we should require that the last bit is zero or use the scalars directly.
```
joyqvq, line 41: 👍

joyqvq (approved):
lgtm

fastcrypto/src/bulletproofs.rs
jonas-lj: Lets add them later if they are needed.

PR #796, opened 2024-05-24
Poseidon clean up

jonas-lj:
Small clean ups:
* Docs
* Use iterator when possible
* `InputTooLong` takes the target size as parameter, not the actual size.


joyqvq (approved):
lgtm, is the plan to use this for zklogin util as well or just the poseidon move impl?

jonas-lj:
> lgtm, is the plan to use this for zklogin util as well or just the poseidon move impl?

It'll also be used in zklogin, but the behaviour doesn't change.

PR #801, opened 2024-06-05
[DKG] ECIES v1

benr-ml:
Adding v1 of the DKG that uses a different ecies encryption.

To simplify the transition, we do the following:
- old dkg & ecies were renamed to X_v0 (in a previous PR)
- moved shared structs to dkg and ecies files
- new ecies is called ecies_v1
- dkg_v1 was copied from dkg_v0 and modified to use ecies_v1 (both impl Party, when functions in dkg_v1 use the prefix _v1)
- unit tests of both was copied as well and updated to use the new interfaces

Once the transition is complete we will remove the X_v0 files.

### DIFF
--- a/fastcrypto-tbls/benches/dkg.rs
+++ b/fastcrypto-tbls/benches/dkg.rs
@@ -3,8 +3,8 @@
 
 use criterion::{criterion_group, criterion_main, BenchmarkGroup, Criterion};
 use fastcrypto::groups::{bls12381, ristretto255};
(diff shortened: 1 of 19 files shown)


jonas-lj (approved):
Looks good to me, but a bit hard to review. It looks like the differences in the dkg_x files are minimal, so there I only have a few editorial comments. For ecies, I've checked that ecies_v1 follows the paper, and that looks fine.

PR #811, opened 2024-06-27
change viz for reduce_mod_uniform_buffer

joyqvq:
useful for bls key derivation

### DIFF
--- a/fastcrypto/src/groups/bls12381.rs
+++ b/fastcrypto/src/groups/bls12381.rs
@@ -792,15 +792,25 @@ impl ScalarType for Scalar {
 ///
 /// The input buffer must be at least 48 bytes long to ensure that there is only negligible bias in
 /// the output.
-pub fn reduce_mod_uniform_buffer(buffer: &[u8]) -> Scalar {
-    assert!(buffer.len() >= 48);
+pub(crate) fn reduce_mod_uniform_buffer(buffer: &[u8]) -> Scalar {
+    match buffer_to_scalar_mod_r(buffer) {
+        Ok(scalar) => scalar,
+        Err(_) => panic!("Invalid input length"),
+    }
+}
+
+/// Similar to `reduce_mod_uniform_buffer`, returns a result of scalar, and does not panic on invalid length.
+pub fn buffer_to_scalar_mod_r(buffer: &[u8]) -> FastCryptoResult<Scalar> {
+    if buffer.len() < 48 {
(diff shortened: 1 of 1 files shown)


fastcrypto/src/groups/bls12381.rs
```diff
@@ -749,7 +749,7 @@ impl ScalarType for Scalar {
 ///
 /// The input buffer must be at least 48 bytes long to ensure that there is only negligible bias in
 /// the output.
-pub(crate) fn reduce_mod_uniform_buffer(buffer: &[u8]) -> Scalar {
+pub fn reduce_mod_uniform_buffer(buffer: &[u8]) -> Scalar {
     assert!(buffer.len() >= 48);
```
jonas-lj, line 753: If this is to be exposed publicly, I think it should return an error if the input is too small instead of panicking.

jonas-lj (changes requested):

fastcrypto/src/groups/bls12381.rs
joyqvq: revised for returning result instead of panic for public function. this is used for bls key derivation SIP impl.

fastcrypto/src/groups/bls12381.rs
```diff
@@ -793,14 +794,24 @@ impl ScalarType for Scalar {
...
-    assert!(buffer.len() >= 48);
+    match buffer_to_scalar_mod_r(buffer) {
+        Ok(scalar) => scalar,
+        Err(_) => panic!("Invalid input length"),
+    }
+}
+
+/// Similar to `reduce_mod_uniform_buffer`, returns a result of scalar, and does not panic on invalid length.
+pub fn buffer_to_scalar_mod_r(buffer: &[u8]) -> FastCryptoResult<Scalar> {
+    if buffer.len() < 48 {
```
jnaulty, line 805: perhaps a stupid question, but is there a max size to expect for this buffer?

jonas-lj: In all calls inside fastcrypto, it'll be 64 bytes. I'm not sure how the new public function is going to be used in Sui though.

joyqvq: this will be used as part of the implementation for an example custodial key server for encryption key. it will not be a dep for sui, but an util function for the key server, in this case the max length is also 64 bytes. 

revised to add a max for invalid input as well 

see more: https://github.com/sui-foundation/sips/pull/27

PR #818, opened 2024-08-04
Add Pietrzak's VDF construction

jonas-lj:
Add Pietrzak's VDF construction (https://eprint.iacr.org/2018/627.pdf). Compared to Wesolowski's construction, uniqueness of the output is unconditional, which makes it suitable for some use cases.

Proof generation works but is not optimised.


fastcrypto/src/groups/mod.rs
```diff
@@ -54,9 +54,14 @@ pub trait Scalar:
...
 
 /// Trait for group elements that has a fast doubling operation.
-pub trait Doubling {
+pub trait Doubling: Clone {
     /// Compute 2 * Self = Self + Self.
     fn double(&self) -> Self;
+
+    /// Compute input * 2^repetitions by repeated doubling.
+    fn repeated_doubling(&self, repetitions: u64) -> Self {
+        (0..repetitions).fold(self.clone(), |acc, _| acc.double())
```
benr-ml, line 63: this allocates & releases acc in every round, can't we receive `self` instead (also in double())?

fastcrypto-vdf/src/vdf/pietrzak/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+use fastcrypto::error::FastCryptoError::{InvalidInput, InvalidProof};
+use fastcrypto::error::FastCryptoResult;
+use fastcrypto::hash::{HashFunction, Keccak256};
+use num_bigint::BigUint;
+use num_integer::Integer;
+use serde::Serialize;
+use std::mem;
+
+/// Default size in bytes of the Fiat-Shamir challenge used in proving and verification.
+pub const DEFAULT_CHALLENGE_SIZE_IN_BYTES: usize = 32;
```
benr-ml, line 15: please document here the soundness of the protocol (i.e., the interactive public coin one) and why 32 bytes is enough here

fastcrypto-vdf/src/vdf/pietrzak/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+
+/// This implements Pietrzak's VDF construction from https://eprint.iacr.org/2018/627.pdf.
+/// Proofs are larger and verification is slower than in Wesolowski's construction, but the
+/// output of a VDF is unique, assuming that the used group have no small subgroups, and proving
```
benr-ml, line 19: what does it mean that proving is faster - for the same security? or, for the same number of iterations (which implies weaker security)?

fastcrypto-vdf/src/vdf/pietrzak/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+}
+
+/// Compute the Fiat-Shamir challenge used in Pietrzak's VDF construction.
+fn compute_challenge<G: ParameterizedGroupElement + Serialize>(
+    input: &G,
+    output: &G,
+    iterations: u64,
+    mu: &G,
+) -> BigUint {
+    let seed = bcs::to_bytes(&(input, output, iterations, mu))
```
benr-ml, line 115: group params?

fastcrypto-vdf/src/vdf/pietrzak/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+        .expect("Failed to serialize Fiat-Shamir input.");
+    let hash = Keccak256::digest(seed);
+    BigUint::from_bytes_be(&hash.digest[..DEFAULT_CHALLENGE_SIZE_IN_BYTES])
+}
+
+/// Replace t with (t+1) >> 1 and return true iff the input was odd.
+#[inline]
+fn check_parity_and_iterate(t: &mut u64) -> bool {
```
benr-ml, line 123: assert that t < MAX_U64

fastcrypto-vdf/src/vdf/pietrzak/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+mod tests {
+    use crate::class_group::discriminant::Discriminant;
+    use crate::class_group::QuadraticForm;
+    use crate::math::parameterized_group::Parameter;
+    use crate::vdf::pietrzak::PietrzaksVDF;
+    use crate::vdf::VDF;
+
+    #[test]
+    fn test_vdf() {
+        let iterations = 136u64;
```
benr-ml, line 137: test also edge cases (e.g., 0 iterations)

fastcrypto-vdf/src/vdf/pietrzak/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+        }
+    }
+}
+
+impl<G: ParameterizedGroupElement + Serialize> VDF for PietrzaksVDF<G> {
+    type InputType = G;
+    type OutputType = G;
+    type ProofType = Vec<G>;
+
+    fn evaluate(&self, input: &G) -> FastCryptoResult<(G, Vec<G>)> {
```
benr-ml, line 42: add a comment saying that "Proof generation works but is not optimised."

fastcrypto-vdf/src/vdf/pietrzak/mod.rs
jonas-lj: Yes, it's for the same number of iterations. Note that evaluation of the VDF is the part that should be linear in the number of iterations and proving should just be faster than evaluation, so it will not impact security.

fastcrypto-vdf/src/vdf/pietrzak/mod.rs
benr-ml: I feel that the current version is still risky in case the adversary controls t. can we gracefully handle MAX_U64?

jonas-lj: We can use the fact that (t + 1) >> 1 = t >> 1 + t%2 to avoid potentially overflowing addition.

benr-ml (approved):

PR #840, opened 2024-09-27
BLS fast sums

jonas-lj:
Adds a method to deserialize from uncompressed serialization format for G1 elements in the BLS12381 construction and add them together. This is much faster than if we first have to decompress each term:
```
Sum/BLS12381-G1/500 uncompressed
                        time:   [233.90 µs 234.58 µs 235.34 µs]

Sum/BLS12381-G1/500 compressed
                        time:   [6.2162 ms 6.2344 ms 6.2557 ms]
```

### DIFF
--- a/fastcrypto/src/groups/bls12381.rs
+++ b/fastcrypto/src/groups/bls12381.rs
@@ -349,8 +349,9 @@ impl ToFromUncompressedBytes<{ 2 * G1_ELEMENT_BYTE_LENGTH }> for G1Element {
     ) -> FastCryptoResult<G1Element> {
         // See https://github.com/supranational/blst for details on the serialization format.
 
-        // The compressed bit flag and the third bit flag (used to indicate sign of the y-coordinate
-        // for compressed representations) should not be set.
+        // Note that `blst_p1_deserialize` accepts both compressed and uncompressed serializations,
+        // so we check for the compressed bit flag which are the first and third bit flags (used to
+        // indicate sign of the y-coordinate for compressed representations) should not be set.
         if bytes[0] & 0xA0 != 0 {
             return Err(InvalidInput);
         }
@@ -359,8 +360,7 @@ impl ToFromUncompressedBytes<{ 2 * G1_ELEMENT_BYTE_LENGTH }> for G1Element {
         unsafe {
             let mut affine = blst_p1_affine::default();
 
-            // Note that `blst_p1_deserialize` accepts both compressed and uncompressed serializations
-            // which is why we checked for the compressed bit flag above.
+            // blst_p1_deserialize checks that the point is on the curve but NOT that it is in the G1 subgroup.
             if blst_p1_deserialize(&mut affine, bytes.as_ptr()) != BLST_ERROR::BLST_SUCCESS {
                 return Err(InvalidInput);
             }
@@ -576,8 +576,9 @@ impl ToFromUncompressedBytes<{ 2 * G2_ELEMENT_BYTE_LENGTH }> for G2Element {
     ) -> FastCryptoResult<G2Element> {
         // See https://github.com/supranational/blst for details on the serialization format.
 
-        // The compressed bit flag and the third bit flag (used to indicate sign of the y-coordinate
-        // for compressed representations) should not be set.
+        // Note that `blst_p2_deserialize` accepts both compressed and uncompressed serializations,
+        // so we check for the compressed bit flag which are the first and third bit flags (used to
+        // indicate sign of the y-coordinate for compressed representations) should not be set.
         if bytes[0] & 0xA0 != 0 {
             return Err(InvalidInput);
         }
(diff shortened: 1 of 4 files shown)


fastcrypto/src/groups/bls12381.rs
```diff
@@ -4,6 +4,7 @@
 use crate::bls12381::min_pk::DST_G2;
 use crate::bls12381::min_sig::DST_G1;
 use crate::encoding::{Encoding, Hex};
+use crate::error::FastCryptoError::InvalidInput;
```
benr-ml, line 7: nit - merge with next line

fastcrypto/src/groups/bls12381.rs
```diff
@@ -333,6 +335,71 @@ impl Debug for G1Element {
 serialize_deserialize_with_to_from_byte_array!(G1Element);
 generate_bytes_representation!(G1Element, G1_ELEMENT_BYTE_LENGTH, G1ElementAsBytes);
 
+/// An uncompressed seralization of a G1 element. This format is two times longer than the compressed
```
benr-ml, line 338: type

fastcrypto/src/groups/bls12381.rs
```diff
@@ -333,6 +335,71 @@ impl Debug for G1Element {
...
+
+        let affine_points: Vec<blst_p1_affine> = terms
+            .iter()
+            .map(G1ElementUncompressed::to_blst_p1_affine)
+            .collect::<FastCryptoResult<Vec<_>>>()?;
+
+        // Inspired by https://github.com/supranational/blst/blob/6f3136ffb636974166a93f2f25436854fe8d10ff/bindings/rust/src/pippenger.rs#L334-L337
+        let mut ret = blst_p1::default();
+        let p: [*const _; 2] = [&affine_points[0], ptr::null()];
+        unsafe { blst_p1s_add(&mut ret, &p[0], terms.len()) };
```
benr-ml, line 398: can you explain why the previous line is needed? why here it is not a simple pointer?

fastcrypto/src/groups/bls12381.rs
```diff
@@ -333,6 +335,71 @@ impl Debug for G1Element {
...
+/// An uncompressed seralization of a G1 element. This format is two times longer than the compressed
+/// format used by `G1Element::serialize`, but is much faster to deserialize.
+///
+/// The intended use of this struct is to deserialize and sum a large number of G1 elements without
+/// having to decompress them first.
+#[derive(Clone, Debug)]
+#[repr(transparent)]
+pub struct G1ElementUncompressed(pub(crate) [u8; 2 * G1_ELEMENT_BYTE_LENGTH]);
+
+impl From<&G1Element> for G1ElementUncompressed {
```
benr-ml, line 347: we also need the other direction (uncompressed -> compressed)

fastcrypto/src/tests/bls12381_group_tests.rs
```diff
@@ -653,3 +656,64 @@ fn test_serialization_gt() {
...
+    assert_eq!(sum, b);
+
+    // Adding zero's
+    let mut bytes = vec![G1ElementUncompressed::from(&G1Element::zero())];
+    let sum = G1ElementUncompressed::sum(&bytes).unwrap();
+    assert_eq!(sum, G1Element::zero());
+    bytes.push(G1ElementUncompressed::from(&G1Element::zero()));
+    let sum = G1ElementUncompressed::sum(&bytes).unwrap();
+    assert_eq!(sum, G1Element::zero());
+}
```
benr-ml, line 719: also let's add a short test that: chooses 100 random scalars -> computes sum using compressed vs uncompressed -> compares

fastcrypto/src/groups/bls12381.rs
```diff
@@ -333,6 +335,71 @@ impl Debug for G1Element {
...
+    /// See [the blst docs](https://github.com/supranational/blst/tree/master?tab=readme-ov-file#serialization-format) for details about the uncompressed serialization format.
+    pub fn from_trusted_byte_array(bytes: [u8; 2 * G1_ELEMENT_BYTE_LENGTH]) -> Self {
+        Self(bytes)
+    }
+
+    /// Get the byte array representation of this element.
+    pub fn into_byte_array(self) -> [u8; 2 * G1_ELEMENT_BYTE_LENGTH] {
+        self.0
+    }
+
```
benr-ml, line 370: add a comment here and above sum that the function should never fail if the uncompressed elements are indeed group elements, and might fail otherwise (but not always).

fastcrypto/src/groups/bls12381.rs
jonas-lj: Do you mean conversion from `G1ElementUncompressed` to `G1Element` or to a compressed serialization?

benr-ml: G1ElementUncompressed to G1Element

fastcrypto/src/groups/bls12381.rs
jonas-lj: I'm not sure. This is how it's done in the Rust bindings in blst, and using `&affine_points[0].as_ptr()` instead gives indeterministic results, probably because it's reading into unallocated memory. I'll say we use this method and test it throughly.

benr-ml: let's add a link to the c++ code

benr-ml (approved):

PR #853, opened 2025-02-02
fix crate vulnerabilities

rex4539:
https://rustsec.org/advisories/RUSTSEC-2024-0373

Also updated yanked crates `ahash`, `elliptic-curve`, `futures-util` and `hermit-abi`.


rex4539:
Closing due to no interest.

PR #856, opened 2025-07-10
Support non-inclusion proofs in Merkle trees

mskd12:
- Add support for non-inclusion proofs: Useful in cases the tree leaves are sorted
- Add alternative merkle tree build option where leaves are serialized (so slightly more expensive than the original which expects serialized leaves)
- Add benchmarks for merkle tree building

### DIFF
--- a/fastcrypto/benches/merkle.rs
+++ b/fastcrypto/benches/merkle.rs
@@ -9,7 +9,6 @@ mod merkle_benches {
     use super::*;
     use criterion::*;
     use fastcrypto::merkle::*;
-    use once_cell::sync::OnceCell;
     use serde::{Deserialize, Serialize};
 
     // Mimicking ObjectCheckpointState in Sui
@@ -18,22 +17,27 @@ mod merkle_benches {
         id: [u8; 32],
         data: Option<[u8; 32]>,
         #[serde(skip)]
-        bytes: OnceCell<Vec<u8>>,
+        bytes: Vec<u8>,
     }
 
     impl TestLeaf {
-        fn random() -> Self {
+        fn new(id: [u8; 32], data: Option<[u8; 32]>) -> Self {
+            let serialized = bcs::to_bytes(&(id, data)).unwrap();
             Self {
-                id: rand::random::<[u8; 32]>(),
-                data: Some(rand::random::<[u8; 32]>()),
-                bytes: OnceCell::new(),
+                id,
+                data,
+                bytes: serialized,
             }
+        }
+
+        fn random() -> Self {
+            Self::new(rand::random::<[u8; 32]>(), Some(rand::random::<[u8; 32]>()))
         }
     }
 
     impl AsRef<[u8]> for TestLeaf {
         fn as_ref(&self) -> &[u8] {
-            self.bytes.get_or_init(|| bcs::to_bytes(&self).unwrap())
+            &self.bytes
         }
     }
 


fastcrypto/src/merkle.rs
```diff
@@ -0,0 +1,519 @@
...
+            Node::Empty => EMPTY_NODE.as_ref(),
+            Node::Digest(val) => val.as_ref(),
+        }
+    }
+}
+
+/// The operations required to authenticate a Merkle proof.
+pub trait MerkleAuth: Clone + std::fmt::Debug {
+    /// Verifies the proof given a Merkle root and the leaf data.
+    #[tracing::instrument(level = Level::DEBUG, skip(self, leaf))]
```
benr-ml: why is this needed here?

fastcrypto/src/error.rs
```diff
@@ -66,3 +66,9 @@ impl From<signature::Error> for FastCryptoError {
         FastCryptoError::InvalidSignature
     }
 }
+
+impl From<bcs::Error> for FastCryptoError {
```
benr-ml: let's instead convert to the relevant error where we use bcs

fastcrypto/src/merkle.rs
```diff
@@ -0,0 +1,519 @@
...
+    fn verify_proof(&self, root: &Node, leaf: &[u8], leaf_index: usize) -> bool {
```
benr-ml: we usually use FastCryptoResult<()> for verify functions, and return error if invalid

benr-ml (commented):
It would be simpler to review this PR if we split it to two:
- One that copies the existing walrus implementation (already reviewed and used in prod, thus no need for another round of review)
- One that adds the new non inclusion proofs

fastcrypto/src/merkle.rs
mskd12: This is from the Walrus codebase. Presumably it adds logging if you initialize a tracing subscriber. Happy to remove it. cc @karlwuest if he has any thoughts / preferences.

karlwuest: IIRC, @giac-mysten had added this trait to reduce the depth of generic parameters on the symbols that get verified using the Merkle proofs. I'm not sure if it's necessary to keep this in fastcrypto, but maybe @giac-mysten can provide more context.

giac-mysten: Are we talking about the trait or about the tracing? 😅 
- tracing instrumentation is nice when debugging as it adds more structure to the logs. 
- regarding the trait, it's bee a while 😬 but IIRC karl is right, this is to have an auth interface that is independent of the details of how the merkle tree is implemented (i.e., which hash function is used internally)

fastcrypto/src/merkle.rs
```diff
@@ -144,6 +155,142 @@ where
...
+            level_index /= 2;
+        }
+        true
+    }
+}
+
+/// A proof that some data is not in a Merkle tree.
+#[derive(Serialize, Deserialize)]
+#[serde(bound(
+    serialize = "L: Serialize",
```
benr-ml, line 187: why is this needed?

fastcrypto/src/merkle.rs
```diff
@@ -144,6 +155,142 @@ where
...
+    deserialize = "L: serde::de::DeserializeOwned"
+))]
+pub struct MerkleNonInclusionProof<L, T = Blake2b256>
+where
+    L: Ord + Serialize,
+{
+    _hash_type: PhantomData<T>,
```
benr-ml: is it needed given that T is used below?

fastcrypto/src/merkle.rs
```diff
@@ -144,6 +155,142 @@ where
...
+    }
+}
+
+/// A proof that some data is not in a Merkle tree.
+#[derive(Serialize, Deserialize)]
+#[serde(bound(
+    serialize = "L: Serialize",
+    deserialize = "L: serde::de::DeserializeOwned"
+))]
+pub struct MerkleNonInclusionProof<L, T = Blake2b256>
```
benr-ml, line 190: please document above this struct what is a non inclusion proof here (what is the proof, etc)

fastcrypto/src/merkle.rs
mskd12: Hmm, it seems unnecessary. We can add a getter for the hash type later when needed.

fastcrypto/src/merkle.rs
```diff
@@ -144,6 +155,148 @@ where
...
+
+    // Proves non-inclusion of a leaf assuming sorted tree.
+    // Edge case explanations
+    // - Empty tree: no leaves, automatically valid
+    // - Target leaf smaller than all: no left neighbor, right neighbor must be at position 0
+    // - Target leaf larger than all: left neighbor must be rightmost leaf
+    pub fn verify_proof(&self, root: &Node, target_leaf: &L) -> FastCryptoResult<()> {
+        // Note: For empty trees, we don't need to check anything
+        if root.as_ref() == EMPTY_NODE.as_ref() {
+            return Ok(());
```
benr-ml, line 259: although the proof itself can be invalid?

fastcrypto/src/merkle.rs
```diff
@@ -144,6 +155,148 @@ where
...
+        }
+
+        let right_leaf_index = self.index;
+
+        if let Some((left_leaf, left_proof)) = &self.left_leaf {
+            let left_leaf_index = self.index - 1;
```
benr-ml: index can be 0 here

fastcrypto/src/merkle.rs
```diff
@@ -183,6 +336,27 @@ where
...
+        I::IntoIter: ExactSizeIterator,
+        I::Item: Serialize,
+    {
+        let leaf_hashes: FastCryptoResult<Vec<Node>> = iter
+            .into_iter()
+            .map(|leaf| {
+                let bytes = to_bytes(&leaf).map_err(|_| FastCryptoError::InvalidInput)?;
+                Ok(leaf_hash::<T>(&bytes))
+            })
+            .collect();
```
benr-ml:
```suggestion
            .collect()?;
```

fastcrypto/src/merkle.rs
mskd12: Without this, serialization fails to compile as Rust thinks that it needs to serialize T (probably because it is not able to look inside MerkleProof<T>).

fastcrypto/src/merkle.rs
mskd12: Well, there's no need for a proof in case the tree itself is empty. But if you prefer to deal with this corner case in a different way, let me know!

fastcrypto/src/merkle.rs
mskd12: thanks for the catch! i added code to handle that error

fastcrypto/src/merkle.rs
mskd12: i had to do a little more than that to make the code look clean - take a look now!

PR #857, opened 2025-07-14
Merkle tree from Walrus codebase

mskd12:
Pulling in Merkle tree from the Walrus codebase. I've only made two minor changes:
- Replaces `alloc::vec::Vec` with the one from standard library
- Removes tracing

### DIFF
--- a/fastcrypto/src/merkle.rs
+++ b/fastcrypto/src/merkle.rs
@@ -109,7 +109,12 @@ where
     }
 
     /// Verifies the proof given a Merkle root and the leaf data.
-    pub fn verify_proof(&self, root: &Node, leaf: &[u8], leaf_index: usize) -> FastCryptoResult<()> {
+    pub fn verify_proof(
+        &self,
+        root: &Node,
+        leaf: &[u8],
+        leaf_index: usize,
+    ) -> FastCryptoResult<()> {
         if self.compute_root(leaf, leaf_index).as_ref() != Some(root) {
             return Err(FastCryptoError::InvalidProof);
         }


benr-ml (approved):

fastcrypto/src/merkle.rs
```diff
@@ -0,0 +1,316 @@
...
+use alloc::{format, vec::Vec};
+use core::{fmt::Debug, marker::PhantomData};
+
+use crate::error::{FastCryptoError, FastCryptoResult};
+use crate::hash::{Blake2b256, Digest, HashFunction};
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+
+/// The length of the digests used in the merkle tree.
+pub const DIGEST_LEN: usize = 32;
```
karlwuest, line 16: Remark: I'm not sure how generic the code in fastcrypto should be. For Walrus, we were fine with a fixed size digest length, but this could also be made a generic const parameter on the Merkle Tree/Proof.

fastcrypto/src/merkle.rs
```diff
@@ -0,0 +1,316 @@
...
+
+pub const LEAF_PREFIX: [u8; 1] = [0];
+pub const INNER_PREFIX: [u8; 1] = [1];
+pub const EMPTY_NODE: [u8; DIGEST_LEN] = [0; DIGEST_LEN];
+
+/// Returned if the specified index is out of bounds for a Merkle tree or proof.
+#[derive(Error, Debug, PartialEq, Eq)]
+#[error("index {0} is too large")]
+pub struct LeafIndexOutOfBounds(usize);
```
karlwuest: Question/Suggestion: Should this also be a `FastCryptoError`?

fastcrypto/src/merkle.rs
```diff
@@ -0,0 +1,313 @@
...
+    fn as_ref(&self) -> &[u8] {
+        match self {
+            Node::Empty => EMPTY_NODE.as_ref(),
+            Node::Digest(val) => val.as_ref(),
+        }
+    }
+}
+
+/// The operations required to authenticate a Merkle proof.
+pub trait MerkleAuth: Clone + Debug {
```
karlwuest: See comment on #856, this trait can potentially be removed.

fastcrypto/src/merkle.rs
benr-ml: i think it's ok for now, we can generalise it later if needed

fastcrypto/src/merkle.rs
benr-ml: +1

PR #858, opened 2025-07-15
fix: skip instead of error for jwk parsing

joyqvq:


benr-ml (approved):

joyqvq:
force merging since the test failing is unrelated, and requires a prover deploy - @after-ephemera is looking into it.

PR #859, opened 2025-07-18
fix: cargo deny, use skip_on_invalid_jwk flag

joyqvq:
fix a few cargo denies. 

i also negated the flag, since sui does not like a feature flag that goes from true -> false (snapshots are incorrectly generated, also unconventional to other feature flags), but the other way around. changing it to reflect this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -728,9 +728,12 @@ dependencies = [
 
 [[package]]
 name = "cc"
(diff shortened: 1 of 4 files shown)


after-ephemera (approved):
lgtm!

benr-ml (approved):

PR #860, opened 2025-08-22
Jonas/secp256k1

jonas-lj:


fastcrypto/src/groups/secp256k1.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+serialize_deserialize_with_to_from_byte_array!(ProjectivePoint);
+
+impl MultiScalarMul for ProjectivePoint {
+    fn multi_scalar_mul(
+        scalars: &[Self::ScalarType],
+        points: &[Self],
+    ) -> Result<Self, FastCryptoError> {
+        multi_scalar_mul(scalars, points, &HashMap::new(), 5, ProjectivePoint::zero())
```
jonas-lj, line 90: This is 3-4x faster than naive MSM. This has been tested for 8-1024 points.

fastcrypto/src/groups/secp256k1.rs
```diff
@@ -0,0 +1,169 @@
...
+serialize_deserialize_with_to_from_byte_array!(ProjectivePoint);
+
+impl MultiScalarMul for ProjectivePoint {
+    fn multi_scalar_mul(
+        scalars: &[Self::ScalarType],
+        points: &[Self],
+    ) -> Result<Self, FastCryptoError> {
+        let scalars = scalars.iter().map(|s| s.0).collect::<Vec<_>>();
+        let points = points.iter().map(|g| g.0.into_affine()).collect::<Vec<_>>();
+        let result = Projective::msm(&points, &scalars).map_err(|_| FastCryptoError::InvalidInput)?;
```
jonas-lj: Using our own implementation of MSM is a bit faster (10%+) for small input sizes (< 64), perhaps because Arkworks requires conversion to affine coordinates, but for simplicity I think using just Arkworks' version is fine.

fastcrypto/src/tests/secp256k1_group_tests.rs
```diff
@@ -0,0 +1,118 @@
...
+        let actual = ProjectivePoint::multi_scalar_mul(&scalars, &points).unwrap();
+        assert_eq!(expected, actual);
+        assert!(ProjectivePoint::multi_scalar_mul(&scalars[1..], &points).is_err());
+    }
+}
+
+#[test]
+fn test_msm_edge_cases() {
+    assert_eq!(
+        ProjectivePoint::zero(),
```
jonas-lj, line 85: This input (two empty lists) returns an error on BLS, but I think returning zero here is also fine.

fastcrypto/src/groups/secp256k1.rs
```diff
@@ -0,0 +1,215 @@
...
+    fn multi_scalar_mul(
+        scalars: &[Self::ScalarType],
+        points: &[Self],
+    ) -> Result<Self, FastCryptoError> {
+        // Fail early if the lengths do not match
+        if scalars.len() != points.len() {
+            return Err(FastCryptoError::InvalidInput);
+        }
+        let scalars = scalars.iter().map(|s| s.0).collect::<Vec<_>>();
+        let points = points.iter().map(|g| g.0.into_affine()).collect::<Vec<_>>();
```
benr-ml: why is this conversion needed?

fastcrypto/src/groups/secp256k1.rs
benr-ml: we would use msm for sizes of 500-1200

benr-ml: what is the current performance for ~1000?

fastcrypto/src/groups/secp256k1.rs
jonas-lj: It's because `Projective::msm` (a bit confusingly, and redundant imo) takes a list of points in affine form as input. I found a batch conversion method, but I don't think the conversion can be avoided.

fastcrypto/src/groups/secp256k1.rs
jonas-lj: An MSM for 1024 points takes about 12 ms.

PR #861, opened 2025-08-25
Adapt the DKG for Seal

benr-ml:
Minimal changes to code used by the beacon.

DKG
- Allow high threshold with `new_advanced` (alternative to `new`)
- Complete the protocol without waiting for complaints with `complete_optimistic` (alternative to `complete`)

Key rotation
- Allow specifying the shared secret in `new_advanced`
- Allow specifying the threshold of previous committee in `new_advanced`
- Check that the shared secret is consistent with old partial keys in `process_message_and_check_pk`
- Allow merging also with old threshold in `merge`
- Compute shares using interpolation in `complete_optimistic_key_rotation`

See `test_e2e_dkg_and_key_rotation` for how to use those.

### DIFF
--- a/fastcrypto-tbls/src/dkg_v1.rs
+++ b/fastcrypto-tbls/src/dkg_v1.rs
@@ -190,7 +190,7 @@ where
         rng: &mut R,
     ) -> FastCryptoResult<Self> {
         // Sanity check that the threshold makes sense.
-        if t > nodes.total_weight() || t == 0 {
+        if t > nodes.total_weight() || t == 0 || old_t.is_some_and(|ot| ot == 0) {
             return Err(FastCryptoError::InvalidInput);
         }
 
@@ -772,6 +772,14 @@ where
         if self.nodes.iter().any(|n| n.weight != 1) {
             return Err(FastCryptoError::InvalidInput);
         }
+        // Check the mapping is valid: unique and all parties exist in the new committee.
+        if new_to_old_party_ids.values().collect::<HashSet<_>>().len() != new_to_old_party_ids.len()
+            || new_to_old_party_ids
+                .keys()
+                .any(|id| self.nodes.node_id_to_node(*id).is_err())
+        {
+            return Err(FastCryptoError::InvalidInput);
+        }
 
         // Do not filter out any message, assume all are valid.
         let messages = VerifiedProcessedMessages::filter_from(messages, &[]);
@@ -796,9 +804,10 @@ where
         }
 
         // Interpolate my shares.
-        let mut shares = Vec::new();
-        for m in messages.0.iter() {
-            let share = Share {
+        let shares = messages
+            .0
+            .iter()
+            .map(|m| Share {
                 index: *new_sender_to_old_share_id
                     .get(&m.message.sender)
                     .expect("checked above"),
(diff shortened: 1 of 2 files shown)


fastcrypto-tbls/src/polynomial.rs
```diff
@@ -201,6 +211,14 @@ impl<C: Scalar> Poly<C> {
         Self::from(coeffs)
     }
 
+    /// Returns a new polynomial of the given degree where the constant term is
+    /// fixed to `c0` and the rest of the coefficients are sampled at random.
+    pub fn rand_fixed_c0<R: AllowedRng>(degree: u16, c0: C, rng: &mut R) -> Self {
+        let mut coeffs: Vec<C> = (0..degree).map(|_| C::rand(rng)).collect();
+        coeffs.insert(0, c0);
```
jonas-lj: I think it'd be better to handle this in the iterator to avoid the O(n) insertion here and the vector of unclear size.

fastcrypto-tbls/src/dkg_v1.rs
```diff
@@ -702,6 +749,105 @@ where
...
+                + 1; // +1 since the old party ids are 0-based.
+            new_sender_to_old_share_id.insert(
+                m.message.sender,
+                ShareIndex::new(old_share_id).expect("nonzero"),
+            );
+        }
+
+        // Interpolate my shares.
+        let mut shares = Vec::new();
+        for m in messages.0.iter() {
```
jonas-lj: Is there a reason the `Share` structs are not just constructed in the loop above? Or perhaps as an iterator?

fastcrypto-tbls/src/dkg_v1.rs
```diff
@@ -702,6 +749,105 @@ where
...
+        Ok(self.aggregate(&verified_messages))
+    }
+
+    /// Alternative to complete() - Optimistic variant that interpolates the shares instead of
+    /// summing them, to be used for key rotation. Works only if all weights are 1.
+    pub fn complete_optimistic_key_rotation(
+        &self,
+        messages: &UsedProcessedMessages<G, EG>,
+        // Mapping party id from new committee to its id in the previous committee
+        new_to_old_party_ids: &HashMap<PartyId, PartyId>,
```
joyqvq, line 769: validate the mapping here? 1) check no duplicate for new ids? 2) id < committee size?

fastcrypto-tbls/src/dkg_v1.rs
```diff
@@ -453,7 +499,8 @@ where
                     .weight as u32
             })
             .sum::<u32>();
-        if total_weight < (self.t as u32) {
+        let required_t = self.old_t.unwrap_or(self.t);
+        if total_weight < (required_t as u32) {
```
joyqvq, line 503:
```suggestion
        if total_weight < (required_t as u32) || required_t == 0 {
```

fastcrypto-tbls/src/dkg_v1.rs
```diff
@@ -168,19 +169,44 @@ where
...
+        enc_sk: ecies_v1::PrivateKey<EG>,
+        nodes: Nodes<EG>,
+        t: u16, // The number of parties that are needed to reconstruct the full key/signature (f+1).
+        random_oracle: RandomOracle, // Should be unique for each invocation, but the same for all parties.
+        secret: Option<G::ScalarType>, // Should be used only for key rotation.
+        old_t: Option<u16>,          // Used only for key rotation
+        rng: &mut R,
+    ) -> FastCryptoResult<Self> {
+        // Sanity check that the threshold makes sense.
+        if t > nodes.total_weight() || t == 0 {
```
joyqvq: also check old_t != 0 if not none?

fastcrypto-tbls/src/dkg_v1.rs
benr-ml: here the caller cannot do much about it, tested in new_advanced

PR #862, opened 2025-08-26
Add Schnorr acc. to BIB-0340

jonas-lj:
Implement Schnorr signatures according the the Bitcoin specs: https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki

It's tested that it follows the test vectors in https://github.com/bitcoin/bips/blob/master/bip-0340/test-vectors.csv.

Note that this does _not_ implement the `Authenticator` trait like the other signature schemes in fastcrypto since I don't think we're going to use it in Sui and because the `as_ref` trait is a bit hard to work with, but it wouldn't be too much work to do that.

### DIFF
--- a/fastcrypto/src/groups/secp256k1.rs
+++ b/fastcrypto/src/groups/secp256k1.rs
@@ -294,7 +294,6 @@ pub mod schnorr {
     use crate::hash::HashFunction;
     use crate::serde_helpers::ToFromByteArray;
     use crate::{hash, serialize_deserialize_with_to_from_byte_array};
-    use ark_ec::CurveGroup;
 
     pub const SIGNATURE_SIZE_IN_BYTES: usize = 64;
     pub const PUBLIC_KEY_SIZE_IN_BYTES: usize = 32;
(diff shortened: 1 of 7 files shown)


fastcrypto/src/groups/secp256k1.rs
```diff
@@ -199,25 +273,446 @@ impl ScalarTrait for Scalar {
...
     }
 }
 
-impl ToLittleEndianBytes for Scalar {
-    fn to_le_bytes(&self) -> Vec<u8> {
-        self.to_byte_array().to_vec()
+serialize_deserialize_with_to_from_byte_array!(Scalar);
+
+/// A Schnorr signature scheme as defined in BIP-0340 (https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki).
+pub mod schnorr {
```
jonas-lj, line 301: I'm not sure how much of this we are going to use, but we can mark it test-only until we know.

fastcrypto/src/groups/secp256k1.rs
```diff
@@ -199,25 +273,446 @@ impl ScalarTrait for Scalar {
...
+        Ok(Scalar::from_bytes_mod_order(bytes))
     }
 
     fn to_byte_array(&self) -> [u8; SCALAR_SIZE_IN_BYTES] {
-        let mut bytes = [0u8; SCALAR_SIZE_IN_BYTES];
-        self.0
-            .serialize_uncompressed(&mut bytes[..])
-            .expect("Byte array not large enough");
-        bytes
+        self.0.into_bigint().to_bytes_be().try_into().unwrap()
```
jonas-lj: Align with serialization in schnorr

benr-ml (approved):

PR #863, opened 2025-08-28
Batched AVSS

jonas-lj:


fastcrypto-tbls/src/batched_avss/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+use serde::{Deserialize, Serialize};
+use std::borrow::Borrow;
+use std::marker::PhantomData;
+use tracing::debug;
+use zeroize::Zeroize;
+
+/// This represents a Dealer in the AVSS. There is exactly one dealer, who creates the shares and broadcasts the encrypted shares.
+pub struct Dealer<G, EG: GroupElement> {
+    number_of_nonces: u16,
+    f: u16,
```
benr-ml: let's use t for the threshold

fastcrypto-tbls/src/batched_avss/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+    public_keys: Vec<PublicKey<EG>>,
```
benr-ml: let's use Nodes?

fastcrypto-tbls/src/batched_avss/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+}
+
+#[derive(Clone, Debug)]
+pub struct Complaint<EG: GroupElement> {
+    party_id: PartyId,
+    proof: RecoveryPackage<EG>,
+}
+
+pub trait Certificate<G: GroupElement, EG: GroupElement> {
+    fn message(&self) -> Message<G, EG>;
```
benr-ml:
```suggestion
    fn message(&self) -> Message<G, EG>;
```
let's use bytes here (since this should be the hash of the message anyhow)

fastcrypto-tbls/src/batched_avss/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+        }
+
+        Ok(shares)
+    }
+
+    /// 4. When 2f+1 signatures have been collected in the certificate, the receivers can now verify it.
+    ///  - If the receiver is already in the certificate, do nothing.
+    ///  - If it is not in the certificate, but it is able to decrypt and verify its shares, store the shares and do nothing.
+    ///  - If it is not in the certificate and cannot decrypt or verify its shares, it creates a complaint with a recovery package for its shares.
+    pub fn process_certificate(
```
benr-ml: this should receive Message and a certificate on its hash, no?

fastcrypto-tbls/src/batched_avss/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+    EG::ScalarType: FiatShamirChallenge + Zeroize,
+{
+    /// 2. Each receiver processes the message, verifies and decrypts its shares. If this works, the shares are stored and the receiver can contribute a signature on the message to a certificate.
+    pub fn process_message(
+        &self,
+        message: &Message<G, EG>,
+    ) -> FastCryptoResult<Shares<G::ScalarType>> {
+        // TODO: Sanity checks?
+
+        let random_oracle_encryption = self.random_oracle_extension(Encryption);
```
benr-ml: verify the degree of p_double_prime

fastcrypto-tbls/src/batched_avss/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+                        bcs::from_bytes::<Shares<G::ScalarType>>(bytes.as_slice())
+                            .map_err(|_| FastCryptoError::InvalidInput)
+                            .map(|decrypted| (ShareIndex::new(*id).unwrap(), decrypted))
+                    })
+                    .ok()
+            })
+            .collect::<FastCryptoResult<Vec<_>>>()?;
+
+        if shares.len() < (self.f + 1) as usize {
+            return Err(FastCryptoError::InvalidInput);
```
benr-ml: InputTooShort? the caller cannot do much with InvalidInput here

fastcrypto-tbls/src/batched_avss/mod.rs
jonas-lj: The message was included in the certificate, but we can have them as two inputs also

fastcrypto-tbls/src/batched_avss/mod.rs
jonas-lj: I don't think InputTooShort makes sense, since this error indicates that there is enough responses but some of them are invalid. I've changed it to a general error with a more informative message.

fastcrypto/Cargo.toml
```diff
@@ -56,7 +56,6 @@ num-bigint = "0.4.4"
 bech32 = "0.9.1"
 getrandom = {version = "0.2.15", features = ["js"], optional = true}
 bcs.workspace = true
-k256 = { version = "0.11.6", features = ["hash2curve"] }
```
jonas-lj, line 59: Hash2Group is not needed for k1 anyway.

fastcrypto-tbls/src/polynomial.rs
```diff
@@ -245,3 +253,37 @@ impl<C: GroupElement + MultiScalarMul> Poly<C> {
         Ok(res)
     }
 }
+
+pub fn interpolate_at_index<C: Scalar>(
```
benr-ml: add tests

fastcrypto-tbls/src/hashi/avss.rs
```diff
@@ -0,0 +1,959 @@
...
+use itertools::Itertools;
+use serde::{Deserialize, Serialize};
+use std::fmt::Debug;
+use std::marker::PhantomData;
+use tap::TapFallible;
+use tracing::warn;
+use zeroize::Zeroize;
+
+/// This represents a Dealer in the AVSS.
+/// There is exactly one dealer,who creates the shares and broadcasts the encrypted shares.
```
benr-ml:
```suggestion
/// There is exactly one dealer, who creates the shares and broadcasts the encrypted shares.
```

fastcrypto-tbls/src/hashi/avss.rs
```diff
@@ -0,0 +1,959 @@
...
+            })
+            != message.response.eval(self.index).value
+        {
+            return Err(InvalidInput);
+        }
+        Ok(())
+    }
+}
+
+/// A batch of shares for a single share index, containing shares for each secret and one for the "blinding" polynomial.
```
benr-ml: move above the impl

fastcrypto-tbls/src/hashi/avss.rs
```diff
@@ -0,0 +1,959 @@
...
+        Ok(Self {
+            secrets,
+            threshold,
+            nodes,
+            random_oracle,
+            _group: PhantomData,
+        })
+    }
+
+    /// 1. The Dealer generates shares for the secrets and broadcasts the encrypted shares.
```
benr-ml: create the ecies key -> create nodes -> new dealer -> create_message

fastcrypto-tbls/src/hashi/avss.rs
```diff
@@ -0,0 +1,959 @@
...
+    EG::ScalarType: FiatShamirChallenge + Zeroize,
+    G::ScalarType: FiatShamirChallenge,
+{
+    pub fn new(
+        secrets: Vec<G::ScalarType>,
+        nodes: Nodes<EG>,
+        threshold: u16, // The number of parties that are needed to reconstruct the full key/signature (f+1).
+        random_oracle: RandomOracle, // Should be unique for each invocation, but the same for all parties.
+    ) -> FastCryptoResult<Self> {
+        // Sanity check that the threshold makes sense (t <= n/2 since we later wait for 2t-1).
```
benr-ml: we will need to wait for n-f where f is the maximal number of non responding nodes (liveness), i think we don't need to restrict the threshold here

fastcrypto-tbls/src/hashi/avss.rs
```diff
@@ -0,0 +1,959 @@
...
+
+    pub fn my_weight(&self) -> usize {
+        self.nodes
+            .total_weight_of(std::iter::once(&self.id))
+            .unwrap() as usize
+    }
+
+    /// 2. Each receiver processes the message, verifies and decrypts its shares. If this works, the shares are stored and the receiver can contribute a signature on the message to a certificate.
+    pub fn process_message(&self, message: &Message<G, EG>) -> FastCryptoResult<ReceiverOutput<G>> {
+        if message.response.degree() > self.threshold as usize {
```
benr-ml: why not != t-1?

fastcrypto-tbls/src/hashi/avss.rs
```diff
@@ -0,0 +1,959 @@
...
+    }
+
+    pub fn my_weight(&self) -> usize {
+        self.nodes
+            .total_weight_of(std::iter::once(&self.id))
+            .unwrap() as usize
+    }
+
+    /// 2. Each receiver processes the message, verifies and decrypts its shares. If this works, the shares are stored and the receiver can contribute a signature on the message to a certificate.
+    pub fn process_message(&self, message: &Message<G, EG>) -> FastCryptoResult<ReceiverOutput<G>> {
```
benr-ml: we should distinguish here between sanity checks that would fail for everyone, and checks that fail for this specific receiver. in the first case, all parties will simply ignore that message (no need for fraud proofs, etc, no one should have shares).

fastcrypto-tbls/src/hashi/avss.rs
```diff
@@ -0,0 +1,959 @@
...
+
+        // Check that we received the correct number of shares.
+        if my_shares.weight() != self.my_weight() {
+            return Err(InvalidInput);
+        }
+
+        self.verify_shares(message, &my_shares)?;
+
+        // Verify that g^{p''(0)} == c' * prod_l c_l^{gamma_l}
+        let challenge = self.compute_challenge_from_message(message);
```
benr-ml: those are sanity checks that should move above the checks that use my decryption key

fastcrypto-tbls/src/hashi/avss.rs
```diff
@@ -0,0 +1,959 @@
...
+        })
+    }
+
+    /// 3. When 2t+1 signatures have been collected in the certificate, the receivers can now verify it.
+    ///  - If the receiver is already in the certificate, return [ProcessCertificateResult::Ignore].
+    ///  - If it is not in the certificate, but it is able to decrypt and verify its shares, return [ProcessCertificateResult::Valid] with its shares.
+    ///  - If it is not in the certificate and cannot decrypt or verify its shares, it returns a [ProcessCertificateResult::Complaint] with a complaint.
+    ///
+    /// Returns an error if the certificate or the encrypted shares are invalid.
+    pub fn process_certificate(
```
benr-ml: i don't think this is needed - we can assume that all receivers will always call `process_message`. if ok, the higher level protocol should sign (to be part of the cert).
next, if the higher level protocol sees a cert but received error from process_message, it should call a function like create_complaint and broadcast it.

fastcrypto-tbls/src/hashi/avss.rs
```diff
@@ -0,0 +1,959 @@
...
+        Ok(ComplaintResponse::create(
+            self.id,
+            &message.ciphertext,
+            &self.enc_secret_key,
+            self,
+            rng,
+        ))
+    }
+
+    /// 5. Upon receiving f+1 valid responses to a complaint, the accuser can recover its shares.
```
benr-ml:
```suggestion
    /// 5. Upon receiving t valid responses to a complaint, the accuser can recover its shares.
```

fastcrypto-tbls/src/hashi/avss.rs
```diff
@@ -0,0 +1,959 @@
...
+            })
+            .collect_vec();
+
+        // Compute the total weight of the valid responses
+        let response_weight = response_shares
+            .iter()
+            .map(SharesForNode::weight)
+            .sum::<usize>();
+        if response_weight < (self.threshold + 1) as usize {
+            return Err(FastCryptoError::GeneralError(
```
benr-ml: this should InputTooShort and can remove the above weight checks

fastcrypto-tbls/src/hashi/avss.rs
```diff
@@ -0,0 +1,959 @@
...
+                "Not enough valid responses".to_string(),
+            ));
+        }
+
+        let my_shares = SharesForNode::recover(self.my_indices(), &response_shares)?;
```
benr-ml: when should those fail?

fastcrypto-tbls/src/hashi/avss.rs
```diff
@@ -0,0 +1,959 @@
...
+    fn compute_challenge(
+        &self,
+        c: &[G],
+        c_prime: &G,
+        e: &MultiRecipientEncryption<EG>,
+    ) -> Vec<G::ScalarType> {
+        let random_oracle = self.extension(Challenge);
+        c.iter()
+            .enumerate()
+            .map(|(l, c_l)| random_oracle.evaluate(&(l, c_l, c_prime, e)))
```
benr-ml: this is insecure, it needs to evaluate the RO on all inputs

fastcrypto-tbls/src/hashi/avss.rs
```diff
@@ -0,0 +1,959 @@
...
+
+impl<C: Scalar + Serialize> BCSSerialized for SharesForNode<C> {}
+
+impl<G: GroupElement + Serialize, EG: GroupElement + HashToGroupElement + Serialize> Dealer<G, EG>
+where
+    EG::ScalarType: FiatShamirChallenge + Zeroize,
+    G::ScalarType: FiatShamirChallenge,
+{
+    pub fn new(
+        secrets: Vec<G::ScalarType>,
```
benr-ml: do we need them to be provided by the caller? can't we generate them internally?

fastcrypto-tbls/src/hashi/avss.rs
```diff
@@ -0,0 +1,959 @@
...
+use serde::{Deserialize, Serialize};
+use std::fmt::Debug;
+use std::marker::PhantomData;
+use tap::TapFallible;
+use tracing::warn;
+use zeroize::Zeroize;
+
+/// This represents a Dealer in the AVSS.
+/// There is exactly one dealer,who creates the shares and broadcasts the encrypted shares.
+pub struct Dealer<G: GroupElement, EG: GroupElement> {
```
benr-ml: what about adding the batch size to the generics? to use fixed sized arrays everywhere
i think in any case we would work with a small set of fixed batch sizes in the higher level protocol (e.g., 10, 50, 100)

fastcrypto-tbls/src/hashi/dkg.rs
```diff
@@ -0,0 +1,752 @@
...
+    G::ScalarType: FiatShamirChallenge,
+{
+    pub fn new(
+        secrets_batch: Vec<G::ScalarType>,
+        nodes: Nodes<EG>,
+        threshold: u16, // The number of parties that are needed to reconstruct the full key/signature (f+1).
+        random_oracle: RandomOracle, // Should be unique for each invocation, but the same for all parties.
+    ) -> FastCryptoResult<Self> {
+        // Sanity check that the threshold makes sense (t <= n/2 since we later wait for 2t-1).
+        if threshold > (nodes.total_weight() / 2) {
```
benr-ml: as before, let's remove

fastcrypto-tbls/src/hashi/dkg.rs
```diff
@@ -0,0 +1,752 @@
...
+    pub fn my_weight(&self) -> usize {
+        self.nodes
+            .total_weight_of(std::iter::once(&self.id))
+            .unwrap() as usize
+    }
+
+    /// 2. Each receiver processes the message, verifies and decrypts its shares. If this works, the shares are stored and the receiver can contribute a signature on the message to a certificate.
+    pub fn process_message(&self, message: &Message<G, EG>) -> FastCryptoResult<ReceiverOutput<G>> {
+        let random_oracle_encryption = self.extension(Encryption);
+
```
benr-ml: - sanity checks should fail differently (as above)
- check the deg of commitments

fastcrypto-tbls/src/hashi/dkg.rs
```diff
@@ -0,0 +1,752 @@
...
+            .commitments
+            .iter()
+            .zip(&self.previous_round_commitments)
+        {
+            if commitment.c0() != previous {
+                return Err(InvalidInput);
+            }
+        }
+
+        // Verify shares against commitments.
```
benr-ml: let's add a todo to verify them more efficiently using msm (see dkg_v1)

benr-ml (commented):
let's rename a bit:
- hashi -> tschnorr or threshold_schnorr
- avss -> batch_avss
- dkg -> avss

also, should we use concrete groups instead of generics? it will simplify the code, and I don't see a use case in the near future for other groups

fastcrypto-tbls/src/hashi/avss.rs
jonas-lj: The dealer could generate shares in such a way that the highest coefficient is zero. It could also happen with negligible probability, so that can probably be ignored. But the check should be >= t instead.

benr-ml: why larger and not exact?

jonas-lj: Oh, yea. It should be >=

fastcrypto-tbls/src/hashi/avss.rs
jonas-lj: If response_shares is empty or if they are malformed (have different batch sizes)

fastcrypto-tbls/src/hashi/avss.rs
jonas-lj: Sure, that would simplify it a bit.

fastcrypto-tbls/src/threshold_schnorr/avss.rs
```diff
@@ -0,0 +1,815 @@
...
+use std::array;
+use std::collections::HashMap;
+use tap::TapFallible;
+use tracing::warn;
+
+/// This represents a Dealer in the AVSS. There is exactly one dealer, who creates the shares and broadcasts the encrypted shares.
+pub struct Dealer<const BATCH_SIZE: usize> {
+    t: u16,
+    nodes: Nodes<EG>,
+    random_oracle: RandomOracleWrapper,
```
benr-ml: maybe let's store the sid instead, for debugging purposes

fastcrypto-tbls/src/threshold_schnorr/avss.rs
```diff
@@ -0,0 +1,815 @@
...
+///
+/// These can be created either by decrypting the shares from the dealer (see [Receiver::process_message]) or by recovering them from complaint responses.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct SharesForNode<const BATCH_SIZE: usize> {
+    pub batches: Vec<ShareBatch<BATCH_SIZE>>,
+}
+
+impl<const BATCH_SIZE: usize> ShareBatch<BATCH_SIZE> {
+    fn verify(&self, message: &Message<BATCH_SIZE>) -> FastCryptoResult<()> {
+        for (share, c) in self.shares.iter().zip(message.feldman_commitments.iter()) {
```
benr-ml: add a TODO to do faster using msms

fastcrypto-tbls/src/threshold_schnorr/batch_avss.rs
```diff
@@ -0,0 +1,807 @@
...
+            .filter_map(|response| {
+                self.nodes
+                    .node_id_to_node(response.responder_id)
+                    .and_then(|node| {
+                        response.decrypt_with_response(
+                            &self.random_oracle,
+                            &node.pk,
+                            &message.ciphertext,
+                        )
+                    })
```
benr-ml, line 444: we should also check here if the decrypted share is valid (in case the dealer & that receiver are colluding)

fastcrypto-tbls/src/threshold_schnorr/avss.rs
```diff
@@ -0,0 +1,698 @@
...
+use serde::{Deserialize, Serialize};
+use tap::TapFallible;
+use tracing::warn;
+
+/// This represents a Dealer in the AVSS. There is exactly one dealer, who creates the shares and broadcasts the encrypted shares.
+#[allow(dead_code)]
+pub struct Dealer {
+    t: u16,
+    nodes: Nodes<EG>,
+    sid: String,
```
benr-ml: hmm, perhaps just bytes? it will likely be randomOracle.evaluate(strings...)

fastcrypto-tbls/src/threshold_schnorr/avss.rs
```diff
@@ -0,0 +1,698 @@
...
+    random_oracle: RandomOracleWrapper,
```
benr-ml: do we still need this given sid?

fastcrypto-tbls/src/threshold_schnorr/avss.rs
```diff
@@ -0,0 +1,698 @@
...
+    Valid(ReceiverOutput),
+    Complaint(Complaint),
+}
+
+/// The output of a receiver: The shares for each nonce + commitments for the next round.
+#[derive(Debug, Clone)]
+pub struct ReceiverOutput {
+    pub my_shares: SharesForNode,
+
+    /// The commitments to the polynomials for the next round.
```
benr-ml:
```suggestion
    /// The commitments to the polynomials, will be used for key rotation.
```

fastcrypto-tbls/src/threshold_schnorr/avss.rs
```diff
@@ -0,0 +1,698 @@
...
+use tracing::warn;
+
+/// This represents a Dealer in the AVSS. There is exactly one dealer, who creates the shares and broadcasts the encrypted shares.
+#[allow(dead_code)]
+pub struct Dealer {
+    t: u16,
+    nodes: Nodes<EG>,
+    sid: String,
+    random_oracle: RandomOracleWrapper,
+    secret: S,
```
benr-ml: do we use secret?

fastcrypto-tbls/src/threshold_schnorr/avss.rs
```diff
@@ -0,0 +1,698 @@
...
+            .filter_map(|response| {
+                self.nodes
+                    .node_id_to_node(response.responder_id)
+                    .and_then(|node| {
+                        response.decrypt_with_response(
+                            &self.random_oracle,
+                            &node.pk,
+                            &message.ciphertext,
+                        )
+                    })
```
benr-ml, line 329: need to verify the response as well

fastcrypto-tbls/src/threshold_schnorr/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+pub mod ro_extension;
+pub mod si_matrix;
+
+/// The group to use for the signing
+pub type G = groups::secp256k1::ProjectivePoint;
+
+/// Default scalar
+pub type S = <G as GroupElement>::ScalarType;
+
+/// The group used for multi-recipient encryption
```
benr-ml: pls remind me - why bls? let's document that

fastcrypto-tbls/src/threshold_schnorr/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+type EG = G1Element;
+
+/// Generate partial threshold Schnorr signatures for a given message using a presigning triple.
+///
+/// Returns also the public nonce R.
+pub fn generate_partial_signatures<const BATCH_SIZE: usize>(
+    my_indices: &[ShareIndex],
+    message: &[u8],
+    presignatures: &mut Presignatures<BATCH_SIZE>,
+    my_signing_key_shares: &[S],
```
benr-ml: why not use the shares object?

fastcrypto-tbls/src/threshold_schnorr/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+) -> FastCryptoResult<(G, Vec<Eval<S>>)> {
+    // One share per weight for this party
+    if my_signing_key_shares.len() != my_indices.len() {
+        return Err(FastCryptoError::InvalidInput);
+    }
+
+    // TODO: Each output from an instance of Presigning has a unique index. Perhaps this is needed for coordination?
+    let (_, secret_presigs, public_presig) = presignatures
+        .next()
+        .ok_or(GeneralError("No more pre-signatures".to_string()))?;
```
benr-ml: let's add a concrete error

fastcrypto-tbls/src/threshold_schnorr/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+
+    let r = public_presig + G::generator() * beacon_value;
```
benr-ml: let's call it R, or r_g? to distinguish between field and group elements

fastcrypto/src/groups/secp256k1.rs
```diff
@@ -444,7 +412,7 @@ pub mod schnorr {
         hasher.finalize().into()
     }
 
-    fn hash_to_scalar<'a>(tag: Tag, data: impl IntoIterator<Item = &'a [u8]>) -> Scalar {
+    pub fn hash_to_scalar<'a>(tag: Tag, data: impl IntoIterator<Item = &'a [u8]>) -> Scalar {
```
benr-ml:
```suggestion
    pub fn bip0340_hash_to_scalar<'a>(tag: Tag, data: impl IntoIterator<Item = &'a [u8]>) -> Scalar {
```

fastcrypto-tbls/src/threshold_schnorr/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+
+    let sigma_prime = Poly::recover_c0(
+        threshold,
+        partial_signatures.iter().take(threshold as usize),
+    )?;
+    let s = sigma_prime + beacon_value;
+
+    let signature = (r, s);
+    verify(vk, &signature, message)?;
```
benr-ml, line 99: add a todo to handle invalid sigs

fastcrypto-tbls/src/threshold_schnorr/avss.rs
jonas-lj: No, it's just for convenience so it won't have to be constructed every time it's used.

fastcrypto-tbls/src/threshold_schnorr/avss.rs
jonas-lj: Yes, in case we're using it to share a share from a previous round. Perhaps there needs to be an option for the very first round where there are no commitments?

fastcrypto-tbls/src/threshold_schnorr/mod.rs
jonas-lj: It just needs a good hash-to-group function, so it could also be Ristretto. I'm removing the hash-to-group function for k1 because it doesn't seem to be standardized, but we could also use that.

fastcrypto-tbls/src/threshold_schnorr/avss.rs
jonas-lj: I've removed it as a field element and just kept the sid.

fastcrypto-tbls/src/threshold_schnorr/mod.rs
jonas-lj: I've updated the comment and changed to Ristretto. I think that's a bit more efficient.

fastcrypto-tbls/src/random_oracle.rs
```diff
@@ -35,6 +39,11 @@ impl RandomOracle {
         }
     }
 
+    pub fn from_sid(sid: &[u8]) -> Self {
```
benr-ml: the current design of RO guarantees that no matter how the inputs to evaluate and extend were chosen, as long as they are unique, we get unique outputs. however this function breaks that since now (in theory) i can  create new RO with new("sid"|sid) and from_sid(sid). let's change that to new(sid).extend(Extensions.SID)?

fastcrypto-tbls/src/threshold_schnorr/avss.rs
benr-ml: oh, let's leave it

fastcrypto-tbls/src/threshold_schnorr/avss.rs
```diff
@@ -0,0 +1,703 @@
...
+impl Dealer {
+    /// Create a new dealer.
+    ///
+    /// * `secret`: The secret to share.
+    /// * `nodes`: The set of nodes (parties) participating in the protocol, including their public keys and weights.
+    /// * `t`: The threshold number of shares required to reconstruct the secret. One party can have multiple shares according to its weight.
+    /// * `f`: An upper bound on the number of Byzantine parties counted by weight.
+    /// * `sid`: A session identifier that should be unique for each invocation of the protocol but the same for all parties in a single invocation.
+    pub fn new(
+        secret: S,
```
benr-ml: i'd have this one as optional and generate it if needed

fastcrypto/src/error.rs
```diff
@@ -52,6 +52,10 @@ pub enum FastCryptoError {
     #[error("Message should be ignored")]
     IgnoredMessage,
 
+    /// The presigs iterator has no more values. Please create a new iterator.
+    #[error("Out of presigs in the iterator, please create a new iterator")]
```
benr-ml:
```suggestion
    #[error("Out of presigs in the iterator, please create new presigs")]
```

fastcrypto-tbls/src/threshold_schnorr/presigning.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+}
+
+impl<const BATCH_SIZE: usize> ExactSizeIterator for Presignatures<BATCH_SIZE> {}
+
+impl<const BATCH_SIZE: usize> Presignatures<BATCH_SIZE> {
+    /// Based on the output of a batched AVSS from multiple dealers, create a presignature generator.
+    pub fn new(
+        my_indices: &[ShareIndex],
+        avss_outputs: &[ReceiverOutput<BATCH_SIZE>],
```
benr-ml: nit/todo - maybe we should consume this param, to prevent double use (risky)

benr-ml (approved):

fastcrypto-tbls/src/threshold_schnorr/avss.rs
jonas-lj: And then do the same for the commitment on the receiver, right?

benr-ml: yes

PR #866, opened 2025-09-22
Ensure that signatures are bib340 compatible

jonas-lj:


fastcrypto-tbls/src/threshold_schnorr/signing.rs
```diff
@@ -0,0 +1,130 @@
...
+    presignatures: &mut Presignatures<BATCH_SIZE>,
+    my_signing_key_shares: &avss::SharesForNode,
+    verifying_key: &G,
+    beacon_value: &S,
+) -> FastCryptoResult<(G, Vec<Eval<S>>)> {
+    // TODO: Each output from an instance of Presigning has a unique index. Perhaps this is needed for coordination?
+    let (_, mut secret_presigs, public_presig) = presignatures.next().ok_or(OutOfPresigs)?;
+
+    let r_g = public_presig + G::generator() * beacon_value;
+    if r_g == G::zero() {
```
benr-ml, line 36: let's doc that this should never happen

fastcrypto-tbls/src/threshold_schnorr/signing.rs
```diff
@@ -0,0 +1,130 @@
...
+/// `GeneralOpaqueError` is returned if the computed nonce R is the identity element.
+/// `InvalidSignature` is returned if the aggregated signature does not verify.
+/// `InvalidInput` is returned if the provided verifying key is the identity element.
+pub fn aggregate_signatures(
+    message: &[u8],
+    public_presig: &G,
+    partial_signatures: &[Eval<S>],
+    beacon_value: &S,
+    threshold: u16,
+    vk: &G,
```
benr-ml:
```suggestion
    verifying_key: &G,
```

fastcrypto-tbls/src/threshold_schnorr/signing.rs
```diff
@@ -0,0 +1,130 @@
...
+) -> FastCryptoResult<SchnorrSignature> {
+    if partial_signatures.len() < threshold as usize {
```
benr-ml, line 94: check uniqueness of share ids

fastcrypto-tbls/src/threshold_schnorr/signing.rs
```diff
@@ -0,0 +1,130 @@
...
+    )?;
+
+    // Compute the nonce R for the signature.
+    let r_g = public_presig + G::generator() * beacon_value;
+    if r_g == G::zero() {
+        return Err(FastCryptoError::GeneralOpaqueError);
+    }
+
+    // In acc. with BIP-0340, we need to ensure the nonce R has an even Y coordinate.
+    // If it doesn't, we subtract the beacon value instead of adding it.
```
benr-ml:
```suggestion
    // If it doesn't, we subtract the beacon value instead of adding it like done for the secret shares.
```

benr-ml (approved):

PR #871, opened 2025-09-30
Key derivation from index in threshold Schnorr

jonas-lj:


fastcrypto-tbls/src/threshold_schnorr/key_derivation.rs
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::threshold_schnorr::{G, S};
+use fastcrypto::groups::GroupElement;
+use fastcrypto::hmac::{hkdf_sha3_256, HkdfIkm};
+use fastcrypto::traits::ToFromBytes;
+
+const HKDF_DST: &[u8] = b"FASTCRYPTO-TBLS-KEY-DERIVATION-v1";
```
benr-ml: update

fastcrypto-tbls/src/threshold_schnorr/key_derivation.rs
```diff
@@ -0,0 +1,25 @@
...
+
+/// Compute a tweak from a verifying key and a derivation index.
+pub(crate) fn compute_tweak(vk: &G, index: u64) -> S {
+    let mut ikm: Vec<u8> = vk.x_as_be_bytes().unwrap().to_vec(); // 32 bytes
+    ikm.extend_from_slice(&index.to_be_bytes()); // 8 bytes
+
+    // Derive 64 uniform bytes to avoid bias from modular reduction
```
benr-ml: more bytes to reduce bias, no?

fastcrypto-tbls/src/threshold_schnorr/key_derivation.rs
```diff
@@ -0,0 +1,25 @@
...
+    let bytes = hkdf_sha3_256(&HkdfIkm::from_bytes(&ikm).unwrap(), &[], HKDF_DST, 64).unwrap();
+    S::from_bytes_mod_order(&bytes)
+}
+
+/// Derive a new verifying key from an existing one and a derivation index.
+/// This is computed as vk + [compute_tweak](vk, index) * G.
+pub(crate) fn derive_verifying_key(vk: &G, index: u64) -> G {
```
benr-ml: should this already be a valid bip0340 key? i think it should since we would use that as a deposit address

fastcrypto-tbls/src/threshold_schnorr/key_derivation.rs
jonas-lj: I'm not sure I follow. What should this be instead?

fastcrypto-tbls/src/threshold_schnorr/key_derivation.rs
jonas-lj: Yes, that is not precise. Will update.

fastcrypto-tbls/src/threshold_schnorr/key_derivation.rs
jonas-lj: I've added another public function to derive the actual key and made it explicit that this function is only for internal use.

fastcrypto-tbls/src/threshold_schnorr/key_derivation.rs
jonas-lj: I've removed it completely. We can add an info/context field if needed.

PR #872, opened 2025-09-30
Add RS decoder

jonas-lj:
Implement a RS decoder with Gao's algorithm. This relies on implementations of polynomial division and extended GCD for polynomials, so the polynomial module has been extended accordingly.

### DIFF
--- a/fastcrypto-tbls/src/polynomial.rs
+++ b/fastcrypto-tbls/src/polynomial.rs
@@ -48,7 +48,7 @@ impl<C: GroupElement> Poly<C> {
     }
 
     /// Removes leading zero coefficients.
-    fn reduce(&mut self) {
+    pub(crate) fn reduce(&mut self) {
         self.0.truncate(self.degree() + 1);
     }
 }
--- a/fastcrypto-tbls/src/tests/polynomial_tests.rs
+++ b/fastcrypto-tbls/src/tests/polynomial_tests.rs
@@ -22,7 +22,7 @@ mod scalar_tests {
     use itertools::Itertools;
 
     #[test]
-    fn test_degree<S: Scalar>() {
+    fn test_degree_bound<S: Scalar>() {
         let s: usize = 5;
         let p = Poly::<S>::rand(s as u16, &mut thread_rng());
         assert_eq!(p.degree_bound(), s);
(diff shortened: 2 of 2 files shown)


fastcrypto-tbls/src/polynomial.rs
```diff
@@ -29,18 +30,29 @@ pub type PublicPoly<C> = Poly<C>;
...
-impl<C> From<Vec<C>> for Poly<C> {
+impl<C: GroupElement> From<Vec<C>> for Poly<C> {
     fn from(c: Vec<C>) -> Self {
-        Self(c)
+        if c.is_empty() {
+            return Self::zero();
+        }
+        let mut p = Self(c);
+        p.reduce();
+        p
```
jonas-lj: @benr-ml This changes the existing implementation, which is used in the random beacon, but it should be okay. The `degree` function is at least not working correctly until this is done.

arnab-roy:
Jonas, I ran some additional tests on the decoder. It's producing correct results and seems to be pretty efficient around n = O(100). However, it's running ~300-400s on my machine for n=1000. Is that acceptable or we want to optimize more? I can push the new tests to your branch if that helps.

jonas-lj:
> Jonas, I ran some additional tests on the decoder. It's producing correct results and seems to be pretty efficient around n = O(100). However, it's running ~300-400s on my machine for n=1000. Is that acceptable or we want to optimize more? I can push the new tests to your branch if that helps.

That's a good point. The implementation is rather naive, so there are a few low hanging optimisations, we could try. If it breaks that much for n=1000, I suspect just using iterators could be a worthwhile optimisation. I'll work on it and get back to you.

fastcrypto-tbls/src/polynomial.rs
benr-ml: this change should be behind a flag when merged to sui, since otherwise an attacker that sends a message with a polynomial that is reduced can cause a fork between validators that run the new code vs ones that run the old code

jonas-lj: @benr-ml I've renamed the old `degree` function to `degree_bound` to ensure that the old code is not changed. Please take a look.

PR #899, opened 2025-12-06
Arnab/swiper

arnab-roy:
This PR adds a "super_swiper" algorithm for weight reduction in the `fastcrypto-tbls` crate, adds supporting infrastructure for weight reduction experiments, and enhances the codebase with tools and data for working with real validator sets.

The super_swiper code is adapted from: https://github.com/tolikzinovyev/weight-reduction

The implementation of the algorithms presented in the paper:
Weight reduction in distributed protocols: new algorithms and analysis, by Anatoliy Zinovyev
(https://eprint.iacr.org/2025/1076).


fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,539 @@
...
+}
+
+// Helper types for generating deltas
+#[derive(Eq, PartialEq)]
+struct QueueElement {
+    s: Ratio,
+    i: usize,
+}
+
+impl Ord for QueueElement {
```
jonas-lj, line 177: You can just derive `Ord` and `PartialOrd` here since that automatically does lexicographical ordering from the first field and down.

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,539 @@
...
+    fn update(&mut self, index: usize) {
+        while index >= self.tickets.len() {
+            self.tickets.push(0);
+        }
+        self.tickets[index] += 1;
+        self.total += 1;
+    }
+
+    fn get(&self, index: usize) -> u64 {
+        match self.tickets.get(index) {
```
jonas-lj: This is the same as `unwrap_or`

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,539 @@
...
+impl Tickets {
+    fn new() -> Self {
+        Self {
+            tickets: Vec::new(),
+            total: 0,
+        }
+    }
+
+    fn update(&mut self, index: usize) {
+        while index >= self.tickets.len() {
```
jonas-lj: Here, you can use `self.tickets.resize` instead.

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,539 @@
...
+    fn adversarial_tickets(&self) -> u64 {
+        for (t, &w) in self.dp.iter().enumerate().rev() {
+            if w != 0 {
+                return t as u64;
+            }
+        }
+        0
+    }
+    fn adv_tickets_target(&self) -> u64 {
+        self.dp.capacity() as u64
```
jonas-lj: Do you mean length here? I'm not a 100% sure, but I don't think you can guarantee anything about the capacity except that it's larger than what you've requested.

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,539 @@
...
+            self.dp[t] = w;
+        }
+
+        Some(self)
+    }
+
+    // Returns the maximum achievable adversarial number of tickets.
+    #[cfg(test)]
+    fn adversarial_tickets(&self) -> u64 {
+        for (t, &w) in self.dp.iter().enumerate().rev() {
```
jonas-lj: You can use `find_position` here.

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,539 @@
...
+            dp.push(x);
+        }
+
+        Some(DP {
+            max_weight: self.max_weight,
+            dp,
+        })
+    }
+
+    // Apply an element with weight w and t tickets. Returns None iff
```
jonas-lj: Function docs should start with a triple dash, `///`

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,539 @@
...
+struct Generator<'a> {
+    weights: &'a [u64],
+    c: Ratio,
+    r: usize,
+    queue: BinaryHeap<QueueElement>,
+}
+
+impl<'a> Generator<'a> {
+    fn new(weights: &'a [u64], c: Ratio) -> Self {
+        assert!(!weights.is_empty());
```
jonas-lj: For at least some of these, we'd probably want to return an error instead of asserting

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,539 @@
...
+        debug_assert!(weights.windows(2).all(|w| w[0] >= w[1]));
+        assert!(*weights.last().unwrap() > 0);
+        assert!(c >= 0.into());
+        assert!(c < 1.into());
+
+        let mut queue = BinaryHeap::new();
+        queue.push(QueueElement {
+            s: (Ratio::from_integer(1) - c) / Ratio::from_integer(*weights.first().unwrap()),
```
jonas-lj, line 207: nit: It looks like there's an `Ratio::one` function

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,539 @@
...
+    Generator::new(weights, c)
+}
+
+// Calculates the head indices for the current batch.
+fn calc_indices_head(tickets_len: usize, deltas: &[usize]) -> Vec<usize> {
+    let mut indices_tail = deltas.to_vec();
+    indices_tail.sort_unstable_by(|a, b| b.cmp(a));
+    indices_tail.dedup();
+
+    let mut indices_head: Vec<_> = (0..tickets_len).collect();
```
jonas-lj: Using swap_remove here may modify the order of the remaining elements, so is this safe? Also, isn't there a better way to do this instead of creating a vector and then remove elements?

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,539 @@
...
+pub fn solve(alpha: Ratio, beta: Ratio, weights: &[u64]) -> Vec<u64> {
+    debug_assert!(weights.windows(2).all(|w| w[0] >= w[1]));
+
+    let max_adv_weight = calc_max_adv_weight_from_weights(alpha, weights);
+
+    let mut tickets = Tickets::new();
+    let mut g = generate_deltas(weights, alpha);
+
+    let mut batch_size: usize = 1;
+    loop {
```
jonas-lj, line 402: Can you write a note about why this doesn't run forever?

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,539 @@
...
+            Some(&x) => x,
+            None => 0,
+        }
+    }
+
+    fn data(&self) -> &[u64] {
+        &self.tickets
+    }
+
+    fn extract_data(self) -> Vec<u64> {
```
jonas-lj: It's common to use the name `to_vec` for something like this and `as_slice` for the one above.

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,539 @@
...
+
+        for i in (1..self.dp.len()).rev() {
+            if self.dp[i] != 0 {
+                let ww = self.dp[i] + w;
+                let tt = i + t as usize;
+                if ww <= self.max_weight {
+                    if tt >= adv_tickets_target {
+                        return None;
+                    }
+                    while self.dp.len() <= tt {
```
jonas-lj: Use `resize` here

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,539 @@
...
+    fn make_copy(&self, adv_tickets_target: u64) -> Option<DP> {
+        assert!(adv_tickets_target <= self.dp.capacity() as u64);
+
+        let adv_tickets_target = adv_tickets_target as usize;
+        if adv_tickets_target < self.dp.len() {
+            return None;
+        }
+
+        let mut dp = Vec::with_capacity(adv_tickets_target);
+        for &x in self.dp.iter().take(adv_tickets_target) {
```
jonas-lj: You can just collect the iterator as `dp` here

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,539 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// The implementation of the algorithms presented in the paper
```
jonas-lj: Module level docs should start with `//!`

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,539 @@
...
+            return Some(self);
+        }
+        let adv_tickets_target = self.dp.capacity();
+        if t as usize >= adv_tickets_target {
+            return None;
+        }
+
+        for i in (1..self.dp.len()).rev() {
+            if self.dp[i] != 0 {
+                let ww = self.dp[i] + w;
```
jonas-lj: Are these names, ww and tt, from the paper? I find them a bit confusing

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,539 @@
...
+                    }
+                    if (ww < self.dp[tt]) || (self.dp[tt] == 0) {
+                        self.dp[tt] = ww;
+                    }
+                }
+            }
+        }
+
+        let t = t as usize;
+        while self.dp.len() <= t {
```
jonas-lj: Use `resize`

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,539 @@
...
+    tickets: &Tickets,
+) -> Option<DP> {
+    let mut dp_head = DP::new(
+        max_adv_weight,
+        calc_adv_tickets_target(beta, tickets.total() + deltas.len() as u64),
+    )
+    .unwrap();
+
+    let indices_head = calc_indices_head(tickets.data().len(), deltas);
+    for index in indices_head {
```
jonas-lj: nit: This can be done with an iterator and the reduce function, but it's perhaps a matter of taste if it's cleaner

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,539 @@
...
+    dp_head: &DP,
+    tickets: &Tickets,
+    adv_tickets_target: u64,
+    add_indices: &[usize],
+    exclude_indices: &[usize],
+) -> Option<DP> {
+    let mut dp = dp_head.make_copy(adv_tickets_target)?;
+
+    let exclude_indices = exclude_indices.iter().copied().collect::<BTreeSet<_>>();
+    let add_indices = add_indices.iter().copied().collect::<BTreeSet<_>>();
```
jonas-lj: Are these already sorted? In that case you don't need to copy into a sorted set but can just use the `binary_search` function

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,539 @@
...
+
+    for index in add_indices {
+        if !exclude_indices.contains(&index) {
+            dp = dp.apply(weights[index], tickets.get(index))?;
```
jonas-lj: This can be done with an iterator and the `filter` and `reduce` functions

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,539 @@
...
+        }
+    }
+
+    Some(dp)
+}
+
+fn update_tickets(deltas: &[usize], tickets: &mut Tickets) {
```
jonas-lj: Why is this (and perhaps also the functions below) not a function defined on `Tickets`?

fastcrypto-tbls/src/weight_reduction/mod.rs
arnab-roy: Renamed to be clearer.

fastcrypto-tbls/src/weight_reduction/mod.rs
arnab-roy: Added explanation.

fastcrypto-tbls/src/weight_reduction/mod.rs
arnab-roy: Added cleaner code.

fastcrypto-tbls/src/weight_reduction/mod.rs
arnab-roy: Added as an attribute.

fastcrypto-tbls/src/weight_reduction/mod.rs
arnab-roy: Generally doesn't seem to be sorted.

fastcrypto-tbls/src/weight_reduction/mod.rs
arnab-roy: Refactored update_tickets, but not sure about batch_recursive as it's coordinating many structures.

fastcrypto-tbls/src/weight_reduction/mod.rs
arnab-roy: Deriving seems to order in the opposite direction than expected.

jonas-lj: In that case, you can just switch the order of the fields in the struct definition.

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,552 @@
...
+    loop {
+        tickets.update(g.next().unwrap());
+        let deltas: Vec<_> = (&mut g).take(batch_size - 1).collect();
+
+        let ret = process_batch(beta, weights, max_adv_weight, &deltas, &mut tickets);
+        if ret {
+            return tickets.into_vec();
+        }
+
+        batch_size *= 2;
```
jonas-lj: Is there an upper bound on the number of iterations which guarantees that this doesn't overflow?

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,552 @@
...
+}
+
+/// Apply `deltas` to provided `tickets`. If after applying 0 or more
+/// deltas, a valid ticket assignment is found, returns true with
+/// `tickets` containing the corresponding ticket assignment.
+/// Otherwise, returns false with `tickets` containing
+/// the new ticket assignment after applying all deltas.
+/// `dp_head` must have all indices applied except those in `deltas` with data
+/// in `tickets`.
+fn process_batch_recursive(
```
jonas-lj, line 303: Rust has a max call depth, so we need some upper bound on the number of recursions to make sure this will not panic - do we have that?

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,552 @@
...
+    /// Returns the maximum achievable adversarial number of tickets.
+    #[cfg(test)]
+    fn adversarial_tickets(&self) -> u64 {
+        self.dp
+            .iter()
+            .rposition(|&w| w != 0)
+            .map(|t| t as u64)
+            .unwrap_or(0)
+    }
+    fn adv_tickets_target(&self) -> u64 {
```
jonas-lj: It doesn't seem like this function is used anywhere?

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,552 @@
...
+    fn new() -> Self {
+        Self {
+            tickets: Vec::new(),
+            total: 0,
+        }
+    }
+
+    fn update(&mut self, index: usize) {
+        if index >= self.tickets.len() {
+            self.tickets.resize(index + 1, 0);
```
jonas-lj: You don't need the if statement here - if the resize target is already met, nothing happens.

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,552 @@
...
+        // Sanity check: `dp_head` must have the right target.
+        let adv_tickets_target = calc_adv_tickets_target(beta, tickets.total());
+        assert!(dp_head.adv_tickets_target() == adv_tickets_target);
+        return true;
+    }
+
+    // Number of tickets assignments in the left branch.
+    let left_branch_size = (deltas.len() + 1) / 2;
+
+    {
```
jonas-lj: Why do you need these scopes?

fastcrypto-tbls/src/weight_reduction/mod.rs
arnab-roy: It's used in line 363

fastcrypto-tbls/src/weight_reduction/mod.rs
arnab-roy: Removing the if check and calling resize(index + 1, 0) unconditionally would truncate the vector when index < self.tickets.len(), causing data loss and "N/A" results.

fastcrypto-tbls/src/weight_reduction/mod.rs
arnab-roy: Removed

fastcrypto-tbls/src/weight_reduction/mod.rs
arnab-roy: Yes, it's log(delta.len()) roughly as each recursive call splits the delta vector into half. Adding a comment to clarify this.

fastcrypto-tbls/src/weight_reduction/mod.rs
arnab-roy: Yes, it's capped by weights.len(). Adding an explicit check to ensure and a comment to clarify.

fastcrypto-tbls/src/weight_reduction/mod.rs
arnab-roy: It's not just the order of the fields, but also descending order. I can modify the iterator, but this way seems cleaner. Happy to change if derive is better for other reasons as well.

fastcrypto-tbls/src/weight_reduction/mod.rs
jonas-lj: Oh yes, you're right

fastcrypto-tbls/src/weight_reduction/mod.rs
```diff
@@ -0,0 +1,569 @@
...
+    /// adversarial tickets target. Returns None iff the new adversarial tickets
+    /// target has already been achieved.
+    fn make_copy(&self, adv_tickets_target: u64) -> Option<DP> {
+        assert!(adv_tickets_target <= self.adv_tickets_target);
+
+        if self.dp.len() > adv_tickets_target as usize {
+            return None;
+        }
+
+        let dp = self
```
jonas-lj: @arnab-roy If dp.len > adv_tickets_target we return, so there's no need to take the first adv_tickets_target of dp here, right?

fastcrypto-tbls/src/weight_reduction/weight_reduction_checks.rs
```diff
@@ -0,0 +1,194 @@
...
+///
+/// # Returns
+/// The maximum delta value from all checks, or None if target cannot be reached
+///
+/// # Delta Calculation
+/// For each validity check (top, bottom, or random):
+/// 1. Take reduced weights to reach 2t' - 1
+/// 2. Let w1 = sum of old weights corresponding to the subset
+/// 3. Then delta = w1 - (2t - 1), where t is the input threshold
+pub fn get_delta(
```
jonas-lj, line 30: This is not used anymore, right?

arnab-roy: Kept it there for future testing, but we can remove once confident in the precision loss based parameter tuning.

benr-ml:
@arnab-roy let's remove everything related to super_swiper since we are not planning to use it

PR #901, opened 2025-12-08
Add DKG Observer

jonas-lj:


fastcrypto-tbls/src/dkg_v1.rs
```diff
@@ -983,3 +985,488 @@ where
...
+
+/// [Output] is the final output of the DKG protocol in case it runs
+/// successfully. It can be used later with [ThresholdBls], see examples in tests.
+///
+/// If shares is None, the object can only be used for verifying (partial and full) signatures.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct ObserverOutput<G: GroupElement, EG: GroupElement> {
+    pub nodes: Nodes<EG>,
+    pub vss_pk: Poly<G>,
+}
```
akichidis: Currently `Output` bears the following structure

```
pub struct Output<G: GroupElement, EG: GroupElement> {
    pub nodes: Nodes<EG>,
    pub vss_pk: Poly<G>,
    pub shares: Option<Vec<Share<G::ScalarType>>>, // None if some shares are missing or weight is zero.
}
```

where the `shares` is already an `Option`. Can the Observer use the `Output` as well in this case and just pass the `shares` as `None`? It would help on the API alignment between `Party` and `Observer`

fastcrypto-tbls/src/dkg_v1.rs
```diff
@@ -983,3 +985,488 @@ where
...
+            .fold(Poly::zero(), |acc, m| acc + &m.vss_pk);
+
+        ObserverOutput {
+            nodes: self.nodes.clone(),
+            vss_pk,
+        }
+    }
+
+    /// Execute the previous two steps together.
+    pub fn complete<R: AllowedRng>(
```
akichidis: Why is the generic `R: AllowedRng` needed?

jonas-lj: You're right - it's not used anymore. Probably leftovers from the code I copied.

fastcrypto-tbls/src/tests/dkg_v1_tests.rs
```diff
@@ -1017,3 +1018,238 @@ fn test_e2e_dkg_and_key_rotation() {
...
+
+    assert!(o0.shares.is_some());
+    assert!(o1.shares.is_some());
+    assert!(o2.shares.is_some());
+    assert_eq!(o0.vss_pk, o1.vss_pk);
+    assert_eq!(o0.vss_pk, o2.vss_pk);
+
+    // Run DKG as observer:
+
+    // Compute the observers output. The Observer has no shares but only the vss_pk
```
akichidis: Maybe we should add an assert that indeed no shares produced for Observer

fastcrypto-tbls/src/dkg_v1.rs
```diff
@@ -983,3 +998,421 @@ where
...
+            random_oracle,
+            old_t,
+            g: PhantomData,
+        })
+    }
+
+    /// Compute the output of a DKG protocol given all the dealer messages from the protocol.
+    /// If successful, return the Output which contains the verifying key. The shares field will
+    /// be None since the observer does not hold any shares.
+    pub fn observe_dkg(
```
akichidis: Do I understand correctly that this is mostly a helper method? Assuming I have all the messages I can just call this to compute the DKG output? Otherwise I have to do the `process_message` my own for one by one the messages and then eventually call the `complete` method?

fastcrypto-tbls/src/dkg_v1.rs
```diff
@@ -983,3 +998,421 @@ where
...
+            .fold(Poly::zero(), |acc, m| acc + &m.vss_pk);
+
+        Output {
+            nodes: self.nodes.clone(),
+            vss_pk,
+            shares: None,
+        }
+    }
+
+    /// Execute the previous two steps together.
```
akichidis: I would suggest to mention specifically what the previous steps are specifically. Although someone can tell from the code, but the comment it self depends on some kind of order which is not clear.

fastcrypto-tbls/src/dkg_v1.rs
```diff
@@ -983,3 +998,421 @@ where
...
+                    .weight as u32
+            })
+            .sum::<u32>();
+        let required_t = self.old_t.unwrap_or(self.t);
+        if total_weight < (required_t as u32) {
+            debug!("Merge failed with total weight {total_weight}");
+            return Err(FastCryptoError::NotEnoughInputs);
+        }
+
+        info!("DKG: Merging messages with total weight {total_weight}");
```
akichidis, line 1101: Is `info` the right log level here?

fastcrypto-tbls/src/dkg_v1.rs
```diff
@@ -983,3 +998,421 @@ where
...
+        let filtered_messages = self.merge(all_messages)?;
+        self.complete_optimistic_key_rotation(&filtered_messages, new_to_old_party_ids)
+    }
+
+    /// The threshold needed to reconstruct the full key/signature.
+    pub fn t(&self) -> u16 {
+        self.t
+    }
+
+    /// 5. Process a message and create the second message to be broadcasted.
```
akichidis: I guess for the Observer broadcasting the second message doesn't apply right? Since we just verify the messages here, we do not send any confirmations to other parties

akichidis (commented):
I don't have opinion on the inner crypto mechanics, but thinks look good to me. For sure @benr-ml will need to confirm. 
As an overall comment, and we have discussed this offline, the code duplication with `Party` is not ideal for future maintenance, but if we believe it's safer that way then I am fine with it. Otherwise if there are parts that could safely be deduplicated I would advise to do so.
In any case thanks a lot @jonas-lj for crafting this so quickly 💯

fastcrypto-tbls/src/dkg_v1.rs
jonas-lj: Yes, exactly. I think I worked with an assumption that this was exactly what you needed for your integration, but if it's not needed anymore, we could consider getting rid of it.

fastcrypto-tbls/src/dkg_v1.rs
jonas-lj: Yes, you're right - that doc is not accurate. I'll update

fastcrypto-tbls/src/dkg_v1.rs
jonas-lj: `info` is aligned with how it's done for `Party`, but I don't have a strong preference

akichidis: If it's aligned with `Party` then no issue from my side.

fastcrypto-tbls/src/dkg_v1.rs
```diff
@@ -983,3 +998,421 @@ where
...
+/// An observer of a the DKG ceremony.
+///
+/// Can be instantiated with G1Curve or G2Curve.
+impl<G, EG> Observer<G, EG>
+where
+    G: GroupElement + MultiScalarMul + Serialize + DeserializeOwned,
+    EG: GroupElement + Serialize + HashToGroupElement + DeserializeOwned,
+    EG::ScalarType: FiatShamirChallenge + Zeroize,
+{
+    /// 1. Create a new ECIES private key and send the public key to all parties.
```
benr-ml: update the comments here and below (e.g., ECIES private key)

fastcrypto-tbls/src/dkg_v1.rs
```diff
@@ -983,3 +998,421 @@ where
...
+        if t > (nodes.total_weight() / 2) {
+            return Err(FastCryptoError::InvalidInput);
+        }
+        Self::new_advanced(nodes, t, random_oracle, None)
+    }
+
+    /// Alternative to new(), to be used for:
+    /// - Any threshold t, useful in the sync setting.
+    /// - Passing old threshold for key rotation (should be set to None for DKG).
+    pub fn new_advanced(
```
benr-ml: can move to new, no need for new_advanced

fastcrypto-tbls/src/dkg_v1.rs
```diff
@@ -983,3 +998,421 @@ where
...
+            message.vss_pk.c0()
+        );
+        // Ignore if invalid (and other honest parties will ignore as well).
+        message.sanity_check(&self.nodes, self.t, |sender| {
+            self.encryption_random_oracle(sender)
+        })
+    }
+
+    /// Alternative to process_message for a known vss.c0 public key, useful for key rotation.
+    pub fn process_message_and_check_pk(
```
benr-ml: no need for this one

fastcrypto-tbls/src/dkg_v1.rs
```diff
@@ -983,3 +998,421 @@ where
...
+        used_messages: &[Message<G, EG>],
+        confirmations: &[Confirmation<EG>],
+    ) -> FastCryptoResult<Output<G, EG>> {
+        let verified_messages = self.process_confirmations(used_messages, confirmations)?;
+        Ok(self.aggregate(&verified_messages))
+    }
+
+    /// Alternative to complete() - Optimistic version that assumes all messages are valid and if
+    /// other parties received invalid shares, the higher level protocol will handle it.
+    pub fn complete_optimistic(
```
benr-ml: no need

fastcrypto-tbls/src/dkg_v1.rs
```diff
@@ -983,3 +998,421 @@ where
...
+        &self,
+        used_messages: &[Message<G, EG>],
+    ) -> FastCryptoResult<Output<G, EG>> {
+        // Do not filter out any messages, assume all are valid.
+        Ok(self.aggregate(used_messages))
+    }
+
+    /// Alternative to complete() - Optimistic variant that interpolates the shares instead of
+    /// summing them, to be used for key rotation. Works only if all weights are 1.
+    pub fn complete_optimistic_key_rotation(
```
benr-ml: no need

jonas-lj: This is because we only need the observer for the DKG and not the key rotation?

benr-ml: right

fastcrypto-tbls/src/dkg_v1.rs
```diff
@@ -983,3 +998,280 @@ where
...
+            g: PhantomData,
+        })
+    }
+
+    /// The threshold needed to reconstruct the full key/signature.
+    pub fn t(&self) -> u16 {
+        self.t
+    }
+
+    /// 5. Process a message and create the second message to be broadcasted.
```
benr-ml: 5?
also numbers below

benr-ml (approved):

PR #940, opened 2026-04-10
New computation for allowed loss of liveness in new_reduce

jonas-lj:


fastcrypto-tbls/src/nodes.rs
benr-ml, line 171: add a TODO about the extra gap, to fix later

benr-ml (approved):

PR #942, opened 2026-04-14
Add hash_to_ristretto255 following RFC 9380

jonas-lj:
## Summary
- Add `hash_to_ristretto255` method to `RistrettoPoint` implementing the `ristretto255_XMD:SHA-512_R255MAP_RO_` suite from RFC 9380 Appendix B
- Uses `expand_message_xmd` with SHA-512 to produce 64 uniform bytes, then maps to a point via the Ristretto Elligator 2 map
- Add test vectors verified against `@noble/curves` for cross-library compatibility
- Refactor to let the VRF implementation use this new function. This is a refactor and doesn't change the behaviour of the implementation.

### DIFF
--- a/fastcrypto/src/tests/ristretto255_tests.rs
+++ b/fastcrypto/src/tests/ristretto255_tests.rs
@@ -6,6 +6,7 @@ use crate::groups::ristretto255::RistrettoScalar;
 use crate::groups::{GroupElement, HashToGroupElement, MultiScalarMul};
 use crate::serde_helpers::ToFromByteArray;
 use hex_literal::hex;
+use crate::encoding::{Encoding, Hex};
 
 pub(crate) const GROUP_ORDER: [u8; 32] = [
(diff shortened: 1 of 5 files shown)


fastcrypto/src/tests/ristretto255_tests.rs
```diff
@@ -232,3 +237,26 @@ fn test_multiscalar_mul() {
     )
     .is_err());
 }
+
+#[test]
+fn test_hash_to_point() {
```
benr-ml, line 242: i'm assuming those tests passed the previous version as well?

benr-ml (approved):
pls make sure that the new regressions tests pass before and after this pr

fastcrypto/src/tests/ristretto255_tests.rs
jonas-lj: No, because `hash_to_ristretto` didn't exist before this PR.

PR #953, opened 2026-04-28
Add prop_reduce weight reduction

arnab-roy:
This PR adds Nodes::prop_reduce to fastcrypto-tbls mirroring new_reduced_with_f. It picks d on a 0.01 grid in [2,40], floors weights, ceilings t and f, and enforces the same budget: remainder mass plus ceiling slack for t and f must be at most allowed_delta. It scans integers down to the first feasible d, then refines (d,d+1) using scaled integer arithmetic only. The enlarged search never worsens reduced total weight and can improve it when slack falls between integer divisors. Embedded Sui mainnet voting‑power snapshot tests on five recent epochs assert prop_reduce is never looser than new_reduced_with_f across representative thresholds and budgets.

### DIFF
--- a/fastcrypto-tbls/src/nodes.rs
+++ b/fastcrypto-tbls/src/nodes.rs
@@ -292,9 +292,243 @@ impl<G: GroupElement + Serialize> Nodes<G> {
             new_f,
         ))
     }
+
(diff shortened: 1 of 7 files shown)


zhouwfang (commented):
Could you update the PR description? It still describes the two-stage API and the $\delta + 2d$ budget.

zhouwfang (approved):

PR #960, opened 2026-05-19
Handle zero-weight nodes in threshold-schnorr

jonas-lj:
Zero-weight nodes (produced by `Nodes::new_reduced`) panicked or were previously rejected in `avss::complete_key_rotation`, `Presignatures::new` and its iterator, and `batch_avss::verify_shares`. They are now treated as observers that hold no shares. Adds a zero-weight test for each.

### DIFF
--- a/fastcrypto-tbls/src/threshold_schnorr/presigning.rs
+++ b/fastcrypto-tbls/src/threshold_schnorr/presigning.rs
@@ -19,18 +19,18 @@ impl Iterator for Presignatures {
     type Item = (Vec<S>, G);
 
     fn next(&mut self) -> Option<Self::Item> {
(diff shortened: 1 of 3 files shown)


fastcrypto-tbls/src/threshold_schnorr/presigning.rs
zhouwfang: This becomes misleading now. Consider .`ok_or(InvalidInput)?` for symmetry with the earlier change on old line 71?

zhouwfang (approved):

fastcrypto-tbls/src/threshold_schnorr/presigning.rs
jonas-lj: Good catch — done in 9d94f7db4.

PR #964, opened 2026-05-20
Doc/code consistency fixes

jonas-lj:
## Summary
Three small doc/code inconsistency fixes found during a docs audit.

1. **`RSAModulus::ensure_in_subgroup`** ([fastcrypto-vdf/src/rsa_group/modulus.rs](fastcrypto-vdf/src/rsa_group/modulus.rs)): the doc said the function panics when `value >= N`, but `value == N` returns 0 silently (precondition violated) — only `value > N` actually panics (via BigUint underflow). Tightened the doc to "greater than" and noted that `value == N` is unspecified.

2. **`Monomial::divider`** ([fastcrypto-tbls/src/polynomial.rs](fastcrypto-tbls/src/polynomial.rs)): the doc claimed the function panics if the dividend's degree is smaller than the divisor's, but the underlying `usize - usize` only panics in debug builds and silently wraps in release. Added an explicit `assert!` so the documented panic fires in both modes.
   - Note: the only caller (`Poly::div_rem`) upholds the precondition via its loop guard, and both `divider` and `Monomial` are module-private, so the assert cannot fire today. It is purely defense-in-depth against future modifications.

3. **`BLS12381PublicKey::from_bytes`** ([fastcrypto/src/bls12381/mod.rs](fastcrypto/src/bls12381/mod.rs)): the comment referenced `key_validate()`, which is not called in the function. Rewrote to state directly that `blst::PublicKey::from_bytes` skips the subgroup membership check and that callers should use `validate()` on untrusted input.

## Test plan
- [x] `cargo test -p fastcrypto-tbls --lib polynomial` (40 passed)
- [x] `cargo test -p fastcrypto-vdf --lib` (29 passed)
- [x] `cargo test -p fastcrypto --lib bls12381` (92 passed)

### DIFF
--- a/fastcrypto-tbls/src/polynomial.rs
+++ b/fastcrypto-tbls/src/polynomial.rs
@@ -566,9 +566,15 @@ impl<C: Scalar> Monomial<C> {
     /// Panics if the degree of `x` is smaller than `self` or if `self` is zero.
     fn divider(self) -> impl Fn(&Monomial<C>) -> Monomial<C> {
         let inverse = self.coefficient.inverse().unwrap();
-        move |p: &Monomial<C>| Monomial {
-            coefficient: p.coefficient * inverse,
-            degree: p.degree - self.degree,
+        move |p: &Monomial<C>| {
+            assert!(
+                p.degree >= self.degree,
+                "Monomial::divider: dividend degree is smaller than divisor"
+            );
+            Monomial {
+                coefficient: p.coefficient * inverse,
+                degree: p.degree - self.degree,
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


fastcrypto-tbls/src/polynomial.rs
```diff
@@ -566,9 +566,15 @@ impl<C: Scalar> Monomial<C> {
     /// Panics if the degree of `x` is smaller than `self` or if `self` is zero.
     fn divider(self) -> impl Fn(&Monomial<C>) -> Monomial<C> {
         let inverse = self.coefficient.inverse().unwrap();
-        move |p: &Monomial<C>| Monomial {
-            coefficient: p.coefficient * inverse,
-            degree: p.degree - self.degree,
+        move |p: &Monomial<C>| {
+            assert!(
```
jonas-lj, line 570: This cannot happen today - there's only one caller (Polynomial::div_rem) who explicitly checks this before calling. It's just to prevent errors in the future where an assert will fire in debug mode but not in release mode.

benr-ml (approved):

PR #966, opened 2026-05-22
Bound input length in split_to_two_frs

jonas-lj:
Reject inputs outside `[33, 47]` bytes in `split_to_two_frs`: shorter inputs would underflow `len() - 16`, and `≥ 48` byte inputs let the first half exceed the BN254 modulus and silently reduce (enabling collisions).

Also adds a note in `get_nonce` about a related latent underflow that is unreachable in practice.

This has **no** impact on Sui which always builds `extended_pk_bytes = flag (1 byte) || public_key_bytes`, producing 33 bytes for ed25519 and 34 bytes for compressed secp256k1 / secp256r1 both of which are inside the new bounds. The `verify_zk_login` path in `sui-types::zk_login_authenticator` and the CLI `keytool` nonce-generation path are both unaffected.


jonas-lj:
Covered in #939. Closing

joyqvq (approved):

PR #967, opened 2026-05-22
Add DST to bulletproofs, DDH and key consistency proofs

jonas-lj:
Adds a domain-separation-tag (`dst`) to the bulletproof range proofs and the Twisted ElGamal / DDH sigma proofs (`DdhTupleNizk`, `ZeroProof`, `ConsistencyProof`, `KeyConsistencyProof`).

The sigma-proof challenges are made byte-for-byte compatible with Contra's Move/TS implementations: `blake2b256(dst || canonical-element-bytes…)`, top byte zeroed, canonical LE scalar (`RistrettoScalar::fiat_shamir_challenge`). Breaking change: prove/verify call sites now require a `dst`.

This changes range proof verification, so Sui must be updated.

## Test plan
- [x] `cargo test -p fastcrypto --lib` (nizk, twisted_elgamal, bulletproofs)
- [x] `cargo fmt --check` + `cargo xclippy -D warnings`

### DIFF
--- a/fastcrypto/src/groups/ristretto255.rs
+++ b/fastcrypto/src/groups/ristretto255.rs
@@ -10,7 +10,7 @@ use crate::groups::{
     Doubling, FiatShamirChallenge, FromTrustedByteArray, GroupElement, HashToGroupElement,
     MultiScalarMul, Scalar,
 };
-use crate::hash::{ReverseWrapper, Sha512};
+use crate::hash::{Blake2b256, ReverseWrapper, Sha512};
 use crate::serde_helpers::ToFromByteArray;
 use crate::traits::AllowedRng;
 use crate::{
@@ -223,7 +223,10 @@ impl HashToGroupElement for RistrettoScalar {
 
 impl FiatShamirChallenge for RistrettoScalar {
     fn fiat_shamir_reduction_to_group_element(msg: &[u8]) -> Self {
-        Self::hash_to_group_element(msg)
+        // Matches Contra's Move/TS Fiat-Shamir construction.
+        let mut digest = Blake2b256::digest(msg).digest;
+        digest[RISTRETTO_SCALAR_BYTE_LENGTH - 1] = 0;
+        Self::from_byte_array(&digest).expect("Top byte is zero so the scalar is always canonical")
     }
 }
 
--- a/fastcrypto/src/nizk.rs
+++ b/fastcrypto/src/nizk.rs
@@ -4,7 +4,6 @@
 use crate::error::{FastCryptoError, FastCryptoResult};
 use crate::groups::MultiScalarMul;
 use crate::groups::{FiatShamirChallenge, GroupElement, Scalar};
-use crate::hash::{HashFunction, Sha3_256};
 use crate::traits::AllowedRng;
 use serde::{Deserialize, Deserializer, Serialize};
 use std::ops::Neg;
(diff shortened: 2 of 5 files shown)


fastcrypto/src/groups/ristretto255.rs
```diff
@@ -136,6 +136,20 @@ impl RistrettoScalar {
     pub fn from_bytes_mod_order(bytes: &[u8; 32]) -> Self {
         RistrettoScalar(ExternalScalar::from_bytes_mod_order(*bytes))
     }
+
+    /// Derive a Fiat-Shamir challenge scalar from `dst` followed by a sequence of byte chunks.
+    pub(crate) fn fiat_shamir_challenge(
+        dst: &[u8],
+        parts: impl IntoIterator<Item = impl AsRef<[u8]>>,
```
benr-ml: is the assumption that those are fixed sizes on the app level?

fastcrypto/src/nizk.rs
```diff
@@ -2,57 +2,112 @@
...
+use crate::groups::{GroupElement, Scalar};
+use crate::serde_helpers::ToFromByteArray;
 use crate::traits::AllowedRng;
 use serde::{Deserialize, Deserializer, Serialize};
+use std::marker::PhantomData;
 use std::ops::Neg;
 
+/// Strategy for deriving the Fiat-Shamir challenge of a DDH-tuple proof from its public points
+/// `(xG, xH, A, B)` and a domain separation tag.
+pub trait FiatShamirChallenge<G: GroupElement> {
```
benr-ml: do we need this abstraction?

fastcrypto/src/twisted_elgamal.rs
```diff
@@ -53,20 +150,36 @@ pub struct VerifiableKeyEncapsulation<const N: usize> {
...
 }
 
-/// Sigma protocol that a Twisted ElGamal ciphertext encrypts the message 0.
+/// Sigma protocol that a Twisted ElGamal ciphertext encrypts the message 0. This is a DDH proof, so
+/// it reuses the [FiatShamirChallenge] reduction strategy `C` with `x_g = pk`,
+/// `x_h = decryption_handle`, `a = a1`, `b = a2`.
 #[derive(Debug, Serialize, Deserialize)]
-pub struct ZeroProof {
+#[serde(bound = "")]
+pub struct ZeroProof<C = Blake2bCanonicalChallenge> {
```
benr-ml: why do we need this generic?

fastcrypto/src/groups/ristretto255.rs
jonas-lj: Yes, I'll add that to the docs.

fastcrypto/src/nizk.rs
jonas-lj: Yes. We need a very specific way of computing the FiatShamirChallenge for Contra which depends on the group structure and encoding, and I don't want to hardcode that here. So there are two options, imo, either get rid of all generics in the proofs (also G: GroupElement) or keep this -- and I prefer the latter.

fastcrypto/src/twisted_elgamal.rs
jonas-lj: Same as above

fastcrypto/src/nizk.rs
jonas-lj: I've changed this s.t. the default way to compute FS challenges on Ristretto is now this. It's only used in the avss/batch_avss and this way we don't have to maintain two ways of computing a challenge.

benr-ml (approved):

fastcrypto/src/nizk.rs
```diff
@@ -65,10 +67,20 @@ where
...
+    /// reduced via the scalar's [FiatShamirChallenge] impl. For Ristretto255 this matches Contra's
+    /// Move/TS construction.
+    fn challenge(x_g: &G, x_h: &G, a: &G, b: &G, dst: &[u8]) -> G::ScalarType {
+        let chunks: Vec<Vec<u8>> = vec![
+            dst.to_vec(),
+            bcs::to_bytes(x_g).expect("Serialization succeeds"),
+            bcs::to_bytes(x_h).expect("Serialization succeeds"),
+            bcs::to_bytes(a).expect("Serialization succeeds"),
+            bcs::to_bytes(b).expect("Serialization succeeds"),
+        ];
```
Daeinar: Nit: Should we pass parameters to `challenge` in the same order as they are processed, i.e., move dst to the front?

fastcrypto/src/twisted_elgamal.rs
```diff
@@ -591,13 +676,18 @@ pub fn precompute_table() -> HashMap<[u8; RISTRETTO_POINT_BYTE_LENGTH], u16> {
...
 }
 
-/// Derive a Fiat-Shamir challenge scalar from any serializable message.
-///
-/// TODO: Add domain seperation
-fn fiat_shamir_challenge<T: Serialize>(msg: &T) -> RistrettoScalar {
-    let mut digest = Blake2b256::digest(bcs::to_bytes(msg).unwrap()).digest;
-    digest[31] = 0;
-    RistrettoScalar::from_byte_array(&digest).expect("Always in field")
+/// Derive an internal batching coefficient labelled `label` from the challenge `c` and (optional)
```
Daeinar: This should not be derived from the challenge `c` but from a fresh random value. I know that this will be fixed in another PR (right?) but I'm wondering if it would make sense to write here already that the batching coefficient will be derived from randomness `r` instead of challenge `c` (i.e. update the name of the input parameter etc)?

Daeinar (approved):
@jonas-lj Looks good overall. Just some smaller nits. Feel free to ignore if they don't make sense in the context of this PR.

fastcrypto/src/nizk.rs
jonas-lj: I think that's already the case. I'll double-check alignment with the Move impl

fastcrypto/src/twisted_elgamal.rs
jonas-lj: I've updated such that all the coefficients are just sampled at random and we don't derive them from anything

PR #968, opened 2026-05-22
chore: fix publish workflow and bump crate versions

joyqvq:
published successfully using this action: https://github.com/MystenLabs/fastcrypto/actions/runs/26288835519

https://crates.io/crates/fastcrypto

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -3,6 +3,18 @@ on:
   push:
     branches:
       - main
+    paths:
+      - 'Cargo.toml'
+      - 'fastcrypto-derive/Cargo.toml'
+      - 'fastcrypto/Cargo.toml'
+      - 'fastcrypto-zkp/Cargo.toml'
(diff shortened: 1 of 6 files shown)


.github/workflows/publish.yml
```diff
@@ -3,6 +3,18 @@ on:
...
       - main
+    paths:
+      - 'Cargo.toml'
+      - 'fastcrypto-derive/Cargo.toml'
+      - 'fastcrypto/Cargo.toml'
+      - 'fastcrypto-zkp/Cargo.toml'
+  workflow_dispatch:
+    inputs:
+      commit_sha:
+        description: Optional explicit commit SHA to tag and publish (defaults to latest main)
```
benr-ml: why do we want this?

joyqvq: this should always be main, revised

jonas-lj (approved):

PR #971, opened 2026-06-09
Add AWS Nitro attestation verification module

mskd12:
## Summary

Moves AWS Nitro enclave attestation verification out of the sui monorepo (`crates/sui-types/src/nitro_attestation.rs`) into the core `fastcrypto` crate, so other consumers can use it without cloning sui. Verification covers COSE_Sign1 (RFC 8152) CBOR parsing, attestation-document validation, and X.509 chain verification against the AWS Nitro root CA with P-384 ECDSA.

The port is intentionally minimal — this is consensus-critical verification code, so the diff against the sui original is kept to mechanical changes only:

- License header gains the year (required by `scripts/license_check.sh`); added a module doc comment.
- Error plumbing decoupled from sui: `SuiError`/`SuiErrorKind`/`SuiResult` removed. The `NitroAttestationVerifyError` enum is retained (now a `thiserror` derive with byte-identical Display strings) and stays public.
- Public entry points return `FastCryptoResult`, matching the rest of the crate; private helpers still produce the typed `NitroAttestationVerifyError`, bridged at the boundary via `From<NitroAttestationVerifyError> for FastCryptoError`.
- Two edition-2024 let-chains rewritten as nested `if`s (fastcrypto is edition 2021).
- `nitro_root_certificate.pem` copied byte-for-byte.

The public API (including the three parsing bool flags `is_upgraded_parsing` / `include_all_nonzero_pcrs` / `always_include_required_pcrs`) is kept verbatim so the sui-side migration is a pure re-export + deletion.

## Test plan

- `cargo test -p fastcrypto nitro_attestation` — 7/7 pass (parse/verify round-trips on real attestations, cert expiry windows, malformed CBOR, mutation tests).
- `cargo xclippy` and `cargo fmt --all --check` — clean.

## Follow-up (separate sui PR, out of scope)

After this merges and sui bumps its fastcrypto pin: replace `crates/sui-types/src/nitro_attestation.rs` with `pub use fastcrypto::nitro_attestation::*;`, drop the orphaned `From<NitroAttestationVerifyError> for SuiError` impl, and move the unit tests over. The move-native callers match `Err(_)` and compile unchanged.


joyqvq (approved):
added more validation and tests. lgtm!

benr-ml:
Just making sure - we didn't change any logic compared to the version in Sui, right? otherwise we need to push that with a protocol config @joyqvq @mskd12

joyqvq:
> Just making sure - we didn't change any logic compared to the version in Sui, right? otherwise we need to push that with a protocol config @joyqvq @mskd12 

i added two additional validations: 1) timestamp should be nonzero 2) pcrs array length is longer than 1. 

either cases they would have been invalid before, so no protocol config is needed. 

what's the intention for this? if the goal is to eventually upstream to use this version to sui, then yes we should consider protocol config.

benr-ml:
> either cases they would have been invalid before, so no protocol config is needed.
> what's the intention for this? if the goal is to eventually upstream to use this version to sui, then yes we should consider protocol config.

yes, we should eventually have only one impl.
we should try not to need protocol config since otherwise we would have to maintain both versions forever.

PR #974, opened 2026-06-17
Harden (batch) avss

jonas-lj:


fastcrypto-tbls/src/dkg_v1.rs
```diff
@@ -397,13 +397,11 @@ where
...
         let my_share_ids = self.nodes.share_ids_of(self.id).expect("my id is valid");
         let encryption_ro = self.encryption_random_oracle(message.sender);
-        let buffer =
-            message
-                .encrypted_shares
-                .decrypt(&self.enc_sk, &encryption_ro, self.id as usize);
-        let decrypted_shares: Option<Vec<G::ScalarType>> = bcs::from_bytes(buffer.as_slice())
-            .map_err(|_| FastCryptoError::InvalidInput)
-            .ok();
+        let decrypted_shares: Option<Vec<G::ScalarType>> = message
```
jonas-lj: This doesn't change anything here. The new failure path (invalid id) cannot happen since that's checked in `sanity_check`.

fastcrypto-tbls/src/ecies_v1.rs
```diff
@@ -178,7 +182,10 @@ where
         receiver_pk: &PublicKey<G>,
         receiver_index: usize,
     ) -> FastCryptoResult<Vec<u8>> {
-        assert!(receiver_index < self.encs.len());
+        let enc = self
```
benr-ml: also here

fastcrypto-tbls/src/dkg_v1.rs
```diff
@@ -397,13 +397,11 @@ where
...
+            .encrypted_shares
+            .decrypt(&self.enc_sk, &encryption_ro, self.id as usize)
+            .ok()
+            .and_then(|buffer| bcs::from_bytes(buffer.as_slice()).ok());
```
benr-ml: we don't check elsewhere the the encrypted blob is correct

fastcrypto-tbls/src/ecies_v1.rs
```diff
@@ -134,14 +134,18 @@ where
         sk: &PrivateKey<G>,
         encryption_random_oracle: &RandomOracle,
         receiver_index: usize,
-    ) -> Vec<u8> {
+    ) -> FastCryptoResult<Vec<u8>> {
+        let enc = self
```
benr-ml: if the assumption is that Verify is called before, why not panic here? the error cannot be handled in that case

fastcrypto-tbls/src/dkg_v1.rs
jonas-lj: No, but in the old version, that'd still end up as decrypted_shares being `None`, same as in the new version.

fastcrypto-tbls/src/ecies_v1.rs
jonas-lj: Fair point. I'll add an assert and a comment to make that explicit

benr-ml (approved):

PR #976, opened 2026-06-17
Bundle UnsignedAvssCert into AvidMessage for the pessimistic phase

jonas-lj:
Bundle the AVID dispersal and the `UnsignedAvssCert` into a single `AvidMessage` so the receiver can enter the pessimistic phase without an out-of-band cert distribution channel — the dealer broadcasts both together.

### DIFF
--- a/fastcrypto-tbls/src/threshold_schnorr/avid.rs
+++ b/fastcrypto-tbls/src/threshold_schnorr/avid.rs
@@ -87,16 +87,17 @@ impl Avid {
         Ok(Self { nodes, coder, f })
     }
 
-    /// 1. Disperse one payload per recipient. Returns one per-party [Dispersal]. Runs `mutate`
-    ///    over the per-recipient, per-disperser shards before the Merkle trees are built — for
-    ///    production callers pass `|_| {}`.
-    ///    Fails if any of the payloads are empty.
+    /// 1. RS-encode every payload, build the two-level Merkle commitment, and return a
+    ///    [DispersalBuilder] that can mint per-disperser [Dispersal]s on demand via
+    ///    [DispersalBuilder::dispersal_for]. Runs `mutate` over the per-recipient, per-disperser
+    ///    shards before the Merkle tree is built — production callers pass `|_| {}`. Fails if any
+    ///    of the payloads are empty.
     #[cfg_attr(not(test), allow(unused_variables, unused_mut))]
     pub fn disperse_with_mutation(
         &self,
         payloads_by_recipient: &BTreeMap<PartyId, Vec<u8>>,
         mutate: impl FnOnce(&mut BTreeMap<PartyId, Vec<Vec<Shard>>>),
-    ) -> FastCryptoResult<BTreeMap<PartyId, Dispersal>> {
+    ) -> FastCryptoResult<DispersalBuilder> {
         // RS-encode each recipient's payload and bucket the shards by disperser.
         let mut shards_by_recipient: BTreeMap<PartyId, Vec<Vec<Shard>>> = payloads_by_recipient
             .iter()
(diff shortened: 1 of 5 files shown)


fastcrypto-tbls/src/threshold_schnorr/avid.rs
```diff
@@ -41,10 +49,11 @@ pub struct AuthenticatedShards {
     pub(crate) proof: NestedMerkleProof,
 }
 
-/// An endorsement of a dispersal's `top_root`.
+/// An endorsement of a dispersal's `top_root` and the set of pending recipients it covers.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Vote {
     pub top_root: merkle::Node,
+    pub pending_recipients: BTreeSet<PartyId>,
```
benr-ml:
```suggestion
    pub recipients: BTreeSet<PartyId>,
```

pending is a property of the avss protocol

fastcrypto-tbls/src/threshold_schnorr/batch_avss_avid.rs
```diff
@@ -388,36 +393,45 @@ impl Dealer {
...
-        if !pending_recipients.is_subset(&all_ids) {
-            return Err(InvalidInput);
-        }
-        // AVID-disperse each pending recipient's existing ciphertext `E_i`, bound to `H(v)`.
+    ) -> FastCryptoResult<AvidMessageBuilder<C>> {
+        let pending_recipients: BTreeSet<PartyId> = self
+            .nodes
+            .node_ids_iter()
+            .filter(|id| !avss_cert.signers().contains(id))
+            .collect();
```
benr-ml, line 431: sanity check that their weight is <= f

fastcrypto-tbls/src/threshold_schnorr/batch_avss_avid.rs
```diff
@@ -511,82 +526,80 @@ impl Receiver {
...
-    ///    [UnsignedAvssCert] from the published certificate from the optimistic phase.
-    ///    Returns also an [AvidVote] to be signed and returned to the dealer.
-    pub fn prepare_avid_echo_messages_and_vote(
+    /// 4. Verify a pessimistic-phase [AvidMessage] and emit an [EchoBuilder] which can build
+    ///    [Echo]es on demand. Returns also an [AvidVote] to be signed and returned to the dealer.
+    ///
+    ///    Receivers who did not receive their shares through an [AvssMessage] should wait for a
+    ///    published [Certificate] over [AvidVote]s that confirms they are a pending recipient and
+    ///    then get [Echo]es from the signers and verify them with [Self::verify_avid_echo_message].
+    pub fn process_avid_message<C: Certificate<Payload = AvssVote>>(
```
benr-ml, line 560: why doen't it receive the VerifiedAvssMessage?

fastcrypto-tbls/src/threshold_schnorr/batch_avss_avid.rs
```diff
@@ -511,82 +526,80 @@ impl Receiver {
...
-    ///    to decode and to answer complaints.
-    pub fn decode_ciphertext(
+    ///    The caller should retain its decoded ciphertext for the session to answer complaints.
+    pub fn decode_ciphertext<C: Certificate<Payload = AvidVote>>(
         &self,
         echoes: &[VerifiedEcho],
         verified_common: &VerifiedAvssCommonMessage,
+        avid_cert: &VerifiedCertificate<C>,
     ) -> FastCryptoResult<DecodeOutcome> {
+        if avid_cert.payload().common_message_hash != verified_common.0.hash() {
```
benr-ml, line 638: why do we need this? if I recevied a different avss message than the rest, and i see a cert, can't use the certified one?

fastcrypto-tbls/src/threshold_schnorr/batch_avss_avid.rs
jonas-lj: That's  part of the AvidMessage

jonas-lj: Oh, you mean the receivers own AvssMessage?

fastcrypto-tbls/src/threshold_schnorr/batch_avss_avid.rs
jonas-lj: No, you need to fetch the right verified_common before calling this. Otherwise, you can't verify the ciphertext hashes.

benr-ml: i think that what confuse me is that VerifiedAvssCommonMessage is used both after my shares were verified and also if not. here it is the second case

benr-ml (approved):

fastcrypto-tbls/src/threshold_schnorr/batch_avss_avid.rs
jonas-lj: Got it. I'll update the docs to make this clear, and perhaps we can come up with a better solution. Fwiw, I don't think adding yet another type makes it less confusing.

PR #984, opened 2026-07-03
Small refactors in threshold_schnorr AVSS/AVID

jonas-lj:
Follow-up cleanups (stacked on #983):

- Extract `Receiver::challenge_for`, `Vote::recipient_index`, and `SharesForNode::share_for_index` to remove duplicated expressions, plus drop a redundant random-oracle rebind / double-ref in `decrypt_and_verify_shares`.
- Simplify `complete_dkg`'s manual first-element fold to a `reduce` over `Result` items (`try_reduce` would be cleaner but is still nightly-only).


jonas-lj:
Squashed into #983.

PR #988, opened 2026-07-14
[tschnorr] docs and small refactors

benr-ml:
- Refactored and cleaned up batch_avss_avid.rs
- Smaller refactor of avis.rs
- docs updates

### DIFF
--- a/fastcrypto-tbls/src/threshold_schnorr/batch_avss_avid.rs
+++ b/fastcrypto-tbls/src/threshold_schnorr/batch_avss_avid.rs
@@ -30,7 +30,6 @@ use crate::threshold_schnorr::batch_avss_avid::DecodeAndDecryptOutcome::{
 };
 use crate::threshold_schnorr::bcs::BCSSerialized;
 use crate::threshold_schnorr::recovery_proof;
-use crate::threshold_schnorr::reed_solomon::ErasureCoder;
 use crate::threshold_schnorr::Extensions::{Challenge, Encryption, Recovery};
 use crate::threshold_schnorr::{avid, Certificate, VerifiedCertificate};
 use crate::threshold_schnorr::{random_oracle_from_sid, Parameters, EG, G, S};
@@ -247,7 +246,6 @@ impl Dealer {
     ) -> FastCryptoResult<Self> {
         let total_weight = nodes.total_weight();
         params.validate(total_weight)?;
-        params.check_erasure_code_params(total_weight)?;
         let nodes = Arc::new(nodes);
         let avid = avid::Avid::new(Arc::clone(&nodes), params.f)?;
         let batch_size = nodes.weight_of(dealer_id)? as usize * batch_size_per_weight as usize;
@@ -406,7 +404,7 @@ impl Dealer {
     }
 
     // Step 6 happens at the caller level:
-    //   6. Once `t+f` weight of [AvidVote]s has been collected, the dealer can form and
+    //   6. Once `W-f` weight of [AvidVote]s has been collected, the dealer can form and
     //      publish a certificate over those votes on the TOB. This is done by the caller and
     //      completes the dealer's role in the protocol.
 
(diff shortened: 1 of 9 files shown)


fastcrypto-tbls/src/threshold_schnorr/presigning.rs
```diff
@@ -165,18 +173,44 @@ mod tests {
 
         let total_weight_of_outputs = 2;
         let expected_len =
-            (total_weight_of_outputs - params.f as usize) * batch_size_per_weight as usize;
+            (total_weight_of_outputs - (params.t as usize - 1)) * batch_size_per_weight as usize;
```
benr-ml, line 182: @zhouwfang this is not a backward compatible change, should we put it behind a flag/parameter?

zhouwfang: We plan to launch on testnet on Thursday, so I'd first try to take this into the launch build. If it misses the launch, we can keep both formulas caller-selectable. WDYT?

fastcrypto-tbls/src/threshold_schnorr/batch_avss_avid.rs
```diff
@@ -382,57 +380,63 @@ impl Dealer {
...
     ) -> FastCryptoResult<AvidMessageBuilder<C>> {
-        self.create_avid_messages_with_mutation(avss_message_builder, avss_cert, |_| {})
+        let payloads = self.prepare_avid_payloads(avss_message_builder, &avss_cert)?;
+        self.avid
+            .disperse(&payloads)
+            .map(|inner| AvidMessageBuilder { inner, avss_cert })
     }
 
+    // Step 6 happens at the caller level:
+    //   6. Once `t+f` weight of [AvidVote]s has been collected, the dealer can form and
```
zhouwfang: avid.rs line 239 says the dealer collects W-f votes; here it says t+f. I'm fairly sure W-f is the intended one: a cert with quorum q only guarantees q-f honest echo holders and decode needs W-2f, so with q = t+f, laggard decode needs t+2f >= W while worst-case cert formation needs t+2f <= W — sound only at exactly t+2f = W. W-f meets both with equality for any valid params (and is what hashi implements). Copy-paste from step 3, where t+f is the right quorum?

fastcrypto-tbls/src/threshold_schnorr/batch_avss_avid.rs
```diff
@@ -1186,6 +1156,7 @@ impl SharesForNode {
 
 impl Parameters {
     fn check_erasure_code_params(&self, total_weight: u16) -> FastCryptoResult<()> {
+        //?
```
zhouwfang: Leftover. Related: this checks (W, t) as RS-code params, but the only coder actually built is the (W, W-2f) one in Avid::new, which validates its own params, so the effect here is just enforcing 0 < t < W. Intentional, or should it check what's actually constructed?

fastcrypto-tbls/src/threshold_schnorr/presigning.rs
```diff
@@ -39,17 +39,25 @@ impl Iterator for Presignatures {
...
+    /// Caller should wait for at least `params.t` outputs, plus some \delta time to collect more.
     ///
-    /// `params.t` is the reconstruction threshold and `params.f` is the Byzantine bound used to size
-    /// the Pascal matrix, which produces `total_weight - params.f` presignatures per nonce position.
+    /// `params.t` is the reconstruction threshold. The nonce polynomials are shared at degree
+    /// `params.t - 1`, so this produces `total_weight - (params.t - 1)` presignatures per nonce
+    /// position: the privacy threshold of the sharings is `t - 1`, meaning a sub-`t` coalition can
+    /// know or bias up to `t - 1` of the input nonces, so only `total_weight - (t - 1)` combined
+    /// nonces per position remain uniformly random and safe to output.
     /// Requires `params.t > params.f`.
```
zhouwfang: f is no longer read anywhere in this file after the height change, and validate() allows t == f (reachable after weight reduction). Drop the line?

fastcrypto-tbls/src/threshold_schnorr/presigning.rs
benr-ml: sure, will add a flag to control it from your side

benr-ml: added a flag

zhouwfang (approved):

PR #989, opened 2026-07-22
BulletProofs++

Daeinar:


arnab-roy:
- To align with BP verifier, we need to add an RNG to the verify parameters.
- BP doesn't have batch size, bit length in DST, while we do

arnab-roy:
LGTM
- Cross-checked with Lean specs
- All tests pass
- TODO: match interface with fastcrypto/bp

arnab-roy (approved):

Daeinar:
Regarding the BP/BP++ API differences, here's a side-by-side comparison created by Claude:

> - The method surfaces are identical in name, argument types, order and return type for prove, prove_batch, to_bytes, from_bytes and Range::is_in_range.
> - The one real difference: BP's verify and verify_batch take an rng, BP++'s don't. That's not an oversight on either side — dalek randomizes its batched verification equation, while BP++ verification is a single deterministic MSM checked against the identity. Matching the signature would mean BP++ accepting a parameter it discards forever, so I'd leave it as is; a caller switching implementations deletes one argument.
> - Batch sizes differ: BP requires a power of two, BP++ accepts any m >= 1. So BP code ports to BP++ unchanged, but not the reverse.
> - Range is duplicated in bulletproofs.rs and bppp/range_proof.rs — same four variants, same is_in_range. Purely cosmetic unless something needs both types in one scope. Unifying it would mean a shared Range plus a pub use in both modules so fastcrypto::bulletproofs::Range keeps resolving for Sui.
> - Derives differ, with BP++ having more: Range is Clone, Copy, Debug in BP++ and has no derives in BP; RangeProof is Clone, Debug in BP++ and just Debug in BP. None of these is needed internally — I checked by removing them all and the crate still builds — they're there so downstream users aren't blocked, per the Rust API guideline on implementing common traits. Adding them to BP would be a non-breaking one-liner each.
> - On error values: a failed prove_batch gives GeneralOpaqueError in BP and InvalidInput in BP++. Both return InvalidProof from verification.

So if we want to align them, we would have to add the rng parameter to `verify_batch` in BP++, which it won't use, though. Maybe it's fine as it is, since switching any project from BP to BP++ would only require deleting that rng argument from the `verify` function calls. Otherwise, the differences in the derives also seem minor. Wdyt? /cc @benr-ml @arnab-roy @jonas-lj

jonas-lj:
> Regarding the BP/BP++ API differences, here's a side-by-side comparison created by Claude:
> 
> > * The method surfaces are identical in name, argument types, order and return type for prove, prove_batch, to_bytes, from_bytes and Range::is_in_range.
> > * The one real difference: BP's verify and verify_batch take an rng, BP++'s don't. That's not an oversight on either side — dalek randomizes its batched verification equation, while BP++ verification is a single deterministic MSM checked against the identity. Matching the signature would mean BP++ accepting a parameter it discards forever, so I'd leave it as is; a caller switching implementations deletes one argument.
> > * Batch sizes differ: BP requires a power of two, BP++ accepts any m >= 1. So BP code ports to BP++ unchanged, but not the reverse.
> > * Range is duplicated in bulletproofs.rs and bppp/range_proof.rs — same four variants, same is_in_range. Purely cosmetic unless something needs both types in one scope. Unifying it would mean a shared Range plus a pub use in both modules so fastcrypto::bulletproofs::Range keeps resolving for Sui.
> > * Derives differ, with BP++ having more: Range is Clone, Copy, Debug in BP++ and has no derives in BP; RangeProof is Clone, Debug in BP++ and just Debug in BP. None of these is needed internally — I checked by removing them all and the crate still builds — they're there so downstream users aren't blocked, per the Rust API guideline on implementing common traits. Adding them to BP would be a non-breaking one-liner each.
> > * On error values: a failed prove_batch gives GeneralOpaqueError in BP and InvalidInput in BP++. Both return InvalidProof from verification.
> 
> So if we want to align them, we would have to add the rng parameter to `verify_batch` in BP++, which it won't use, though. Maybe it's fine as it is, since switching any project from BP to BP++ would only require deleting that rng argument from the `verify` function calls. Otherwise, the differences in the derives also seem minor. Wdyt? /cc @benr-ml @arnab-roy @jonas-lj

Yea, definitely don't add the rng just to align the API's. The derives doesn't matter either. Re the error types, I don't mind if the impl's are not aligned, but we should check that we use meaningful errors everywhere. It seems like GeneralOpaqueError is a good error for proving to make sure that we don't leak any sensitive information in logs, but I guess it depends on in what cases proving fails?

Daeinar:
@jonas-lj Thanks, that's a good suggestion. I wasn't aware of the `GeneralOpaqueError`. I updated the BP++ prover code to also use that error type.

Daeinar:
@jonas-lj The PR is ready for your review.

fastcrypto/src/lib.rs
```diff
@@ -116,6 +116,8 @@ pub mod utils;
 pub mod vrf;
 
 #[cfg(feature = "experimental")]
+pub mod bppp;
```
jonas-lj: maybe rename to "bulletproofspp" to be aligned with the existing impl of bulletproofs (if we want to keep that around)

fastcrypto/src/bulletproofspp/circuit.rs
```diff
@@ -0,0 +1,1373 @@
...
+            r_l_nonzero: vec![],
+        }
+    }
+
+    /// The honest prover with the witness-derivation steps replaced by the
+    /// caller's choices. The `r_S` solve is left intact, so the cheater still
+    /// cancels every blinded error row it is able to; a rejection therefore
+    /// comes from the circuit constraints, which are aggregated into the
+    /// `T^3` row that no `r_S` slot can reach.
+    fn prove_cheating(
```
jonas-lj, line 856: This is doing (almost) the same as prove_with_witness, no? Try to refactor to avoid too much duplication

fastcrypto/src/bulletproofspp/circuit.rs
```diff
@@ -0,0 +1,1373 @@
...
+
+    /// A single-value prover for a forged commitment
+    /// `V' = v*G + s*H_0 + junk`: it runs the honest algorithm but opens the
+    /// junk coordinate of `hat_V = 2V'` honestly at `T^3`, and cancels every
+    /// blinded error row with `r_S` as usual. Against the unconstrained
+    /// protocol this strategy produces accepting proofs; the shape blocks
+    /// pair the junk into the value row, which no `r_S` slot can reach, so
+    /// verification must fail. With `junk = None` it reduces to the honest
+    /// prover (the control case validating this reimplementation).
+    fn prove_forged(
```
jonas-lj: Same as https://github.com/MystenLabs/fastcrypto/pull/989/changes#r3821698078

fastcrypto/src/bppp/range_proof.rs
```diff
@@ -0,0 +1,609 @@
...
+        // Sanity bound: `rounds` grows with log2 of the norm length, so even
+        // absurdly large statements stay far below 32 rounds.
+        const MAX_ROUNDS: usize = 32;
+        let (rounds, l_len, n_len) = match elems {
+            13 => (3, 1, 2),
+            e if e >= 15 && !e.is_multiple_of(2) && (e - 9) / 2 <= MAX_ROUNDS => {
+                ((e - 9) / 2, 1, 4)
+            }
+            _ => return Err(FastCryptoError::InvalidInput),
+        };
```
jonas-lj: I think this reimplements the norm_linear::proof_shape. Maybe something like:
```suggestion
let (rounds, l_len, n_len) = (4..=MAX_LOG_NM)
    .map(|k| norm_linear::proof_shape(H_LEN, 1 << k))
    .find(|(r, l, n)| 4 + 2 * r + l + n == elems)
    .ok_or(FastCryptoError::InvalidInput)?;
```

jonas-lj: I think we need to implement `to_bytes` and `from_bytes` functions for NormLinearProof and CircuitProof and then use that in this to_bytes function.

fastcrypto/src/bulletproofspp/crs.rs
```diff
@@ -0,0 +1,187 @@
...
+/// only dimension a CRS depends on. The generator derivation (one
+/// hash-to-curve per point) and the MSM table build are paid once per `nm`;
+/// a hit shares the entry rather than copying it.
+static CRS_CACHE: Lazy<RwLock<HashMap<usize, Arc<Generators>>>> =
+    Lazy::new(|| RwLock::new(HashMap::new()));
+
+impl Generators {
+    /// Create the CRS for proofs over `k` values of `n_bits` bits each.
+    /// `n_bits` must be a positive multiple of 4 (at most 64) and `k >= 1`.
+    pub(crate) fn new(n_bits: usize, k: usize) -> FastCryptoResult<Arc<Self>> {
```
jonas-lj: Worth adding a note that this is pretty expensive to compute s.t. we remember not to do this computation on untrusted inputs.

fastcrypto/src/bppp/circuit.rs
```diff
@@ -0,0 +1,1373 @@
...
+        Ok(CircuitParams {
+            n_bits,
+            m,
+            d,
+            n_d,
+            nm,
+        })
+    }
+
+    fn value_in_range(&self, value: u64) -> bool {
```
jonas-lj: I think this is the same as Range::is_in_range?

fastcrypto/src/bppp/norm_linear.rs
```diff
@@ -0,0 +1,669 @@
...
+    }
+    for s in &proof.n_final {
+        transcript.append_scalar(b"n_final", s);
+    }
+
+    let sigma = inner_product(&c, &proof.l_final) + weighted_norm(&proof.n_final, mu);
+
+    // Challenge-product weights via tensor tables (all 2^k products in
+    // 2^{k+1} muls): w[t] = prod_r (set_r if bit_r(t) else unset_r).
+    let tensor_table = |set: &[RistrettoScalar], unset: &[RistrettoScalar]| {
```
jonas-lj: This is also computed in line 246. Perhaps worth a function.

fastcrypto/src/bulletproofspp/norm_linear.rs
```diff
@@ -0,0 +1,669 @@
...
+        dynamic_points.push(*x_point);
+        dynamic_scalars.push(one() - *gamma * *gamma);
+        dynamic_points.push(*r_point);
+    }
+
+    // The precomputed path always uses Straus, which loses to Pippenger for
+    // large MSMs; above the measured crossover (~440 static points for this
+    // workload shape, calibrated by `benches/mixed_msm.rs`) fall back to one
+    // plain MSM.
+    let result = if gens.precomp.num_static_points() <= 440 {
```
jonas-lj: I think this should be done an abstraction up. Perhaps in the mixed_msm impl. for Ristretto?

Daeinar: Yes, good point. I'll check how to best integrate this and send a separate PR.

fastcrypto/src/bulletproofspp/circuit.rs
```diff
@@ -0,0 +1,1255 @@
...
+    let &CircuitParams { m, d, n_d, nm, .. } = params;
+    let mu_inv = mu.inverse()?;
+    // Only two families of lambda-powers are ever read: the per-value weights
+    // `lambda^0..lambda^M`, and the shape-row powers `lambda^{M*j+1}` for
+    // `j = 1..=nm+7`. The dense vector up to `lambda^{M(nm+8)}` that spans
+    // both is quadratic in the batch size and ~99% unread, so build the two
+    // families directly: `shape[j-1] = lambda^{M*j+1}`, stepping by lambda^M.
+    let lambdas = power_vector(lambda, m + 1);
+    let lam_m = lambdas[m];
+    let mut shape = Vec::with_capacity(nm + 7);
```
jonas-lj: The 7 here is related to the H_LEN const, right? If so, use that instead. This is also the case a few more places below.

fastcrypto/src/bulletproofspp/crs.rs
```diff
@@ -0,0 +1,233 @@
...
+/// generator family: the RFC 9380 suite ID plus an application suffix, as in
+/// `fastcrypto::vrf`. The hash (SHA-512) is fixed by fastcrypto's
+/// `hash_to_ristretto255_with_dst` suite.
+// TODO: discuss DST strings.
+const DST_H: &[u8] = b"ristretto255_XMD:SHA-512_R255MAP_RO_fastcrypto-bppp-gen-h-01";
+const DST_G: &[u8] = b"ristretto255_XMD:SHA-512_R255MAP_RO_fastcrypto-bppp-gen-g-01";
+
+/// The provable ranges.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum Range {
```
jonas-lj: This is also defined in the bulletproofs module. Perhaps define it in the pedersen module and use it both places?

fastcrypto/src/bulletproofspp/circuit.rs
```diff
@@ -0,0 +1,1255 @@
...
+    r_v[1] = two
+        * blindings
+            .iter()
+            .enumerate()
+            .fold(S::zero(), |acc, (i, s)| acc + blocks.lambdas[i] * s);
+    let r = Blindings { r_o, r_l, r_r, r_v };
+
+    // Vector coefficients of n(T) at powers T^{-1}..T^3 (the honest
+    // n_hat_V = 0, so T^3 carries only the public block).
+    let n_poly: [Vec<S>; 5] = [
```
jonas-lj: Please double check my math, but if you define

```
w = [n_s, delta*n_o, n_l, n_r, 0]
p = [0, cn_v, cn_r, cn_l, delta_inv*cn_o]
```
then you can define `n_poly = w + p, n_plus_pn = w + 2p, w_weighted = w ∘ bar_mu`. I think that reads a bit simpler.

fastcrypto/src/bulletproofspp/circuit.rs
```diff
@@ -0,0 +1,1255 @@
...
+    let v_commitments: Vec<RistrettoPoint> = values
+        .iter()
+        .zip(blindings)
+        .map(|(&v, s)| {
+            RistrettoPoint::multi_scalar_mul(&[S::from(v), *s], &[gens.g, gens.h_vec[0]])
+        })
+        .collect::<FastCryptoResult<_>>()?;
+
+    transcript.domain_sep(b"bppp_circuit");
+    for v in &v_commitments {
```
jonas-lj: Maybe add an `append_points` helper?

fastcrypto/src/bulletproofspp/circuit.rs
```diff
@@ -0,0 +1,1255 @@
...
+    let n_l = pad_to(
+        &digits.iter().map(|&d| S::from(d)).collect::<Vec<_>>(),
+        params.nm,
+    );
+    let mut n_o: Vec<S> = multiplicities(&digits)
+        .iter()
+        .map(|&m| S::from(m))
+        .collect();
+    n_o.resize(params.nm, S::zero());
+    let r_o = blinding_vector(rng, &[4, 7]);
```
jonas-lj, line 461: These indices read a bit arbitrary (same with the ones for `r_l` and `r_r` below). Could we name them or justify them somehow?

fastcrypto/src/bulletproofspp/circuit.rs
```diff
@@ -0,0 +1,1255 @@
...
+        n_poly[0].clone(),
+        vec_add(&n_poly[1], &blocks.cn_v),
+        vec_add(&n_poly[2], &blocks.cn_r),
+        vec_add(&n_poly[3], &blocks.cn_l),
+        vec_scalar_mul(two, &n_poly[4]),
+    ];
+    let mut fh = [S::zero(); 9];
+    fh[3 + 2] = v_hat + two * (blocks.lambda_al + blocks.mu_am);
+    for (i, w) in w_weighted.iter().enumerate() {
+        for (j, s) in n_plus_pn.iter().enumerate() {
```
jonas-lj:
```suggestion
    for (i, w_i) in w_weighted.iter().enumerate() {
        for (j, s_j) in n_plus_pn.iter().enumerate() {
```

fastcrypto/src/bulletproofspp/circuit.rs
```diff
@@ -0,0 +1,1255 @@
...
+    // with hat_V = 2*sum_i lambda^{i-1} V_i.
+    let two_t3 = S::from(2u64) * t3;
+    let mut extra = vec![
+        (ps_tau, gens.g),
+        (tau_inv, proof.c_s),
+        (delta, proof.c_o),
+        (tau, proof.c_l),
+        (t2, proof.c_r),
+    ];
+    for (i, v) in v_commitments.iter().enumerate() {
```
jonas-lj:
```suggestion
    for (i, v_i) in v_commitments.iter().enumerate() {
```

fastcrypto/src/bulletproofspp/norm_linear.rs
```diff
@@ -0,0 +1,708 @@
...
+    n: &[RistrettoScalar],
+) -> FastCryptoResult<NormLinearProof> {
+    debug_assert_eq!(l.len(), c.len());
+    debug_assert_eq!(l.len(), gens.h_vec.len());
+    debug_assert_eq!(n.len(), gens.g_vec.len());
+    let two = RistrettoScalar::from(2u64);
+    let mut l = l.to_vec();
+    let mut n = n.to_vec();
+    let mut c = c.to_vec();
+    let mut base_h = gens.h_vec.clone();
```
jonas-lj: nit: You don't need to clone here in case it doesn't grow. I've made the suggested changes on this branch: https://github.com/MystenLabs/fastcrypto/compare/philipp/bppp...jonas/bppp-norm-linear-cow?expand=1

fastcrypto/src/bulletproofspp/norm_linear.rs
```diff
@@ -0,0 +1,708 @@
...
+
+        // Fold the scalar vectors: l' = l0 + gamma*l1,
+        // n' = rho^{-1}*n0 + gamma*n1, c' = c0 + gamma*c1.
+        l = vec_add(&l0, &vec_scalar_mul(gamma, &l1));
+        n = vec_add(&vec_scalar_mul(rho_inv, &n0), &vec_scalar_mul(gamma, &n1));
+        c = vec_add(&c0, &vec_scalar_mul(gamma, &c1));
+
+        // Grow the implicit fold tensors by one level:
+        // H' = H0 + gamma*H1, G' = rho*G0 + gamma*G1.
+        w_h = tensor_grow(&w_h, one(), gamma);
```
jonas-lj, line 299: These can be moved inside the loop. See suggestion on this branch: https://github.com/MystenLabs/fastcrypto/compare/philipp/bppp...jonas/bppp-norm-linear-cow?expand=1

fastcrypto/src/bulletproofspp/circuit.rs
```diff
@@ -0,0 +1,1255 @@
...
+    let c_tau = blocks.c_at(tau, tau_inv, beta);
+
+    // Combined commitment, decomposed so it joins the norm-linear verifier's
+    // single MSM (g_vec is touched only there):
+    //   C(tau) = <p_n(tau), G_vec> + p_s(tau)*G
+    //          + tau^{-1}*C_S + delta*C_O + tau*C_L + tau^2*C_R + tau^3*hat_V
+    // with hat_V = 2*sum_i lambda^{i-1} V_i.
+    let two_t3 = S::from(2u64) * t3;
+    let mut extra = vec![
+        (ps_tau, gens.g),
```
jonas-lj: nit: G is already included in the MSM in norm_linear::verify with a different coefficient, so we can get rid of a term.

fastcrypto/src/bulletproofspp/circuit.rs
Daeinar: I'll add some additional explanation on the doc of `blinding_vector`.

fastcrypto/src/bulletproofspp/circuit.rs
```diff
@@ -0,0 +1,1243 @@
...
+        (0..self.bar_mu.len())
+            .map(|k| self.cn_v[k] + tau * self.cn_r[k] + t2 * self.cn_l[k] + t3di * self.cn_o[k])
+            .collect()
+    }
+
+    /// The linear constraint vector `c(tau) = hat_c_r(tau) + cl_v`:
+    /// beta-weighted tau-monomials plus the constant H-side shape block
+    /// (slot 8 carries only the shape entry).
+    fn c_at(&self, tau: S, tau_inv: S, beta: S) -> Vec<S> {
+        let tau_pows = power_vector(tau, 8);
```
jonas-lj: Use a constant here. I think something like this?
```suggestion
        let tau_pows = power_vector(tau, CR_POWERS[H_LEN - 2] as usize + 1);
```

fastcrypto/src/pedersen.rs
```diff
@@ -73,6 +73,55 @@ impl Blinding {
     }
 }
 
+/// The ranges a committed value can be proven to lie in, shared by
+/// [`crate::bulletproofs`] and `crate::bulletproofspp`.
```
jonas-lj: Don't mention the call sites here - it should be self-contained.

jonas-lj (approved):

fastcrypto/src/bulletproofspp/circuit.rs
Daeinar: Yeah, we need `max power + 1` from the `CR_POWERS` array. I'll adapt it to that.

PR #991, opened 2026-07-26
Knapsack weight reduction

benr-ml:
Total reduced weight W' on real Sui voting-power snapshots (δ_allowed = 800, W = 10000):

| epoch |    t |    f | new_reduced_with_f | prop_reduce | knapsack_reduce | improvement |
|------:|-----:|-----:|-------------------:|------------:|----------------:|------------:|
|   800 | 3400 | 3300 |                661 |         661 |             436 |        -34% |
|   800 | 5200 | 2000 |                661 |         661 |             426 |        -36% |
|   974 | 3400 | 3300 |                780 |         729 |             470 |        -36% |
|   974 | 5200 | 2000 |                780 |         729 |             443 |        -39% |
|  1000 | 3400 | 3300 |                780 |         740 |             452 |        -39% |
|  1000 | 5200 | 2000 |                780 |         740 |             437 |        -41% |
|  1100 | 3400 | 3300 |                713 |         713 |             440 |        -38% |
|  1100 | 5200 | 2000 |                713 |         713 |             432 |        -39% |
|  1200 | 3400 | 3300 |                658 |         658 |             470 |        -29% |
|  1200 | 5200 | 2000 |                658 |         658 |             421 |        -36% |

Improvement is knapsack_reduce vs prop_reduce; average -37%.

Alg description and analysis in the code.


fastcrypto-tbls/src/knapsack_weight_reduction.rs
```diff
@@ -0,0 +1,322 @@
...
+
+    // tp <= t and fp <= f, so both fit u16.
+    Ok((tp as u16, fp as u16))
+}
+
+/// Verifier for a candidate reduction: checks conditions (P), (B), (L1), and (L2)
+/// for an arbitrary `(weights', t', f')` in `O(n * W')` time. Returns
+/// `InvalidInput` for malformed inputs and a `GeneralError` naming the first
+/// violated condition otherwise.
+pub(crate) fn verify_reduction(
```
zhouwfang, line 333: Unlike line 94 this never validates f < t, and it doesn't check t' > f' either. So for any candidate not produced by sweep, the property at line 27 is neither guaranteed nor verified. Also it's reachable only from the debug_assert! at nodes.rs:468, so the "defense in depth" compiles out of every release build.

fastcrypto-tbls/src/knapsack_weight_reduction.rs
```diff
@@ -0,0 +1,322 @@
...
+//!
+//! - (P) Privacy: `w(S) < t  =>  w'(S) < t'`.
+//! - (B) Byzantine bound: `w(S) <= f  =>  w'(S) <= f'`.
+//! - (L1) Liveness: `w(S) >= t + delta  =>  w'(S) >= t'`. Sets that could
+//!   reconstruct originally (with `delta` slack) still reach the reduced
+//!   threshold.
+//! - (L2) Liveness: `w(S) >= t + f + delta  =>  w'(S) >= t' + f'`. Sets that
+//!   could form a Byzantine-tolerant quorum originally still can.
+//!
+//! The reduced-space feasibility conditions hold by construction:
```
zhouwfang: nit: t' + 2f' <= W' doesn't "hold by construction", no? The proof below needs W >= t + 2f + delta, which nothing enforces.

fastcrypto-tbls/src/knapsack_weight_reduction.rs
```diff
@@ -0,0 +1,322 @@
...
+    if weights.is_empty() || weights.len() > MAX_PARTIES || weights.iter().any(|&w| w > MAX_WEIGHT)
+    {
+        return Err(FastCryptoError::InvalidInput);
+    }
+    let total_weight = weights.iter().map(|&w| w as u32).sum::<u32>();
+    if total_weight == 0 || total_weight > u16::MAX as u32 {
+        return Err(FastCryptoError::InvalidInput);
+    }
+    let w_total = total_weight as u16;
+    if t == 0 || t > w_total || f >= t || (t as u32) + (f as u32) + (delta as u32) > total_weight {
```
zhouwfang: This is much stricter than prop_reduce, which validates none of it. With a test weight divisor our committee lands at W=100 in e2e tests, t=f=34, and prop_reduce accepts that today — it returns the identity there since our lower bound equals total weight, but it doesn't reject.

benr-ml: sure, but it's the right restriction, why not use f=33 in tests?

fastcrypto-tbls/src/knapsack_weight_reduction.rs
benr-ml: added

fastcrypto-tbls/src/knapsack_weight_reduction.rs
benr-ml: added checks

fastcrypto-tbls/src/knapsack_weight_reduction.rs
```diff
@@ -0,0 +1,331 @@
...
+
+    // f' is the smallest bound satisfying (B): w(S) <= f => w'(S) <= f'.
+    let fp = max_reduced_weight(&min_original_weight_map, f as u32);
+
+    // Liveness (L1): w(S) >= t + delta => w'(S) >= t'.
+    // Checked via the complement T of the worst such S:
+    //   min { w'(S) : w(S) >= t + delta } = W' - max { w'(T) : w(T) <= W - t - delta }.
+    let b1 = (w_total as u32) - (t as u32 + delta as u32);
+    if reduced_total - max_reduced_weight(&min_original_weight_map, b1) < tp {
+        return Err(FastCryptoError::GeneralError("L1 violated".to_string()));
```
jonas-lj: nit: This allocates a string on the heap every time we check a candidate. Perhaps just return an Option from this function instead?

fastcrypto-tbls/src/knapsack_weight_reduction.rs
```diff
@@ -0,0 +1,331 @@
...
+    }
+
+    // Liveness (L2): w(S) >= t + f + delta => w'(S) >= t' + f'.
+    // Checked via the complement T of the worst such S:
+    //   min { w'(S) : w(S) >= t + f + delta } = W' - max { w'(T) : w(T) <= W - t - f - delta }.
+    let b2 = (w_total as u32) - (t as u32 + f as u32 + delta as u32);
+    if reduced_total - max_reduced_weight(&min_original_weight_map, b2) < tp + fp {
+        return Err(FastCryptoError::GeneralError("L2 violated".to_string()));
```
jonas-lj: Same as above

fastcrypto-tbls/src/knapsack_weight_reduction.rs
```diff
@@ -0,0 +1,331 @@
...
+        || w_total > u16::MAX as u32
+        || reduced_total > u16::MAX as u32
+        || reduction.t == 0
+    {
+        return Err(FastCryptoError::InvalidInput);
+    }
+    // Same sanity checks as reduce_weights, in both spaces.
+    if f >= t
+        || reduction.f >= reduction.t
+        || (t as u32) + 2 * (f as u32) + (delta as u32) > w_total
```
jonas-lj: Don't we need to check this for the reduced weights too? That's needed in the AVID protocol.

fastcrypto-tbls/src/knapsack_weight_reduction.rs
zhouwfang: Prod is the bigger question. W is at most 10000, and with t/f at 3334/3333bp that's the only total that passes at all: at W=9966 t+2f is 9967, overflowing by 1 even at delta=0. So it'd need f_bps lowered rather than delta. Is that intended?

For e2e tests (also a note to myself), f=33 clears the gate at W=100 if delta goes to 0 as well; 800 alone blows the budget there, which is 2f + delta <= 66. But some of our e2e configs land at W=99, where t=34 and f is already 33, and 34 + 66 = 100 > 99, so those fail whatever we do.

benr-ml: > Prod is the bigger question. W is at most 10000, and with t/f at 3334/3333bp that's the only total that passes at all: at W=9966 t+2f is 9967, overflowing by 1 even at delta=0. So it'd need f_bps lowered rather than delta. Is that intended?

not sure i follow - given W we define t as a percentage + 1, e.g., W/3 + 1, and f as (W-t)/2. are we using fixed weights today?

fastcrypto-tbls/src/knapsack_weight_reduction.rs
benr-ml: ack - it's negligible compared to knapsack_min_original

fastcrypto-tbls/src/knapsack_weight_reduction.rs
benr-ml: added

fastcrypto-tbls/src/knapsack_weight_reduction.rs
zhouwfang: Yes, t and f are independent on-chain, not f derived from t. W moves every epoch. At W=10000 that lands on exactly your numbers; it only diverges on rounding, which is where the overflow comes from: at W=9966 your (W−t)/2 gives f=3321, our ceil gives 3322.

But your formula wouldn't help much either: it leaves at most 1 for delta. I swept W=1..10000 and t + 2f is W or W−1 every time, so delta=800 fails regardless. That's really what I was asking: is delta meant to be charged on top of a budget that t \approx W/3, f \approx (W−t)/2 already spends?

fastcrypto-tbls/src/knapsack_weight_reduction.rs
```diff
@@ -0,0 +1,332 @@
...
+    };
+
+    while d_candidate > 100 {
+        let reduced = weights
+            .iter()
+            .map(|&w| reduce_weight(w, d_candidate, rounding))
+            .collect::<Vec<_>>();
+        let reduced_total = reduced.iter().map(|&w| w as u32).sum::<u32>();
+        if reduced_total >= lower_bound as u32 {
+            if let Ok((tp, fp)) = check_candidate(weights, w_total, t, f, delta, &reduced) {
```
zhouwfang, line 164: One perf idea. The sweep spends nearly all its time running the DP on candidates it then rejects. Restating L1 as `g(t-1) + g(W-t-delta) <= W' - 1` and L2 as `g(t-1) + g(f) + g(W-t-f-delta) <= W' - 1`, any lower bound on `g` gives a sound reject test, so a greedy first-fit by density rejects in `O(nlogn)` per candidate instead of `O(n*W')`.

I prototyped it: outputs identical to yours across ~140k random instances and all 8 epoch files, ~5–8x on the delta=800 configs, 15–55x at delta=0 where the sweep runs to the end, and ~350x on an n=1000 input that takes 11s today. It's a wash when there's nothing to reject (e.g. the first candidate is already feasible), measured 0.92–1.19x, so within noise.

WDYT? Happy to put up a PR.

benr-ml: looks good! added

fastcrypto-tbls/src/knapsack_weight_reduction.rs
```diff
@@ -0,0 +1,396 @@
...
+//! - (B) Byzantine bound: `w(S) <= f  =>  w'(S) <= f'`.
+//! - (L1) Liveness: `w(S) >= t + delta  =>  w'(S) >= t'`. Sets that could
+//!   reconstruct originally (with `delta` slack) still reach the reduced
+//!   threshold.
+//! - (L2) Liveness: `w(S) >= t + f + delta  =>  w'(S) >= t' + f'`. Sets that
+//!   could form a Byzantine-tolerant quorum originally still can.
+//!
+//! The reduced-space feasibility conditions follow:
+//! - `t > f` => `t' > f'` - follows from monotonicity of `g`.
+//! - `W >= t + 2f + delta` => `t' + 2f' <= W'` - let T be a set with `w(T) <= f`
```
benr-ml: this is the "Operational liveness" from the draft, but it's a bit messy. i want to think if we can just replace delta in most constraints and only constraint f based on W, t, delta

fastcrypto-tbls/src/knapsack_weight_reduction.rs
benr-ml: discussed in slack, f should be floor((W-t)/2

fastcrypto-tbls/src/knapsack_weight_reduction.rs
benr-ml: decided to keep it to balance between recovery (f) and liveness (f-delta)

fastcrypto-tbls/src/tests/nodes_tests.rs
```diff
@@ -393,3 +405,42 @@ fn test_prop_reduce_vs_new_reduced_with_f() {
...
 }
+
+#[test]
+fn test_knapsack_reduce() {
+    let node_vec = get_nodes::<RistrettoPoint>(100);
+    let nodes = Nodes::new(node_vec.clone()).unwrap();
+    let w = nodes.total_weight();
+    let t = w / 3;
+    let delta = w / 10;
+    // The largest f satisfying the precondition t + 2f + delta <= W.
```
zhouwfang: nit: that precondition is gone; it's t + 2f <= W and t + f + delta <= W separately now. Largest f here is actually (w-t)/2, so this doesn't exercise the split.

fastcrypto-tbls/src/knapsack_weight_reduction.rs
```diff
@@ -0,0 +1,398 @@
...
+    //      (L1) => g(t-1) + g(f) + g(W-t-f-delta) <= W' - 1.
+    // 3. g_lower(b) <= g(b) by construction.
+
+    // Lower bound on (L3): g_lower(t-1) + g_lower(W-t-delta) <= W' - 1.
+    if g_lower((t - 1) as u32) + g_lower(w_total as u32 - t as u32 - delta as u32) >= reduced_total
+    {
+        return true;
+    }
+    // Lower bound on (L1): g_lower(t-1) + g_lower(f) + g_lower(W-t-f-delta) <= W' - 1.
+    g_lower((t - 1) as u32)
```
zhouwfang, line 280: nit: g_lower(t-1) again, already computed in the (L3) check just above.

zhouwfang (approved):

fastcrypto-tbls/src/knapsack_weight_reduction.rs
zhouwfang: The t' + 2f' <= W' check went back out in the "update" commit, with the W >= t + 2f + delta assumption behind it.

It does fail with the new hashi params: t' + 2f' > W' on all eight epoch files, 529 vs 488 at epoch 1200.

Fine for hashi: we get uniqueness from the chain, not the threshold. The cert table is keyed by dealer, first write wins. So it's a consumer obligation now, agree?

PR #999, opened 2026-08-03
[elgamal] add verify and verify_opening

joyqvq:


fastcrypto/src/twisted_elgamal.rs
```diff
@@ -74,6 +74,26 @@ pub fn generate_keypair(rng: &mut impl AllowedRng) -> (PublicKey, PrivateKey) {
     (PublicKey::from(&sk), sk)
 }
 
+impl PublicKey {
```
benr-ml, line 77: do we need those functions?
if we do, can we at least remove as_scalar? returning the secret key as ref is risky

benr-ml (approved):

fastcrypto/src/twisted_elgamal.rs
joyqvq: ah yes as_scalar is not needed, removed

fastcrypto/src/twisted_elgamal.rs
```diff
@@ -218,21 +261,6 @@ impl ConsistencyProof {
 }
 
 impl MultiRecipientCiphertext {
-    pub fn new(commitment: PedersenCommitment, decryption_handles: Vec<RistrettoPoint>) -> Self {
```
joyqvq, line 221: not needed for guardian too bc blinding is different

PR #1001, opened 2026-08-06
[feat] Add fastcrypto-pq: mldsa65 trait layer over mysten-mldsa-native-rs

mahdi-mysten:
New workspace crate `fastcrypto-pq`: the fastcrypto trait layer over [mysten-mldsa-native-rs](https://github.com/MystenLabs/mysten-mldsa-native-rs) (ML-DSA-65, FIPS 204).
- Private key is the 32-byte FIPS 204 seed, the only serialized form. Public keys are re-derived on every deserialization.
- The `native` feature forwards to the wrapper's formally verified assembly backends (NEON / AVX2 behind a runtime CPU probe): ~2.5x faster, identical bytes. Measured on an M2 Max: verify 24 micro sec vs Ed25519's 28 micro sec.
- The wrapper dependency is an exact-commit pin, rev ea1a9794, v0.1.0 release tag.
- 19 tests: trait mechanics, serialization pinned to raw bytes, and an interop gate, five vectors from @noble/post-quantum must reproduce byte for byte through this crate and verify under it.

Review focus: `fastcrypto-pq/src/mldsa65/mod.rs` (the seed / keypair contract) and the interop test in `mldsa65_tests.rs`.


fastcrypto-pq/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+bincode.workspace = true
+
+[features]
+# Forwards to the wrapper's verified assembly backends: NEON on aarch64, AVX2 on x86_64
+# behind a runtime CPU probe. Identical signatures either way, roughly 2.5x faster.
+native = ["mysten-mldsa-native-rs/native"]
+
+[dev-dependencies]
+bcs.workspace = true
+bincode.workspace = true
```
jonas-lj: Redundant since this is already in `dependencies`. Same for rand.

fastcrypto-pq/src/mldsa65/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+        MLDSA65KeyPair {
+            public: MLDSA65PublicKey(verifying_key),
+            private: MLDSA65PrivateKey { seed, signing_key },
+        }
+    }
+}
+
+impl From<MLDSA65PrivateKey> for MLDSA65KeyPair {
+    fn from(private: MLDSA65PrivateKey) -> Self {
+        let public = MLDSA65PublicKey::from(&private);
```
jonas-lj, line 143: This computes the expansion again, even though the signing key is cached on the PrivateKey

fastcrypto-pq/src/mldsa65/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+
+impl_base64_display_fmt!(MLDSA65PublicKey);
+
+impl Debug for MLDSA65PublicKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", Base64::encode(self.as_ref()))
+    }
+}
+
+serialize_deserialize_with_to_from_bytes!(MLDSA65PublicKey, MLDSA65_PUBLIC_KEY_LENGTH);
```
jonas-lj, line 298: Add a similar test for the signing key.

fastcrypto-pq/benches/mldsa65.rs
```diff
@@ -0,0 +1,64 @@
...
+        // Parsing a private key (i.e. the seed) costs a key generation: the wire format is the 32-byte
+        // seed, so from_bytes re-runs the FIPS 204 key expansion.
+        let mut csprng2: ThreadRng = thread_rng();
+        let seed = MLDSA65KeyPair::generate(&mut csprng2).as_ref().to_vec();
+        group.bench_function("MLDSA65 private key from bytes", move |b| {
+            b.iter(|| MLDSA65PrivateKey::from_bytes(&seed).unwrap())
+        });
+    }
+
+    criterion_group! {
```
jonas-lj, line 86: Add benchmarks for serde

fastcrypto-pq/src/mldsa65/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+
+//! This module contains an implementation of the [ML-DSA-65](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf)
+//! signature scheme from FIPS 204, on top of the `mysten-mldsa-native-rs` wrapper around the
+//! mldsa-native C library.
+//!
+//! The private key is the 32-byte seed from which FIPS 204 derives the full signing key.
+//! Serialization emits only the seed, and key expansion is fully specified by the standard.
+//! Signing is hedged (the FIPS 204 default): each signature draws fresh randomness from the
+//! operating system, so two signatures over the same msg differ while both verify. The FIPS
+//! 204 context string is fixed to empty but the wrapper gets context up to 255-bytes.
```
jonas-lj: Would it make sense to have a constant dst for fastcrypto?

fastcrypto-pq/src/mldsa65/mod.rs
mahdi-mysten: True, we only cache PrivateKey though we need to find the PublicKey, so we only read public from this expansion call.
https://github.com/MystenLabs/fastcrypto/blob/0c56dda47923fa234351745c79983f3552e89ab2/fastcrypto-pq/src/mldsa65/mod.rs#L270

One solution could be to cache PublicKey too, let me check the details.

fastcrypto-pq/src/mldsa65/mod.rs
mahdi-mysten: Good point, though I'd keep ctx empty here since it'd be consensus-critical to change later. Sui already domain-separates at the intent layer, so a dst here would be a second copy of that job. Also all the vectors we pin against (KATs, noble/aws-lc-rs, the TS interop test) use empty ctx which is the default setting in FIPS 204, and any external verifier like HSM SDKs would need the same constant wired in. If we ever want a dst it can live in the intent message without touching the scheme.

fastcrypto-pq/src/mldsa65/mod.rs
mahdi-mysten: left a comment note on this: the wrapper, i.e. mldsa-native, has no pk-from-signing-key accessor and the code now says so.

jonas-lj: Oh, so we'd have to cache the verifying key as part of MLDSA65PrivateKey then?

fastcrypto-pq/src/mldsa65/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+    fn from_seed(seed: mldsa::SigningKeySeed) -> Self {
+        let (signing_key, verifying_key) = seed.expand();
+        MLDSA65KeyPair {
+            public: MLDSA65PublicKey(verifying_key),
+            private: MLDSA65PrivateKey { seed, signing_key },
+        }
+    }
+}
+
+/// The one construction path that must expand the seed a second time: recovering
```
jonas-lj: This doc is too long. Just focus on what it does and put the other details as code comments if needed.

jonas-lj (approved):

fastcrypto-pq/src/mldsa65/mod.rs
mahdi-mysten: True and this is the only way to save 1 expansion, the only issue is caching the verifying key in MLDSA65PrivateKey is quite ugly, bc it'd put public data in the private-key struct. We have keypair-first path: `from_seed` that expands once and keeps both halves. I'd keep that as it is, slightly slow path rather than caching.

PR #1003, opened 2026-08-12
[zkp] Migrate from im to imbl

stefan-mysten:
The im crate is unmaintained (RUSTSEC-2025-0034) and trips cargo-deny in downstream repos such as sui. 

This PR replaces it with `imbl 7`, the maintained fork with an identical API. Note that `verify_zk_login`'s public signature now takes an `imbl::HashMap`, so downstream callers must switch alongside the fastcrypto bump.


joyqvq (approved):

stefan-mysten:
@joyqvq thanks @joyqvq. I am not sure how releases work here, but I think for `sui`, I need to pin this crate to a specific sha. Will try that when trying to fix the cargo deny stuff there.

PR #1004, opened 2026-08-13
Precomputed MSMs

Daeinar:
## Summary

Adds a `PrecomputedMultiScalarMul` capability trait to `fastcrypto::groups` and implements it for Ristretto255.

The trait covers multi-scalar multiplications where part of the point set is fixed and known in advance: `precompute(points)` builds per-point tables once, and `mixed_multi_scalar_mul` then evaluates

```
a_1P_1 + ... + a_nP_n + b_1Q_1 + ... + b_mQ_m
```

where `P_1, ..., P_n` are the precomputed (static) points whose scalars change per call, and the `Q_j` are dynamic points supplied fresh each call, evaluated in a single joint MSM. This trades memory for speed when the same points (e.g. a commitment key) recur across many MSMs.

## Motivation

An upcoming Bulletproofs++ range proof module (follow-up PR) reduces verification to one large MSM in which the majority of points are fixed CRS generators shared by every verification. Precomputing tables for those generators once and reusing them per proof is a significant verify-time win. The trait is generic so other fixed-basis workloads (Pedersen vector commitments, batch verification) can use it too.

## Design notes

- The associated `Precomputation` type leaves the table representation to each group. The Ristretto255 implementation wraps dalek's `VartimeRistrettoPrecomputation`.
- Variable-time, like the existing `MultiScalarMul` for this group; intended for public inputs (verifiers).
- The wrapper stores the precomputed set size and validates both `static_scalars.len()` and the dynamic slice lengths up front, returning `InvalidInput` where dalek would otherwise `assert!` (static-length mismatch) or silently misbehave.

## Testing

`test_precomputed_multiscalar_mul` checks agreement with the plain `multi_scalar_mul` over the concatenated point set, both with and without dynamic terms (`m = 0`), and exercises the three error paths (static scalar count above/below the table size, mismatched dynamic scalar/point lengths).

### DIFF
--- a/fastcrypto/src/groups/mod.rs
+++ b/fastcrypto/src/groups/mod.rs
@@ -131,6 +131,36 @@ pub trait MultiScalarMul: GroupElement {
     fn multi_scalar_mul(scalars: &[Self::ScalarType], points: &[Self]) -> FastCryptoResult<Self>;
 }
 
+/// Trait for groups that support multi-scalar multiplication with precomputed
+/// tables over a fixed set of points, trading memory for speed when the same
+/// points (e.g. a commitment key) are used in many multi-scalar
+/// multiplications.
+pub trait PrecomputedMultiScalarMul: GroupElement {
+    /// Precomputed tables for a fixed set of points.
+    type Precomputation;
+
+    /// Build precomputation tables for `points`.
+    fn precompute(points: &[Self]) -> FastCryptoResult<Self::Precomputation>;
+
+    /// Compute the mixed multi-scalar multiplication
+    ///
+    /// `a_1*P_1 + ... + a_n*P_n + b_1*Q_1 + ... + b_m*Q_m`,
+    ///
+    /// where `P_1, ..., P_n` are the fixed points `precomputation` was built
+    /// for and only their scalars `a_i` (= `static_scalars`, which must have
+    /// exactly the length of the precomputed set) change between calls, while
+    /// the scalars `b_j` (= `dyn_scalars`) and points `Q_j` (= `dyn_points`)
+    /// are both given fresh on every call. The static half is evaluated from
+    /// the precomputed tables, so when the `P_i` are reused across many calls
+    /// this is faster than a regular MSM over all n + m points.
+    fn mixed_multi_scalar_mul(
+        precomputation: &Self::Precomputation,
+        static_scalars: &[Self::ScalarType],
+        dyn_scalars: &[Self::ScalarType],
+        dyn_points: &[Self],
+    ) -> FastCryptoResult<Self>;
+}
+
 /// Faster deserialization that skips validation of the result: the subgroup membership check for
 /// curve points and the canonical range check for scalars. Only safe for trusted input; otherwise
 /// use [`crate::serde_helpers::ToFromByteArray::from_byte_array`].
--- a/fastcrypto/src/groups/mod.rs
+++ b/fastcrypto/src/groups/mod.rs
@@ -135,30 +135,40 @@ pub trait MultiScalarMul: GroupElement {
 /// tables over a fixed set of points, trading memory for speed when the same
 /// points (e.g. a commitment key) are used in many multi-scalar
 /// multiplications.
-pub trait PrecomputedMultiScalarMul: GroupElement {
+pub trait PrecomputableMultiScalarMul: GroupElement {
     /// Precomputed tables for a fixed set of points.
-    type Precomputation;
+    type Precomputation: MixedMultiScalarMul<Point = Self>;
 
     /// Build precomputation tables for `points`.
     fn precompute(points: &[Self]) -> FastCryptoResult<Self::Precomputation>;
+}
+
+/// Trait for the precomputed tables built by [PrecomputableMultiScalarMul::precompute].
+pub trait MixedMultiScalarMul {
+    /// The group whose points the tables were built for.
+    type Point: GroupElement;
+
+    /// The number of points these tables were built for.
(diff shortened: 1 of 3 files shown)


fastcrypto/src/groups/mod.rs
```diff
@@ -131,6 +131,36 @@ pub trait MultiScalarMul: GroupElement {
...
+    /// where `P_1, ..., P_n` are the fixed points `precomputation` was built
+    /// for and only their scalars `a_i` (= `static_scalars`, which must have
+    /// exactly the length of the precomputed set) change between calls, while
+    /// the scalars `b_j` (= `dyn_scalars`) and points `Q_j` (= `dyn_points`)
+    /// are both given fresh on every call. The static half is evaluated from
+    /// the precomputed tables, so when the `P_i` are reused across many calls
+    /// this is faster than a regular MSM over all n + m points.
+    fn mixed_multi_scalar_mul(
+        precomputation: &Self::Precomputation,
+        static_scalars: &[Self::ScalarType],
```
jonas-lj: I think it'd be simpler if we have just one slice of scalars and assume that the first n are for the static points.

fastcrypto/src/groups/mod.rs
```diff
@@ -131,6 +131,36 @@ pub trait MultiScalarMul: GroupElement {
...
+    /// `a_1*P_1 + ... + a_n*P_n + b_1*Q_1 + ... + b_m*Q_m`,
+    ///
+    /// where `P_1, ..., P_n` are the fixed points `precomputation` was built
+    /// for and only their scalars `a_i` (= `static_scalars`, which must have
+    /// exactly the length of the precomputed set) change between calls, while
+    /// the scalars `b_j` (= `dyn_scalars`) and points `Q_j` (= `dyn_points`)
+    /// are both given fresh on every call. The static half is evaluated from
+    /// the precomputed tables, so when the `P_i` are reused across many calls
+    /// this is faster than a regular MSM over all n + m points.
+    fn mixed_multi_scalar_mul(
```
jonas-lj, line 167: We can put this function on another trait (maybe even the existing MultiscalarMul trait?) that the `Precomputation` should implement. Then we can call `precomp.multi_scalar_mul(...)` instead of RistrettoPoint::multi_scalar_mul(precomp, ...)`

fastcrypto/src/groups/mod.rs
jonas-lj: Then we should probably add a function which returns how many static points a precomputation has

Daeinar:
@jonas-lj Okay, I addressed both comments. The API is now:

```
pub trait PrecomputableMultiScalarMul: GroupElement {
    type Precomputation: MixedMultiScalarMul<Point = Self>;
    fn precompute(points: &[Self]) -> FastCryptoResult<Self::Precomputation>;
}

pub trait MixedMultiScalarMul {
    type Point: GroupElement;
    fn num_static_points(&self) -> usize;
    fn mixed_multi_scalar_mul(
        &self,
        scalars: &[<Self::Point as GroupElement>::ScalarType],
        dynamic_points: &[Self::Point],
    ) -> FastCryptoResult<Self::Point>;
}
```

Re single slice, there's a trade-off worth knowing: passing the two halves in the wrong order has the correct total length, so it is accepted but returns a different point (resulting in a completeness error, not a soundness one). The ordering requirement is documented on the method and by a test.

I put `mixed_multi_scalar_mul` on another trait (see above). Putting it on `MultiScalarMul` seemed overly complex: it requires `GroupElement`, which a table of precomputed multiples can't implement (it needs `Copy`, and dalek's precomputation type isn't even `Clone`).

fastcrypto/src/tests/ristretto255_tests.rs
```diff
@@ -238,6 +241,69 @@ fn test_multiscalar_mul() {
...
+    };
+    let static_points: Vec<RistrettoPoint> = rand_scalars(5, &mut rng)
+        .iter()
+        .map(|s| RistrettoPoint::generator() * s)
+        .collect();
+    let dynamic_points: Vec<RistrettoPoint> = rand_scalars(3, &mut rng)
+        .iter()
+        .map(|s| RistrettoPoint::generator() * s)
+        .collect();
+    let precomputation = RistrettoPoint::precompute(&static_points).unwrap();
```
jonas-lj, line 258: Make sure we test the case where both the static and dynamic (and both) points are empty.

fastcrypto/src/groups/ristretto255.rs
```diff
@@ -85,6 +86,47 @@ impl MultiScalarMul for RistrettoPoint {
...
+    fn mixed_multi_scalar_mul(
+        &self,
+        scalars: &[RistrettoScalar],
+        dynamic_points: &[RistrettoPoint],
+    ) -> FastCryptoResult<RistrettoPoint> {
+        if scalars.len() != self.num_points + dynamic_points.len() {
+            return Err(InvalidInput);
+        }
+        let (static_scalars, dynamic_scalars) = scalars.split_at(self.num_points);
+        Ok(RistrettoPoint(self.tables.vartime_mixed_multiscalar_mul(
```
jonas-lj, line 122: is there a maximum number of points this function can accept?

Daeinar: No, there's no hard limit. The precomputed path uses the Straus algorithm, which loses to Pippenger for large sets. So past a few hundred points (~440 in the BP++ benchmarks), precomputation stops paying off, and plain MSMs are better. The BP++ verifier will switch based on that threshold.

jonas-lj (approved):

PR #1007, opened 2026-08-21
Ristretto Precomputed MSM Switcher

Daeinar:
Follow-up to the review of #989 (https://github.com/MystenLabs/fastcrypto/pull/989#discussion_r3821767065): the decision whether a mixed MSM should use the precomputed tables or a plain MSM moves from the caller into `RistrettoPrecomputation`.

## What changes

- `RistrettoPrecomputation` stores the static points next to dalek's tables (dalek exposes no way to recover them). `mixed_multi_scalar_mul` runs Straus over the tables while `static + DYNAMIC_POINT_WEIGHT * dynamic <= MAX_STRAUS_POINTS` (3, 600) and otherwise one plain MSM over the chained static and dynamic inputs, with no concatenation.
- `MixedMultiScalarMul::mixed_multi_scalar_mul` doc: the implementation falls back itself; callers need not choose.
- `test_precomputed_multiscalar_mul` covers the fallback in both directions (static-heavy and dynamic-heavy) with the same length checks.
- New `benches/mixed_msm.rs` times the two dalek primitives over a grid of static sizes (powers of two and midpoints, 64..2048) and dynamic sizes (0, 32, 128, 512); it is how the constants were fitted and how they would be retuned on another machine.

## Why a weighted rule

Straus over the tables is linear in the static count with a flat per-point cost, Pippenger's per-point cost falls with size, and the crossover sits far below what operation counts predict (the 7.7 KB per-point tables fall out of cache). Dynamic points shift the crossover: a dynamic point costs a per-call table and projective additions, about three static points' worth. Measured crossovers on an M2 Max with dense scalars: ~600 static points at D = 0, ~490 at D = 32, ~245 at D = 128; at D = 512 the plain MSM wins at every static size. Least-squares fit `589 - 2.7 D`, rounded to the two constants. Near the crossover the two paths are within a few percent of each other, so a threshold misplaced by ~50 points costs under 1%.

A simpler design, deciding at `precompute` by static count alone and skipping the tables above it, regressed BP++ 64-bit x32 proving by 5.5%: its static-only MSMs still profit from the tables at 521 points while the verifier's mixed call does not.

## Impact

BP++ (#989) keeps the decisions its hand-written fallback made (tables for all its provers and for verifiers up to 64-bit x16, plain MSM for the 64-bit x32 verifier) and will drop its own fallback arm once this lands. No other callers of the precomputation exist yet.

### DIFF
--- a/fastcrypto/src/groups/ristretto255.rs
+++ b/fastcrypto/src/groups/ristretto255.rs
@@ -65,6 +65,23 @@ impl RistrettoPoint {
             .fill_bytes(&mut bytes);
         Self::from_uniform_bytes(&bytes)
     }
+
+    /// Build precomputation tables for `points` under `strategy`;
+    /// [PrecomputableMultiScalarMul::precompute] uses `MixedMsmStrategy::default()`.
+    pub fn precompute_with_strategy(
+        points: &[Self],
+        strategy: MixedMsmStrategy,
+    ) -> FastCryptoResult<RistrettoPrecomputation> {
+        let points: Vec<ExternalPoint> = points.iter().map(|p| p.0).collect();
+        let tables = strategy
+            .use_tables(points.len(), 0)
+            .then(|| VartimeRistrettoPrecomputation::new(points.iter()));
+        Ok(RistrettoPrecomputation {
+            tables,
+            points,
+            strategy,
+        })
+    }
 }
 
 impl Doubling for RistrettoPoint {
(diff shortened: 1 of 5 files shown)


Daeinar:
@jonas-lj 👆 Ready for your review. Thanks!

fastcrypto/src/groups/ristretto255.rs
```diff
@@ -86,19 +86,40 @@ impl MultiScalarMul for RistrettoPoint {
     }
 }
 
-/// Precomputed multiplication tables over a fixed set of Ristretto points.
+/// Straus over the precomputed tables beats a plain MSM (dalek's Pippenger
```
jonas-lj: On my machine (M1 Pro) the crossover is ~1000 with no dynamic points, and ~750 at 128, so with this constant we'd take the plain path where the tables are still ~10% faster. A heuristic is fine, but maybe we don't need all the reasoning here if it doesn't transfer.

fastcrypto/src/groups/ristretto255.rs
```diff
@@ -86,19 +86,40 @@ impl MultiScalarMul for RistrettoPoint {
...
     tables: VartimeRistrettoPrecomputation,
-    num_points: usize,
+    points: Vec<ExternalPoint>,
 }
 
 impl PrecomputableMultiScalarMul for RistrettoPoint {
     type Precomputation = RistrettoPrecomputation;
 
     fn precompute(points: &[Self]) -> FastCryptoResult<Self::Precomputation> {
+        let points: Vec<ExternalPoint> = points.iter().map(|p| p.0).collect();
```
jonas-lj: We can create this with more points than MAX_STRAUS_POINTS, but then the table is never used

fastcrypto/src/groups/ristretto255.rs
```diff
@@ -86,19 +86,40 @@ impl MultiScalarMul for RistrettoPoint {
...
+/// `static + DYNAMIC_POINT_WEIGHT * dynamic` stays within this bound. A
+/// heuristic, measured with dense scalars on an Apple M2 Max
+/// (`benches/mixed_msm.rs`): the crossover lies at about 600 static points
+/// with no dynamic ones, 490 with 32 and 245 with 128, and with 512 dynamic
+/// points the plain MSM wins at every static size. The least-squares fit is
+/// `589 - 2.7 * dynamic`; with the weight rounded to 3 the bound refits to
+/// about 605, rounded down here. The crossover sits far below what
+/// operation counts predict, consistent with the ~7.7 KB per-point tables
+/// falling out of cache.
+pub(crate) const MAX_STRAUS_POINTS: usize = 600;
```
jonas-lj: MAX_WEIGHTED_STRAUS_POINTS?

Daeinar:
@jonas-lj I addressed your comments and added the strategy parameterization technique that we discussed earlier today. Please let me know what you think. Thanks!

fastcrypto/src/groups/ristretto255.rs
```diff
@@ -86,28 +103,83 @@ impl MultiScalarMul for RistrettoPoint {
...
+/// without breaking callers.
+#[non_exhaustive]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum MixedMsmStrategy {
+    /// Straus while
+    /// `num_static + dynamic_point_weight * num_dynamic <= max_weighted_points`;
+    /// tables are only built if the static points alone satisfy the rule.
+    /// `max_weighted_points = 0` never builds tables; `dynamic_point_weight = 0`
+    /// with a large bound always uses them.
+    Weighted {
```
jonas-lj: Using the default trait is a good idea, but since there's only one variant, this shouldn't be an enum then. Just make it a struct instead.

fastcrypto/src/groups/ristretto255.rs
```diff
@@ -86,28 +103,83 @@ impl MultiScalarMul for RistrettoPoint {
...
+        max_weighted_points: usize,
+        dynamic_point_weight: usize,
+    },
+}
+
+impl Default for MixedMsmStrategy {
+    /// `Weighted` with constants measured with dense scalars on an Apple M2
```
jonas-lj: Please shorten the docs

jonas-lj (approved):

PR #1013, opened 2026-08-27
perf: cache the Bulletproofs generators instead of rebuilding them per call

jonas-lj:
`prove_batch` and `verify_batch` rebuilt `BulletproofGens` on every call. This PR caches them on first use.

Verification benchmarks:

| Config | before | after | speedup |
| --- | ---: | ---: | ---: |
| 16-bit x4 | 2.12 | 1.34 | 1.6x |
| 16-bit x8 | 3.86 | 1.92 | 2.0x |
| 16-bit x16 | 7.98 | 3.63 | 2.2x |
| 16-bit x32 | 13.37 | 5.58 | 2.4x |


Daeinar (approved):
lgtm

PR #1017, opened 2026-08-31
Reject a zero-nonce dealer on the verifier side in batch AVSS

jonas-lj:
Both batch AVSS receivers derive the expected batch size as `weight_of(dealer_id) * batch_size_per_weight`. When that product is zero — a zero-weight dealer, or a zero `batch_size_per_weight` with an ordinary one — the size check compares 0 against 0, so a hand-built message with an empty `full_public_keys` passes. Checking this in `Dealer::new` would not help, since a malicious party never calls it.

`Receiver::new` now returns `InvalidInput` when `batch_size == 0`. A test in each module records the dealer-side behaviour being rejected: a zero-weight dealer builds its messages, but deals no nonces.

### DIFF
--- a/fastcrypto-tbls/src/threshold_schnorr/batch_avss.rs
+++ b/fastcrypto-tbls/src/threshold_schnorr/batch_avss.rs
@@ -334,7 +334,8 @@ impl Receiver {
     /// * `enc_secret_key` is this Receivers' secret key for the distribution of nonces. The corresponding public key is defined in `nodes`.
     /// * `batch_size_per_weight` is the number of secrets a dealer should deal per weight it has.
     ///
-    /// Returns an `InvalidInput` error if the `id` or `dealer_id` is invalid.
+    /// Returns an `InvalidInput` error if the `id` or `dealer_id` is invalid, or if the dealer has
+    /// no nonces to deal because either its weight or `batch_size_per_weight` is zero.
     pub fn new(
         nodes: Nodes<EG>,
         id: PartyId,
@@ -349,6 +350,9 @@ impl Receiver {
 
         // The dealer is expected to deal a number of nonces proportional to it's weight
         let batch_size = nodes.weight_of(dealer_id)? as usize * batch_size_per_weight as usize;
+        if batch_size == 0 {
+            return Err(InvalidInput);
+        }
 
         Ok(Self {
             id,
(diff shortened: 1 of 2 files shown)


fastcrypto-tbls/src/threshold_schnorr/batch_avss_avid.rs
```diff
@@ -251,7 +254,11 @@ impl Dealer {
         params.validate(total_weight)?;
         let nodes = Arc::new(nodes);
         let avid = avid::Avid::new(Arc::clone(&nodes), params.f)?;
```
zhouwfang, line 253: Worth moving the guard above this line. Avid::new builds an ErasureCoder over the total weight, and on the rejection path it gets thrown away. For a zero-weight node that's every round.

fastcrypto-tbls/src/threshold_schnorr/batch_avss.rs
```diff
@@ -229,7 +230,11 @@ impl Dealer {
             return Err(InvalidInput);
         }
         // Each dealer deals a number of nonces proportional to their weight.
-        let batch_size = nodes.weight_of(dealer_id)? as usize * batch_size_per_weight as usize;
+        let dealer_weight = nodes.weight_of(dealer_id)?;
+        if dealer_weight == 0 {
+            return Err(InvalidInput);
+        }
+        let batch_size = dealer_weight as usize * batch_size_per_weight as usize;
```
zhouwfang: Same at batch_avss_avid.rs:261

The property you're protecting is batch_size == 0, and this guards one of the two factors. A zero batch_size_per_weight gives you the same degenerate session with a normal dealer. presigning.rs:69 already guards that one, so it's a real input. Checking batch_size == 0 after the multiply would cover both and match the doc comment more closely.

fastcrypto-tbls/src/threshold_schnorr/batch_avss_avid.rs
```diff
@@ -240,6 +240,9 @@ impl Dealer {
     /// * `sid` is a session identifier that should be unique for each invocation of a dealer, but
     ///   the same for all parties in the same session.
     /// * `batch_size_per_weight` is the number of secrets a dealer should deal per weight it has.
+    ///
+    /// Returns an `InvalidInput` error if the `dealer_id` is invalid (not part of `nodes`) or if
```
zhouwfang: This closes the honest half only. A malicious zero-weight party skips Dealer::new and hand-builds the message, which passes AvssCommonMessage::verify because the size check is 0 != 0. Worth narrowing the claim, or following up receiver-side.

fastcrypto-tbls/src/threshold_schnorr/batch_avss.rs
```diff
@@ -349,6 +350,9 @@ impl Receiver {
 
         // The dealer is expected to deal a number of nonces proportional to it's weight
         let batch_size = nodes.weight_of(dealer_id)? as usize * batch_size_per_weight as usize;
+        if batch_size == 0 {
```
zhouwfang, line 353: This makes previously valid messages invalid, so it applies retroactively to anything a consumer has already persisted, not just to new traffic. We'll add a guard on our side for that.

zhouwfang (approved):

PR #1021, opened 2026-09-01
[zklogin] versioned zklogin inputs, v2 uses sha2

joyqvq:


fastcrypto-zkp/src/bn254/sha256_transcripts.rs
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) Mysten Labs, Inc.
```
joyqvq: rust functions -> typescript functions:

append_fixed_width_be → bigIntToFixedBytesBE
append_length_prefixed_bytes → encodeLengthPrefixedString
append_length_prefixed_padded_bytes → encodePaddedBytes
sha256_low_253 → sha256 + sha256DigestToField

PR #1022, opened 2026-09-07
Derive serde for BP++ proofs instead of hand-written to/from bytes

jonas-lj:
## What

`CircuitProof` and `NormLinearProof` now derive `Serialize`/`Deserialize`, and `RangeProof::{to_bytes, from_bytes}` go through `bcs`. `RistrettoPoint` and `RistrettoScalar` already have canonical 32-byte serde impls (via `serialize_deserialize_with_to_from_byte_array`), so the derived encoding is byte-identical to the hand-written one apart from one length prefix per vector in the norm-linear proof.

**Cost: 3 bytes per proof.** Every configuration grows by exactly 3 (419 instead of 416 for the 1x64 shape); the README table and the size formula (`64*log2(nm) + 163`) are updated and were checked against measured sizes.

**Saving: ~100 lines of hand-written codec.**

- `CircuitProof::to_bytes`/`from_bytes`, including the shape search over powers of two and the `MAX_LOG_NM` bound it needed
- `NormLinearProof::to_bytes`/`from_bytes`/`serialized_len`/`serialized_len_for`
- `decode_next` in `util.rs`

## Trade-off

The proof shape used to be *derived* from the total byte length, so anything that parsed had self-consistent vector lengths. Now the lengths are declared on the wire and a decoded proof may carry any of them. This was already checked where it matters:

- `norm_linear::verify` rejects any shape other than the one `proof_shape` gives for the statement's base lengths — covered by `test_wrong_shape_fails`, and by the cross-configuration tests in `range_proof.rs`
- element encodings are still validated at parse time: the serde impls call `from_byte_array`, exactly as `decode_next` did
- `bcs::from_bytes` rejects trailing bytes, so there is still exactly one valid encoding per proof

`RangeProof` also implements `Serialize`/`Deserialize` now, so it can be nested in other serde types without going through bytes.

## Regression vector

The frozen vector in `regression_test` is the same proof re-encoded (the old bytes with the three length prefixes spliced in), not a fresh proof, so it still pins the transcript layout, challenge derivation and generator derivation. It verifies against the unchanged committed commitments.

## Testing

`cargo test -p fastcrypto --lib --features experimental` — 409 passed. Clippy clean.

### DIFF
--- a/fastcrypto/src/bulletproofspp/README.md
+++ b/fastcrypto/src/bulletproofspp/README.md
@@ -25,21 +25,21 @@ base to `pedersen::H` and the blinding base to `pedersen::G`, so existing
 ## Proof sizes
 
 Proofs are serialized with `bcs`: every group element and scalar in its canonical
-32-byte encoding, and one length prefix for each of the norm-linear proof's three
-vectors. That gives a size of `64*log2(nm) + 163` bytes, where `nm = max(M*b/4, 16)`
-rounded up to a power of two.
+32-byte encoding, and one length prefix for each of the norm-linear proof's two
+variable-length vectors (its final `l` is a single scalar). That gives a size of
+`64*log2(nm) + 162` bytes, where `nm = max(M*b/4, 16)` rounded up to a power of two.
 The table below compares BulletProofs (BP) and BulletProofs++ proof sizes for different (aggregate) range proof configurations.
 
 | Config | BP (bytes) | BP++ (bytes) | smaller |
 | --- | ---: | ---: | ---: |
-| 16-bit x1 | 544 | 419 | 23% |
-| 32-bit x1 | 608 | 419 | 31% |
-| 64-bit x1 | 672 | 419 | 38% |
-| 16-bit x4 | 672 | 419 | 38% |
-| 16-bit x8 | 736 | 483 | 34% |
-| 32-bit x8 | 800 | 547 | 32% |
-| 64-bit x16 | 928 | 675 | 27% |
-| 64-bit x32 | 992 | 739 | 26% |
+| 16-bit x1 | 544 | 418 | 23% |
+| 32-bit x1 | 608 | 418 | 31% |
+| 64-bit x1 | 672 | 418 | 38% |
+| 16-bit x4 | 672 | 418 | 38% |
+| 16-bit x8 | 736 | 482 | 34% |
+| 32-bit x8 | 800 | 546 | 32% |
+| 64-bit x16 | 928 | 674 | 27% |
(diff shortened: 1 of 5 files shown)


jonas-lj:
Pushed a second commit making `NormLinearProof::l_final` a `RistrettoScalar` instead of a `Vec`, which removes one of the three length prefixes: **proofs are now 418 bytes** for the 1x64 shape — one more than the hand-written codec's 416, not three.

`l_final` was always length 1 for every shape the range proof can produce, because `l_len = H_LEN = 8` always folds down to a single scalar. To make that structural rather than a coincidence, `proof_shape`'s stopping condition gains `l_len > 1`, and it now returns just `(rounds, n_len)`.

For `l_len = 8` that stopping rule is a no-op at every `nm` from 16 to 8192 — identical round count and final lengths — so **the proof contents and the transcript are unchanged**. The frozen regression vector in this commit is the same proof with the `l_final` prefix byte spliced out, and it still verifies. The rule only changes shapes like `(3,5)` and `(4,1)` that the `norm_linear` module's own tests exercise and the circuit never instantiates.

Verification gets slightly simpler too: `sigma`'s inner product over a length-1 vector becomes `c[0] * l_final`, and the folded `l` coordinates all read the one scalar rather than indexing by `i >> k`.

Remaining prefixes are `rounds` (genuinely varies, 3–11) and `n_final` (2 at `nm = 16`, 4 above). Dropping those two as well would need the shape passed in at decode time — `from_bytes(bytes, range, m)` plus a `DeserializeSeed` — which re-adds roughly the code this PR removes.

jonas-lj:
Third commit adds parse-time shape validation, which closes the one real gap the derived encoding opened.

Since the vector lengths are declared on the wire, a decoded proof could previously carry any lengths and only fail later in `norm_linear::verify`. Now `#[serde(deserialize_with = "deserialize_nl_proof")]` on the `nl_proof` field accepts a shape only if some norm length from `BASE` to `2^MAX_LOG_NM` produces it — restoring the hand-written codec's guarantee that anything which decodes has a shape the protocol can produce.

It needs no seed and keeps `derive`, so `RangeProof` still implements `Deserialize` and can be nested in other serde types.

The split of responsibility is now explicit and tested: shapes no norm length gives (`(2,2)`, `(3,3)`) are rejected at decoding, while a shape another norm length *does* give (`(3,4)`, the `nm = 32` shape) decodes and is rejected by `verify` against this statement.

Net against `main`: 165 insertions, 197 deletions. 410 tests pass, clippy clean.

jonas-lj:
Superseded by #1023, which reaches the same place without changing the wire format.

PR #1034, opened 2026-09-17
[feat] Generic SLIP-0010 derivation with ML-DSA-65 master keys

mahdi-mysten:
## Description

Adds generic SLIP-0010 hardened derivation to `fastcrypto`, including the ML-DSA-65 master key constants from satoshilabs/ slips#1968 and `MLDSA65KeyPair::derive_slip10` in `fastcrypto-pq`.

Orthogonal: Also updates `fastcrypto-pq` to the published `mysten-mldsa-native-rs` 0.2.0, which uses `mldsa-native` v2.0.0 instead of the beta. Existing interop vectors remain unchanged.

## Motivation

Sui currently derives Ed25519 keys through the external `slip10_ed25519` crate, which hardcodes `b"ed25519 seed"` in its only public API. That doesn't work for ML-DSA-65, which uses its own SLIP-0010 master key.

Instead of duplicating the HMAC derivation logic in `sui-keys`, this adds it once to `fastcrypto`. The implementation is byte-identical to `slip10_ed25519` on the SLIP-0010 Ed25519 test vectors, while supporting new schemes through a master-key constant and a small constructor. Existing Sui key derivation is unchanged.

## Test plan

* Full SLIP-0010 Ed25519 test vectors
* SLIPS #1968 ML-DSA-65 test vectors, including chain codes and public-key digests
* `cargo xclippy -D warnings`
* `cargo fmt --check`


fastcrypto/src/tests/slip10_tests.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+/// The domain separation the master-key parameter exists for.
+#[test]
+fn master_key_separates_schemes() {
+    let seed = Hex::decode(SEED_1).unwrap();
+    let ed = derive_hardened(b"ed25519 seed", &seed, &[0]);
+    let pq = derive_hardened(b"ML-DSA-65 seed", &seed, &[0]);
+    assert_ne!(ed.secret, pq.secret);
+}
```
benr-ml: add regression tests for ML-DSA

fastcrypto/src/slip10.rs
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Generic [SLIP-0010](https://github.com/satoshilabs/slips/blob/master/slip-0010.md)
+//! hardened-only hierarchical key derivation.
+//!
+//! SLIP-0010 parametrizes its HMAC-SHA512 tree by one per-scheme master key
+//! (the "Curve" string: `"ed25519 seed"`, `"Bitcoin seed"`, ...). This
```
benr-ml, line 8: I don't think that bitcoin is a good example here, lets remove it

fastcrypto/src/slip10.rs
```diff
@@ -0,0 +1,53 @@
...
+    pub secret: [u8; 32],
+    pub chain_code: [u8; 32],
+}
+
+/// Derive the SLIP-0010 node for `indexes` under `master_key`.
+///
+/// `master_key` is the per-scheme HMAC key from the SLIP-0010 curve table (or
+/// a proposed extension of it, e.g. satoshilabs/slips#1968 for ML-DSA).
+pub fn derive_hardened(master_key: &[u8], seed: &[u8], indexes: &[u32]) -> Slip10Node {
+    let mut node = hmac_sha512(master_key, seed);
```
benr-ml, line 29: let's reject empty/too-short seed or master key?

fastcrypto/src/slip10.rs
```diff
@@ -0,0 +1,53 @@
...
+pub struct Slip10Node {
+    pub secret: [u8; 32],
+    pub chain_code: [u8; 32],
+}
+
+/// Derive the SLIP-0010 node for `indexes` under `master_key`.
+///
+/// `master_key` is the per-scheme HMAC key from the SLIP-0010 curve table (or
+/// a proposed extension of it, e.g. satoshilabs/slips#1968 for ML-DSA).
+pub fn derive_hardened(master_key: &[u8], seed: &[u8], indexes: &[u32]) -> Slip10Node {
```
benr-ml, line 28: how about receiving an enum for supported master keys? we better verify it per type before supporting it here

fastcrypto-pq/src/slip10.rs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! SLIP-0010 master keys for the post-quantum schemes.
+//!
+//! One string per parameter set, byte for byte as proposed in
+//! [satoshilabs/slips#1968](https://github.com/satoshilabs/slips/pull/1968)
+
+/// ML-DSA-44 (FIPS 204, NIST security level 2).
```
benr-ml, line 9: we only need 65, lets remove the rest

fastcrypto/src/tests/slip10_tests.rs
benr-ml, line 95: add basic regression tests for 65

fastcrypto-pq/src/mldsa65/mod.rs
```diff
@@ -150,6 +150,14 @@ impl MLDSA65KeyPair {
             },
         }
     }
+
+    /// Derive a keypair from a BIP-39 seed along a hardened SLIP-0010 path,
+    /// per the satoshilabs/slips#1968 construction.
+    pub fn derive_slip10(seed: &[u8], indexes: &[u32]) -> Result<Self, FastCryptoError> {
```
benr-ml, line 156: not sure it should be here, i'd rather leave it to the slip10 lib

PR #1036, opened 2026-09-17
Bind presignatures to the message in threshold Schnorr

jonas-lj:
Signing now takes two presigning tuples and combines them into one nonce bound to the message and verifying key (FROST-style), so security no longer depends on whether presignatures are generated before or after the message is fixed. The beacon value is no longer needed and has been removed from the signing API.


fastcrypto-tbls/src/threshold_schnorr/signing.rs
```diff
@@ -14,8 +15,15 @@ use itertools::Itertools;
...
-/// The presigning tuple must be taken from a [Presignatures] iterator, the other parties should use the same tuple and one tuple may only be used once.
+/// Domain separation prefix for the random oracle used to compute presignature binding factors.
+const BINDING_FACTOR_DOMAIN: &str = "fastcrypto_threshold_schnorr_presignature_binding";
+
+/// Generate partial threshold Schnorr signatures for a given message using two presigning tuples.
+/// The presigning tuples are combined into one nonce which is bound to the message and verifying
+/// key, so it does not matter whether the presignatures are generated before or after the message
+/// is known.
+/// The presigning tuples must be taken from a [Presignatures] iterator, the other parties should use the same tuples in the same order and one tuple may only be used once.
+/// Signing two different messages with the same pair of tuples leaks the signing key.
```
zhouwfang, line 26: Two uses of one pair don't leak the key on their own, that's two equations in three unknowns. Three uses on different messages do, and many pairs each used twice is the 2-for-2 ROS forgery. I'd say each tuple goes into exactly one signature.

fastcrypto-tbls/src/threshold_schnorr/signing.rs
```diff
@@ -14,8 +15,15 @@ use itertools::Itertools;
...
-/// Generate partial threshold Schnorr signatures for a given message using a presigning tuple.
-/// The presigning tuple must be taken from a [Presignatures] iterator, the other parties should use the same tuple and one tuple may only be used once.
+/// Domain separation prefix for the random oracle used to compute presignature binding factors.
+const BINDING_FACTOR_DOMAIN: &str = "fastcrypto_threshold_schnorr_presignature_binding";
+
+/// Generate partial threshold Schnorr signatures for a given message using two presigning tuples.
+/// The presigning tuples are combined into one nonce which is bound to the message and verifying
+/// key, so it does not matter whether the presignatures are generated before or after the message
+/// is known.
+/// The presigning tuples must be taken from a [Presignatures] iterator, the other parties should use the same tuples in the same order and one tuple may only be used once.
```
zhouwfang, line 25: The iterator hands out each tuple once, but the signer takes any two distinct tuples, so sliding pairs like (0,1), (1,2), (2,3) reuse every middle tuple and still verify. A next_pair() on Presignatures would keep that discipline in the crate.

fastcrypto-tbls/src/threshold_schnorr/signing.rs
```diff
@@ -188,20 +206,132 @@ pub fn finalize_schnorr_signature(
...
+
+    #[test]
+    fn test_bind_public_presignatures_vector() {
+        let p_0 = G::generator() * S::from(1u128);
+        let p_1 = G::generator() * S::from(2u128);
+        let vk = G::generator() * S::from(3u128);
+        let address = [4u8; 32];
+
+        let r = bind_public_presignatures(b"Hello, world!", &p_0, &p_1, &vk, None).unwrap();
+        assert_eq!(
```
zhouwfang, line 325: This vector is the only thing pinning that b depends on the message and the derived vk, since the mod.rs tests only check that the final signature verifies, which holds for any b everyone agrees on. A test that two messages, or two different addresses, give a different R would close that.

PR #1038, opened 2026-09-17
Document the aggregation path and its caller obligations

jonas-lj:
Documents on `aggregate_signatures` that partial signatures must be authenticated, that a caller needing to identify faulty parties can decode with `RSDecoder` and finalize, and that a failed aggregation must be retried with another subset rather than a new beacon or message.


fastcrypto-tbls/src/threshold_schnorr/signing.rs
```diff
@@ -90,6 +90,16 @@ pub fn generate_partial_signatures(
...
 ///
+/// The partial signatures must be received over an authenticated channel, and the caller must
+/// reject any whose share index the sender does not hold.
+///
+/// This interpolates the first `threshold` partial signatures, so a single faulty one makes the
+/// aggregation fail without indicating which. A caller that needs to identify the faulty parties
+/// can instead decode the partial signatures with an [RSDecoder](crate::threshold_schnorr::reed_solomon::RSDecoder)
+/// and pass the recovered scalar to [finalize_schnorr_signature]. Either way, a failed aggregation
+/// must be retried with another subset of partial signatures for the same presigning tuple,
+/// message and beacon value. Signing twice with the same tuple discloses the signing key.
```
benr-ml, line 101: > Signing twice with the same tuple discloses the signing key.

how is that related here?

benr-ml (approved):

PR #1040, opened 2026-09-17
Correct faulty partial signatures in aggregate_signatures

jonas-lj:
`aggregate_signatures` interpolated the first `t` partial signatures and reported only that the result failed to verify, so a caller had no way to pick a better subset and could only retry blindly.

It now falls back to Reed-Solomon decoding and returns the share indices the decoding excluded, which correcting `e` faults needs `t + 2e` partial signatures to do. An index is only returned when it is certainly corrupted: verification pins just the constant term, so a decoding without margin can settle on another polynomial sharing it and exclude honest indices. Below `n - m >= t + f` the aggregation logs the indices instead of returning them.

It therefore takes `Parameters` rather than a bare threshold, since it needs `f`. `RSDecoder::compute_message_polynomial` becomes `decode`, returning the error locator alongside the message polynomial so the excluded indices are read off its roots. The `reed_solomon` module and `finalize_schnorr_signature` become crate-private, having no callers left outside the crate.


fastcrypto-tbls/src/threshold_schnorr/signing.rs
```diff
@@ -94,20 +96,26 @@ pub fn generate_partial_signatures(
...
+/// A failed aggregation must be retried with another subset of partial signatures for the same
+/// presigning tuple, message and beacon value. Signing twice with the same tuple discloses the
+/// signing key.
 ///
 /// If a derivation index is provided, a new verifying key is derived for this index (see
 /// [derive_verifying_key]), and the signature is adjusted accordingly.
 /// The signature will be valid for the derived verifying key.
 ///
+/// If the signature does not verify, the partial signatures are decoded as a Reed-Solomon code word
+/// to correct the faulty ones, whose share indices are returned along with the signature.
```
zhouwfang, line 108: Calling these faulty is a bit strong. Near the correction radius an adversary can steer the decode so an honest share shows up here while the signature still verifies, so it isn't a reliable blame signal. We hit this in Hashi and stopped evicting on it; could we rename it to something like excluded and note it's a best-effort result, not an attribution to act on?

fastcrypto-tbls/src/threshold_schnorr/signing.rs
```diff
@@ -94,20 +96,26 @@ pub fn generate_partial_signatures(
...
+/// Correcting `e` faults requires `threshold + 2e` partial signatures.
+///
+/// The partial signatures must be received over an authenticated channel, and the caller must
```
zhouwfang, line 111: This authenticated-channel note duplicates the one at line 96 above.

zhouwfang (commented):
Approach makes sense and nice that RSDecoder::new no longer panics, just the naming thing inline.

fastcrypto-tbls/src/threshold_schnorr/signing.rs
jonas-lj: Good point. I've updated the wording to "excluded" and added a note that this doesn't necessarily mean that a node is dishonest.

jonas-lj:
@zhouwfang I've updated the API such that the aggregate_signatures function such that the indices returned from there are certainly from corrupted parties. In case there are excluded partial sigs but we can't determine if they come from corrupted parties we log this and return an empty vector. Let me know if this can work out for you.

fastcrypto-tbls/src/threshold_schnorr/signing.rs
```diff
@@ -94,59 +96,128 @@ pub fn generate_partial_signatures(
...
+            partial_signatures.iter().map(|s| s.index).collect_vec(),
+        );
+        return Ok((signature, vec![]));
+    }
+    Ok((signature, excluded))
+}
+
+/// Whether excluding `excluded` of the `given` partial signatures proves that those indices'
+/// owners submitted a wrong one.
+fn can_blame_excluded_indices(given: usize, excluded: usize, params: Parameters) -> bool {
```
zhouwfang, line 234: Nothing tests the case where this returns false. If I make it always return true, all 32 threshold_schnorr tests still pass.

fastcrypto-tbls/src/threshold_schnorr/signing.rs
```diff
@@ -94,59 +96,128 @@ pub fn generate_partial_signatures(
...
 /// If a derivation index is provided, a new verifying key is derived for this index (see
 /// [derive_verifying_key]), and the signature is adjusted accordingly.
 /// The signature will be valid for the derived verifying key.
 ///
-/// Returns an `InputTooShort` error if not enough partial signatures are provided.
+/// If the signature does not verify, the partial signatures are decoded as a Reed-Solomon code word,
+/// and the share indices the decoding excluded are returned along with the signature. Correcting
+/// `e` faults requires `params.t + 2e` partial signatures.
+///
+/// A returned index is certainly corrupted: its owner submitted a wrong partial signature. Where
```
zhouwfang: This depends on f being right, and nothing checks it. An understated f reports honest indices, and f = 0 is accepted.

fastcrypto-tbls/src/threshold_schnorr/signing.rs
```diff
@@ -94,59 +96,128 @@ pub fn generate_partial_signatures(
...
+/// the decoding had too little margin to establish that, no indices are returned and the aggregation
+/// logs instead, so an empty set does not mean every partial signature was good.
+///
+/// Returns an `InputTooShort` error if fewer than `params.t` partial signatures are provided.
 /// `GeneralOpaqueError` is returned if the computed nonce R is the identity element.
-/// `InvalidSignature` is returned if the aggregated signature does not verify.
+/// `InvalidSignature` is returned if there are too many corrupted partial signatures to correct. The
```
zhouwfang: If the aggregator's own presig is wrong, every partial can be honest and this still fires, and more partials can't fix that. When kept is at least t + f, a failed finalize points at the context, so a separate error would help.

fastcrypto-tbls/src/threshold_schnorr/reed_solomon.rs
```diff
@@ -80,25 +84,31 @@ impl RSDecoder {
...
 
-    /// Try to correct the input and return the decoded message.
-    /// Returns an error if the input length is wrong or if there are too many errors to correct.
-    pub fn decode(&self, input: &[S]) -> FastCryptoResult<Vec<S>> {
-        let mut f1 = self.compute_message_polynomial(input)?.to_vec();
-        f1.truncate(self.k);
-        Ok(f1)
+    /// Whether the code word had an error here.
+    pub fn is_error(&self, index: ShareIndex) -> bool {
+        // Gao remarks after Algorithm 1 that `v(x)` "is in fact the error locator polynomial", so
```
zhouwfang: Gao's remark says the roots contain the error positions, not exactly them. The conclusion holds, but through extended Euclid minimality, so that's the better thing to cite.

zhouwfang (commented):
The gate handles the steering case from my earlier comment, and the math looks sound. My worry is that a caller finalizing early, like hashi, almost never gets indices back, since at its first successful decode reporting needs at least f errors. hashi uses the uncertain set for a per-validator metric, and with RSDecoder going private it can't compute it itself. Could this return the suspected indices alongside the certain ones instead of logging them?

jonas-lj:
> The gate handles the steering case from my earlier comment, and the math looks sound. My worry is that a caller finalizing early, like hashi, almost never gets indices back, since at its first successful decode reporting needs at least f errors. hashi uses the uncertain set for a per-validator metric, and with RSDecoder going private it can't compute it itself. Could this return the suspected indices alongside the certain ones instead of logging them?

Sure. It could also return the excluded indices and a boolean flag indicating if the indices are certainly faulty?

fastcrypto-tbls/src/threshold_schnorr/reed_solomon.rs
jonas-lj: Well, Gao says "contains", but the paper defines the error locator polynomial as the product over exactly the error positions (in the proof of Theorem 3.3), so I read the remark's phrasing as just a bit loose. I've pointed the comment at that definition.

fastcrypto-tbls/src/threshold_schnorr/signing.rs
jonas-lj: Added a test case. Making it always return true now fails test_signing.

fastcrypto-tbls/src/threshold_schnorr/signing.rs
jonas-lj: Added a note that we assume the parameters have been validated for this committee.

fastcrypto-tbls/src/threshold_schnorr/signing.rs
jonas-lj: Added an InconsistentInput error in that case

jonas-lj:
> > The gate handles the steering case from my earlier comment, and the math looks sound. My worry is that a caller finalizing early, like hashi, almost never gets indices back, since at its first successful decode reporting needs at least f errors. hashi uses the uncertain set for a per-validator metric, and with RSDecoder going private it can't compute it itself. Could this return the suspected indices alongside the certain ones instead of logging them?
> 
> Sure. It could also return the excluded indices and a boolean flag indicating if the indices are certainly faulty?

@zhouwfang Actually, I ended up with an enum with three variants: NoCorrect, Blamable(indices), Inconclusive (indices). I think this is the cleanest api.

fastcrypto-tbls/src/threshold_schnorr/signing.rs
```diff
@@ -88,65 +90,155 @@ pub fn generate_partial_signatures(
     ))
 }
 
+/// Who, if anyone, the aggregation can blame for a partial signature inconsistent with the
+/// signature it recovered.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum Blame {
+    /// Every partial signature given was consistent with the recovered signature.
```
zhouwfang: Nobody just means nothing was excluded. If the first t verify we never check the rest, so it's worth saying that so callers don't read it as everyone being honest.

fastcrypto-tbls/src/threshold_schnorr/signing.rs
```diff
@@ -88,65 +90,155 @@ pub fn generate_partial_signatures(
...
 ///
-/// Returns an `InputTooShort` error if not enough partial signatures are provided.
+/// Only the first `params.t` partial signatures are interpolated, so if any of those is wrong, all
+/// of them are decoded as a Reed-Solomon code word instead and the share indices the decoding
+/// excluded are returned along with the signature. Correcting `e` faults requires `params.t + 2e`
+/// partial signatures.
+///
+/// A failed aggregation, reported as an `InvalidSignature` error, may be retried with a new set of
+/// partial signatures as long as the presigning tuple, message and beacon value stay the same.
+/// Reusing the presigning tuple for a different message or beacon value discloses the signing key.
```
zhouwfang: A different derivation address changes the challenge too, so it also leaks the key. I'd keep the old "any reuse leaks" wording and add the address to the InconsistentInputs list. Otherwise a caller might fix a wrong address and ask for fresh partials.

fastcrypto-tbls/src/threshold_schnorr/signing.rs
jonas-lj: True. Added that to the docs of "Nobody".

fastcrypto-tbls/src/threshold_schnorr/signing.rs
jonas-lj: Updated the docs to mention the derivation address also.

fastcrypto-tbls/src/threshold_schnorr/signing.rs
```diff
@@ -88,65 +90,158 @@ pub fn generate_partial_signatures(
     ))
 }
 
+/// Who, if anyone, the aggregation can blame for a partial signature inconsistent with the
+/// signature it recovered.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum Blame {
+    /// Nothing was excluded: either the first `params.t` partial signatures interpolated to a
+    /// valid signature, leaving the rest unexamined, or the decoding found all of them on one
```
zhouwfang: nit: I don't think the second case can happen. If all points are on one polynomial, the first t interpolate to it too, so it would have failed verification the same way the fast path did.

zhouwfang (approved):

PR #1044, opened 2026-09-18
Document that key derivation depends on the parity of the verifying key

jonas-lj:
Documents on `derive_verifying_key` that it must be given the full verifying key, since lifting the x-only BIP-0340 key to even y derives a different child for about half of all keys.


jonas-lj:
Superseded by #1058, which makes derivation independent of the verifying key's parity instead of documenting it.

benr-ml (approved):

PR #1045, opened 2026-09-18
Remove the superseded batch AVSS module

jonas-lj:
Presigning took the output type of the superseded `batch_avss` module, so the AVID-based output had to be converted with `into_legacy` first. Presigning never reads the share index, the only thing the legacy type adds, so `Presignatures::new` now takes `batch_avss_avid::ReceiverOutput` directly.

This removes the `batch_avss` and `complaint` modules, `ReceiverOutput::into_legacy`, and the `use_legacy` height mode of `Presignatures::new`.

Breaking change: callers of `Presignatures::new` must pass AVID outputs and drop the `use_legacy` argument.


fastcrypto-tbls/src/threshold_schnorr/presigning.rs
```diff
@@ -210,10 +202,9 @@ mod tests {
 
     #[test]
     fn test_presig_count_uses_privacy_threshold_not_f() {
```
zhouwfang, line 204: They differ whenever t != f + 1, not just when t > f + 1. validate only rejects t < f, so t == f is in domain, and there the new formula gives one more presig per position rather than fewer. Worth a second assertion with Parameters{t:2, f:2} since that's the case the comment says can't happen.

zhouwfang (approved):
The breaking-change line in the description is a bit narrow though, it isn't only Presignatures::new callers. into_legacy callers break too and they're separate, in hashi that's production code plus a few sites just typed on the legacy ReceiverOutput.

fastcrypto-tbls/src/threshold_schnorr/presigning.rs
jonas-lj: I'll fix that in a follow up pr: https://github.com/MystenLabs/fastcrypto/pull/1069/changes

PR #1052, opened 2026-09-18
Reject a zero batch size in the batch AVSS dealer

jonas-lj:
`Dealer::new` now rejects a dealer with nothing to deal, because either its weight or `batch_size_per_weight` is zero. `Receiver::new` already rejects this case.


fastcrypto-tbls/src/threshold_schnorr/batch_avss_avid.rs
```diff
@@ -252,6 +255,9 @@ impl Dealer {
         let nodes = Arc::new(nodes);
         let avid = avid::Avid::new(Arc::clone(&nodes), params.f)?;
         let batch_size = nodes.weight_of(dealer_id)? as usize * batch_size_per_weight as usize;
+        if batch_size == 0 {
```
benr-ml, line 258: cc @zhouwfang

benr-ml (approved):

PR #1054, opened 2026-09-18
Fully verify the common message for pending recipients

jonas-lj:
`verify_common_message` only compared the common message's hash with the certificate. It now also runs the full verification that voters run when processing the first-phase message, including the session-bound encryption NIZK. `check_ciphertext_hash` also no longer indexes the ciphertext hashes unchecked.


fastcrypto-tbls/src/threshold_schnorr/batch_avss_avid.rs
```diff
@@ -613,20 +613,18 @@ impl Receiver {
...
             return Err(InvalidMessage);
         }
-        let challenge =
-            compute_challenge_from_common_message(&self.random_oracle(), &common_message);
-        Ok(VerifiedAvssCommonMessage {
-            message: common_message,
-            challenge,
-            hash,
-        })
+        common_message.verify(
```
benr-ml, line 622: add a comment that it's a nice to have check since we assume that the certified common message was already verified by at least 1 honest node

benr-ml (approved):

PR #1056, opened 2026-09-18
Compute the Lagrange coefficients once per index in batch AVSS recovery

jonas-lj:
Recovering a node's shares recomputed the Lagrange coefficients for every secret in the batch, although every secret is interpolated from the same share indices. They are now computed once per share index, leaving a dot product per secret.


jonas-lj:
Not pursuing for now: this only speeds up the complaint-recovery path, and batch sizes are small in practice.

PR #1058, opened 2026-09-18
Make key derivation depend only on the x-only verifying key

jonas-lj:
Key derivation hashed only the x-coordinate of the verifying key but added the tweak to the full point, so a key and its negation shared a BIP-340 key but derived different children. Derivation now starts from the even-Y form of the verifying key, so it only depends on the x-only key, and signing negates the challenge accordingly when the verifying key has odd Y. The tweak's HKDF also gets a context string.

**This changes all derived verifying keys, and therefore all Hashi deposit addresses.**


jonas-lj:
Dropped: derivation is also implemented in hashi-ts-sdk and pinned by cross-language test vectors in Hashi, so changing it needs coordinated changes in three repos. Documenting the parity dependence instead (#1044).

PR #1064, opened 2026-09-21
Commit to the full verifying key in the derivation tweak

jonas-lj:
The tweak was computed from the verifying key's x-coordinate but added to the key with its original Y parity, so a key and its negation share a BIP-0340 key yet derive different children. Anyone deriving from the x-only key silently got a different key for half of all verifying keys.

The tweak now commits to the compressed encoding of the verifying key, including its Y parity, so deriving requires the full key. The HKDF also gets a context string, which resolves the TODO about domain separation.

**This changes all derived keys, so Hashi deposit addresses and the derivation in hashi-ts-sdk change with it.**


jonas-lj:
Leaving the key derivation as it is.

